PR #7, opened 2014-05-20
Add support for streaming multi-page query results

grncdr:


grncdr:
This could be re-opened/worked on by anybody in the future, but I don't have the time to finish it now.

PR #17, opened 2015-02-03
Exponential back off when rate-limited

grncdr:
While the `rate-limit` module maintains the invariant that the client will send no more than 6 requests / second, it's possible to hit a pathological case where enough clients will be retrying requests every second that none of them ever succeed.

This commit adds 3 things:
 - The option to opt-out of automatic retries of rate-limited requests
 - Exponential back-off: on each retry of a rate-limited request we wait 2x as long (1 second, 2 seconds, 4 seconds etc)
 - A maximum number of retries: after 5 retries we let the rate limit errors bubble up.


sdepold:
Nice changes!

sdepold:
You know what I'm going to say, don't you?

index.js
```diff
@@ -60,20 +60,24 @@ var Client = redefine.Class({
     this.options = _.defaults({}, options, {
       host: 'api.contentful.com',
       secure: true,
-      rateLimit: 6
+      rateLimit: 6,
+      retryOn429: true,
```
sdepold: I wonder if we can find a nice name which does not include a number?

sdepold:
Please add tests :)

grncdr:
@sdepold I got a bit carried away when I started adding tests, but this is in a much better place now. Want to give it another look?

index.js
```diff
@@ -1,10 +1,15 @@
 'use strict';
 
+require('es6-promise').polyfill();
```
johanneswuerbach: Should we really patch `window` in a client lib? I hated when libs were doing that.

grncdr: Hm, that's a fair point, will change it

grncdr: Fixed

sdepold:
I think you forgot to delete the client code from `index.js`

grncdr:
> I think you forgot to delete the client code from `index.js`

Whoops, `lib/client.js` was not supposed to be added. We can re-organize it into smaller modules in a later PR.

grncdr:
@sdepold @johanneswuerbach all feedback should be addressed now

lib/backoff.js
```diff
@@ -0,0 +1,16 @@
...
+
+var Promise = require('es6-promise');
+var wait_ = require('./wait');
+
+module.exports = function (maxRetries, wait) {
+  wait = wait || wait_;
+  var attempt = 0
+  return function maybeRetry (error, retry) {
+    if (attempt < maxRetries) {
+      return wait(Math.pow(2, attempt++) * 1000).then(retry)
```
johanneswuerbach: Would love to see some random factor here too, but its a good start.

johanneswuerbach:
LGTM

PR #25, opened 2015-08-14
Add API docs from old www.contentful.com examples

grncdr:


grncdr:
Archive and unarchive are still not documented, but this is a hell of a lot better than nothing ;)

trodrigues:
Hah, no worries. Also build seems to be failing?

grncdr:
@trodrigues passed after a restart, a lot of potential for timeouts in these builds :cry:.

PR #29, opened 2015-09-01
[fixed] Strict mode error

grncdr:
Function declarations inside blocks is an error in strict mode. Certain versions of V8 actually notice this and throw an error.

/cc @elblivion

### DIFF
--- a/example/clone-space.js
+++ b/example/clone-space.js
@@ -70,7 +70,10 @@ var destinationClient = contentful.createClient({
 client.getSpace(sourceSpaceId)
       .catch(reportInvalidSpace)
       .then(getDestinationSpace)
-      .tap(logSummary)
+      .then(function (summary) {
+        logSummary(summary);
+        return summary;
+      })
       .spread(clone)
       .done();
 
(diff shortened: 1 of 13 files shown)


grncdr:
Based on these two fixes, I think that tests are not running at all...

elblivion:
Seems happier now, thanks! Now I'm getting this which is probably unrelated - don't know where this `tap` function is supposed to come from:

```
/Users/stanton/Code/contentful-management.js/example/clone-space.js:73
      .tap(logSummary)
       ^
TypeError: undefined is not a function
```

https://github.com/contentful/contentful-management.js/blob/master/example/clone-space.js#L73

sdepold:
It would come from Bluebird which we aren't using anymore.

elblivion:
In any case it's the `getSpace` doesn't seem to be working, I added some debug and it doesn't throw an error for a nonexistent space:

```
console.log('Getting source Space %s from %s using access token %s', sourceSpaceId, host, accessToken);
client.getSpace(sourceSpaceId)
      .catch(reportInvalidSpace);
//      .then(getDestinationSpace)
//      .tap(logSummary)
//      .spread(clone)
      //.done();
```
```
$ bash wrapper
Getting source Space obviouslyfake from api.contentful.com using access token <snip>
$ echo $?
0
```

example/clone-space.js
```diff
@@ -193,6 +196,9 @@ function clone (sourceSpace, destinationSpace) {
   });
 }
 
+function tap$ (fn) {
```
sdepold: not needed I guess?

grncdr: Yeah that can be removed

lib/wait.js
```diff
@@ -1,4 +1,4 @@
-var Promise = require('es6-promise');
+var Promise = require('es6-promise').Promise;
```
johanneswuerbach: Missing `use strict`.

grncdr:
The last test run was an _actual_ test run. There are still some issues with the test running (builds may not fail correctly) so be careful when merging PRs in the future.

PR #33, opened 2015-09-30
Added locale methods

brennhill:


.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+.idea
```
trodrigues: If you'd like to ignore things that are very specific to your developer environment you should look into creating a .global_gitignore file in your home directory, otherwise we'll end up with things specific to lots of development environments here.

trodrigues:
I'm not entirely sure why the build is failing right now but I'll have to look into this.

brennhill:
It said it was missing some key.

grncdr:
This is a Travis-CI feature: pull-requests from outside the repo aren't given access to the encrypted environment variables. We need to make the integration tests skip out gracefully when the env variables aren't set, and add more unit tests so that PR's can at least get _some_ testing without relying on a live API key.

index.js
```diff
@@ -184,6 +184,42 @@ var Space = redefine.Class({
...
+    return this.client.request('/spaces/' + this.sys.id + '/locales');
+  },
+
+
+  /**
+   * Creates a locale
+   * @param name  The readable name of the locale, e.g. U.S English
+   * @param code  the i18n abbreviation, e.g. en-US
+   */
+  createLocale: function(name, code) {
```
johanneswuerbach: Shouldn't this method take an object as other create methods do? https://github.com/contentful/contentful-management.js/blob/master/index.js#L227

brennhill: I think that makes sense.  Would also remove the first line. However what
Should be the behavior if they submit an object with extra key:value
pairs?  Should we strip those?

johanneswuerbach: Nope, our api takes care of that. Passing an object also allows to add additional properties in the future.

trodrigues: In any case, this whole library is going to be entirely refactored very soon.

PR #35, opened 2015-12-08
Replace questor with axios

madtrick:


grncdr:
:+1: LGTM

trodrigues:
Have you tested this on both browser and node environments? When we did the same replacement on contentful.js there were some issues that weren't surfaced till later with the different Promise implementation and that stuff has changed in axios recently (and it was buggy and broken again, and I haven't followed up since then on how it currently is).

It would be good to test this not only on node, but on browsers with native promise implementations (like most of the latest evergreen ones) and browsers without.

That's also why I hadn't done this replacement here before, because there is some testing effort evolved to make sure it doesn't break all the things again.

PR #37, opened 2016-01-07
[Fixed] Add runtime dep on es6-promise

madtrick:
`es6-promise` is used throughout the code (https://github.com/contentful/contentful-management.js/search?utf8=%E2%9C%93&q=es6-promise)  but it's not exported as a runtime dep.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,10 +14,10 @@
     "axios": "^0.8.1",
     "inherits": "^2.0.1",
     "redefine": "^0.2.0",
-    "underscore-contrib": "0.2.2"
+    "underscore-contrib": "0.2.2",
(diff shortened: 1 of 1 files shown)


trodrigues:
Have a look at the README.

Axios expects a Promise to be available natively or polyfilled and there's no easy way around this. Also, we don't want to polyfill by default and cause problems for people who don't expect that to happen, as it was the case with contentful.js.

The actual problem here is removing the requirements of the Promise in our own code which I forgot.

trodrigues:
Fixed on 0.8.1.

madtrick:
I see. Anyways the latest version is broken

madtrick:
Cool. Thanks

PR #51, opened 2016-06-16
Add Editor Interface Support 

Khaledgarbaya:
TODOS : 
- [x] Add Integration Test when the change are deployed in the backend

### DIFF
--- a/lib/entities/content-type.js
+++ b/lib/entities/content-type.js
@@ -15,6 +15,8 @@ import {
   createUpdatedChecker,
   createDraftChecker
 } from '../instance-actions'
+import {wrapEditorInterface} from './editor-interface'
+import errorHandler from '../error-handler'
 
 /**
  * @memberof ContentType
@@ -90,6 +92,19 @@ function createContentTypeApi (http) {
       wrapperMethod: wrapContentType
     }),
 
+    /**
+     * get the editor interface for the object
+     * @memberof ContentType
+     * @func getEditorInterface
+     * @return {Promise<EditorInterface>} Object returned from the server with the current editor interface.
+     * @example
+     * contentType.getEditorInterface()
+     * .then(editorInterface => console.log(editorInterface.controls))
+     */
+    getEditorInterface: function () {
+      return http.get('content_types/' + this.sys.id + '/editor_interface')
+      .then((response) => wrapEditorInterface(http, response.data), errorHandler)
+    },
     /**
      * Checks if the contentType is published. A published contentType might have unpublished changes (@see {ContentType.isUpdated})
      * @memberof ContentType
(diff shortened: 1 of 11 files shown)


lib/entities/content-type.js
```diff
@@ -91,6 +93,19 @@ function createContentTypeApi (http) {
     }),
 
     /**
+     * get the editor interface for the object
+     * @memberof ContentType
+     * @func getEditorInterface
+     * @return {Promise<EditorInterface>} Object returned from the server with the current editor interface.
```
trodrigues: Should probably also be EditorInterface.EditorInterface here

lib/entities/editor-interface.js
```diff
@@ -0,0 +1,68 @@
...
+     * @return {Promise<EditorInterface>} Object returned from the server with updated changes.
+     * @example
+     * editorInterface.controls[0] = { "fieldId": "title", "widgetId": "singleLine"}
+     * editorInterface.update()
+     * .then(editorInterface => console.log(editorInterface.controls))
+     */
+    update: function () {
+      const raw = this.toPlainObject()
+      const data = omit(raw, ['sys'])
+      return http.put('content_types' + '/' +
```
trodrigues: This first concatenation seems a bit redundant

lib/entities/editor-interface.js
```diff
@@ -0,0 +1,68 @@
...
+       this.sys.contentType.id +
+       '/editor_interface',
```
trodrigues: I'd probably say it would be good to keep this string generation all in one line as it's not too long. You can use ES2015 template strings to make it a bit more compact.

lib/entities/editor-interface.js
```diff
@@ -0,0 +1,68 @@
...
+       data,
+        {
```
trodrigues: This indentation seems a bit inconsistent with the other arguments and it's a bit confusing.

lib/entities/editor-interface.js
```diff
@@ -0,0 +1,68 @@
...
+     * @return {Object} control object for specific field.
+     * @example
+     * const control = editorInterface.getControlForField('fieldId')
+     * console.log(control)
+     */
+    getControlForField: function (fieldId) {
+      const result = this.controls.filter((control) => {
+        return control.fieldId === fieldId
+      })
+      return (result && result.length > 0) ? result[0] : null
```
trodrigues: For cases where the return value can be null, we should document the return value as `{?Object}` to indicate it's nullable. See http://usejsdoc.org/tags-type.html and https://developers.google.com/closure/compiler/docs/js-for-compiler#types for reference.

trodrigues:
Everything looks good but there's a content type test still failing. Is that what you were asking about in #dev maybe?

Khaledgarbaya:
there was a tiny error constructing the editorInterface mock object

PR #75, opened 2017-02-10
Upgrade dependencies and related changes

axe312ger:
* Upgrades dependencies to most current version
* Upgrades to webpack 2 (since it required only minimal effort)
* Fixes some new linting errors due to standard upgrade
* Replaces ghooks with [husky](https://github.com/typicode/husky)
* ~Added a minimal git hook integration since it was not used before~
* switch from standard-cli to eslint with standard config
* upgrade CI test node versions to current state of node (v4, v6 & v7 while dropping v0.10)
* remove vendored axios to prepare for direct file upload

I would love to see this as the first step in using https://greenkeeper.io/ for this repo :)


Khaledgarbaya:
> Added a minimal git hook integration since it was not used before 

It is used [here](https://github.com/contentful/contentful-management.js/blob/master/package.json#L95-L97) also I find husky confusing as a new developer , working on the SDK, will expect to have a `commit` script for the `precommit` which can lead them to remove them thinking that they are not used.

Khaledgarbaya:
👍  for greenkeeper, I will continue the review next week. Good job ;)

axe312ger:
For comparison via https://webpack.github.io/analyse/:

[webpack-build-logs.zip](https://github.com/contentful/contentful-management.js/files/773691/webpack-build-logs.zip)

axe312ger:
> It is used here also I find husky confusing as a new developer , working on the SDK, will expect to have a commit script for the precommit which can lead them to remove them thinking that they are not used.
Jup, I simply overlooked the ghooks config down there in the package.json.

The main reason I switched to husky is that warning:
warning ghooks@1.3.2: Use npmjs.com/husky instead, see https://github.com/gtramontina/ghooks/issues/166

Another option would be using https://github.com/jish/pre-commit but I really like the idea of having linting checks  as precommit and running the whole testsuite as prepush. They will save us and others from force pushing small fixes or doing commits like `fix spacing` etc. after the real commit.

PR #76, opened 2017-02-10
[WIP] proper esdocs implementation

axe312ger:
An updated version of the reference docs, for comparision with the current one, can be found here:
https://axe312ger.github.io/contentful-management.js/


lib/entities/asset.js
```diff
@@ -117,14 +117,14 @@ const ASSET_PROCESSING_CHECK_RETRIES = 5
...
  * // require contentful-management
  * var contentfulManagement = require('contentful-management')
  * var client = contentfulManagement.createClient({
  * // This is the access token for this space. Normally you get both ID and the token in the Contentful web app
  * accessToken: 'YOUR_ACCESS_TOKEN'
- * }) 
- * 
+ * })
+ *
  * let space = await client.getSpace('SPACE_ID')
```
TimBeyer: Not really part of this PR but I feel using `await` is a bit premature since most people won't be able to run it just like that. I think we shouldn't use it in examples until it's widely available.

Khaledgarbaya: Yes @TimBeyer you are right, this was me introducing this in the CDA documentations. We'll work on making all the examples using promises.

lib/create-contentful-api.js
```diff
@@ -1,73 +1,73 @@
 import errorHandler from './error-handler'
 import entities from './entities'
 
+const {wrapSpace, wrapSpaceCollection} = entities.space
 
 /**
- * @typedef {ClientAPI} ClientAPI
```
Khaledgarbaya: I think we can merge this with `ContentfulManagement` Object as it sound like there is a duplicate there

lib/create-contentful-api.js
```diff
@@ -1,73 +1,73 @@
...
+/**
+ * Creates API object with methods to access functionality from Contentful's
+ * Management API
+ *
+ * @param {Object} params - API initialization params
+ * @param {Object} params.http - HTTP client instance
+ * @param {Function} params.shouldLinksResolve - Link resolver preconfigured with global setting
+ * @return {ClientAPI}
+ */
+export default function createClientAPI (params) {
```
Khaledgarbaya: `createClientAPI` should be `@private` so we should not expose in the documentation, as it will be confusing with `createClient`

axe312ger:
Closing that one since we won't switch to esdocs

PR #77, opened 2017-02-14
Direct File Upload

axe312ger:
- [x] integrate new uploads api endpoint
- [x] integration of the actual upload
- [ ] move the upload to the assets entity
- [ ] combine upload with asset creation to ease up usage

### DIFF
--- a/lib/create-space-api.js
+++ b/lib/create-space-api.js
@@ -206,7 +206,8 @@ import omit from 'lodash/omit'
  * @return {ContentfulSpaceAPI}
  */
 export default function createSpaceApi ({
-  http
+  http,
+  httpUploads
 }) {
   const {wrapSpace} = entities.space
   const {wrapContentType, wrapContentTypeCollection} = entities.contentType
@@ -796,6 +797,40 @@ export default function createSpaceApi ({
     }
   }
 
+  /**
+   * Gets an Upload
+   * @memberof ContentfulSpaceAPI
+   * @param  {string} id
+   * @return {Promise<Upload.Upload>} Promise for an Upload
+   * @example
+   * space.getUpload('uploadId')
+   * .then(upload => console.log(upload))
+   */
+  function getUpload (id) {
+    return httpUploads.get('uploads/' + id)
+    .then((response) => wrapEntry(http, response.data), errorHandler)
+  }
+
+  /**
+   * Creates a Upload
+   * @memberof ContentfulSpaceAPI
+   * @see {Upload.Upload}
+   * @param {object} data - File stream representation the files content of the upload to be created. Can be a string, an ArrayBuffer, or a Stream.
+   * @return {Promise<Upload.Upload>} Promise for the newly created Upload
+   * @example
+   * const stream = fs.createReadStream('/path/to/your.file')
+   * space.createUpload(stream)
+   * .then(upload => console.log(upload))
+   */
+  function createUpload (data) {
+    return httpUploads.post('uploads', data, {
+      headers: {
+        'Content-Type': 'application/octet-stream'
+      }
+    })
+    .then((response) => wrapEntry(http, response.data), errorHandler)
+  }
+
   return {
     delete: deleteSpace,
     update: updateSpace,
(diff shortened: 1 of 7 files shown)


lib/create-space-api.js
```diff
@@ -206,7 +206,8 @@ import omit from 'lodash/omit'
  * @return {ContentfulSpaceAPI}
  */
 export default function createSpaceApi ({
-  http
+  http,
+  httpUploads
```
Khaledgarbaya: any reasons you are passing another http client here ?

lib/create-space-api.js
```diff
@@ -796,6 +797,40 @@ export default function createSpaceApi ({
...
+   * space.createUpload(stream)
+   * .then(upload => console.log(upload))
+   */
+  function createUpload (data) {
+    return httpUploads.post('uploads', data, {
+      headers: {
+        'Content-Type': 'application/octet-stream'
+      }
+    })
+    .then((response) => wrapEntry(http, response.data), errorHandler)
```
Khaledgarbaya: this should be `wrapUpload`

Khaledgarbaya:
@axe312ger maybe the instead of createUpload as method of the space we can have an `upload` method in the asset object, Also maybe we should consider more helper methods in the asset object like `Space::createAssetFromUpload('UPLOAD_ID')` or something, what do you think ?

lib/create-space-api.js
axe312ger: That was the only way I could find to nicely use two API endpoints. Since the httpClient has to know the Space ID and I cant simply overwrite the host in the Upload related HTTP requests via the axios request config parameter.

axe312ger:
> @axe312ger maybe the instead of createUpload as method of the space we can have an upload method in the asset object

But the specs for the file upload say's, we have to have an upload (id) first and then create an asset out of it. I am not sure how smart it is to turn the logic in the SDK around. But I am very open to discussion since I had to do this PR without exchanging anything with u 🙈 

> Also maybe we should consider more helper methods in the asset object like Space::createAssetFromUpload('UPLOAD_ID') or something, what do you think ?

The `createAssetFromUpload` method is missing in this PR, first wanted to find a structure I am happy to go for. Then implement all of it :)

lib/create-space-api.js
Khaledgarbaya: Can't you just swap configs whenever you need them  https://github.com/mzabriskie/axios#config-defaults  

Maybe a util function like `swapConfig` and when creating the client we save our two configs for upload and for normal API request. this way we won't have another axios instance, and maybe in the future the direct file Upload will be rate-limited and it is gonna be a bit hard to control the rate of requests per period of time for two axios instances

Khaledgarbaya:
The reason of having the SDK is the ease of use of the API, so if we could have some helper method and some shortcuts that make sense why not.

axe312ger:
Alright. Any more thoughts / ideas on this?

I would continue after lunch, have now access to the stage env to actually test this with the upcoming new API :)

Khaledgarbaya:
That should be it, for now, keep up the good work cheers.

lib/entities/space.js
```diff
@@ -32,7 +32,7 @@ export function wrapSpace (http, data) {
   const spaceScopedUploadHttpClient = wrapHttpClient(http.cloneWithNewParams({
     space: space.sys.id,
     // host: 'upload.contentful.com'
-    host: 'localhost:3000'
+    host: 'upload.flinkly.com'
```
Khaledgarbaya: I think should never be here.

lib/entities/space.js
```diff
@@ -29,8 +29,14 @@ export function wrapSpace (http, data) {
   const spaceScopedHttpClient = wrapHttpClient(http.cloneWithNewParams({
     space: space.sys.id
   }), httpClientParams)
+  const spaceScopedUploadHttpClient = wrapHttpClient(http.cloneWithNewParams({
+    space: space.sys.id,
+    // host: 'upload.contentful.com'
```
Khaledgarbaya: please avoid pushing commented properties, if it is for testing purpose try not to push them

PR #79, opened 2017-02-16
build(sdk core): upgrade sdk core implementation

axe312ger:
- [x] replace vendored axios version with npm version
- [ ] upgrade SDK-core version after it was released without vendored axios


axe312ger:
We will continue with this in the other PR

PR #81, opened 2017-02-22
applies the new SDK core structure

axe312ger:
The new build structure can be found in this PR: https://github.com/contentful/contentful-sdk-core/pull/10


.babelrc
```diff
@@ -1,11 +1,26 @@
 {
-  "presets": [
-    "es2015"
-  ],
   "env": {
```
Khaledgarbaya: Just a comment and an improvement in the future we should reduce the amount of `envs` here. when we review the mocking in the unit test

Khaledgarbaya:
CI error `Error: Cannot find module 'contentful-sdk-core'`

Khaledgarbaya:
which make sense since the sdk-core is not released yet 😂

Khaledgarbaya:
ok @axe312ger moment of truth I will go ahead and merge this to master

PR #83, opened 2017-02-27
fix(errors): throw non axios-errors without further processing

axe312ger:


lib/error-handler.js
```diff
@@ -40,19 +48,20 @@ export default function errorHandler (errorResponse) {
...
+    if ('message' in data) {
       errorData.message = data.message
-      if (data.details) {
-        errorData.details = data.details
-      }
     }
-    if ('sys' in data && 'id' in data.sys) {
-      errorName = data.sys.id
+    if ('details' in data) {
+      errorData.details = data.details
```
Khaledgarbaya: any reason not to do `errorData.details = data.details || {}`

lib/error-handler.js
```diff
@@ -40,19 +48,20 @@ export default function errorHandler (errorResponse) {
...
     }
   }
-
   if (isObject(data)) {
     if ('requestId' in data) {
       errorData.requestId = data.requestId
     }
-    if ('sys' in data && 'type' in data.sys) {
+    if ('message' in data) {
       errorData.message = data.message
```
Khaledgarbaya: Why not `errorData.message = data.message || ''` if no message is there we'll have an empty string which is fine, it will help to identify the errors with no messages coming from the API endpoint and maybe tell the backend to add something there

lib/error-handler.js
```diff
@@ -40,19 +48,20 @@ export default function errorHandler (errorResponse) {
...
-      }
     }
-    if ('sys' in data && 'id' in data.sys) {
-      errorName = data.sys.id
+    if ('details' in data) {
+      errorData.details = data.details
+    }
+    if ('sys' in data) {
+      if ('id' in data.sys) {
+        errorName = data.sys.id
```
Khaledgarbaya: Why the setting the `errorName` to the `data.sys.id` ?

axe312ger:
continue with this in #84

PR #84, opened 2017-02-27
feat(upload): Direct file upload

axe312ger:


lib/contentful-management.js
```diff
@@ -28,7 +28,9 @@ export default function createClient (axios, params) {
     rateLimit: 6,
     rateLimitPeriod: 1000,
     maxRetries: 5,
-    retryOnTooManyRequests: true
+    retryOnTooManyRequests: true,
+    defaultHostname: 'api.contentful.com',
+    defaultHostnameUpload: 'upload.contentful.com'
```
axe312ger: I wonder if and how we could move this to the SDK-Core

lib/create-space-api.js
```diff
@@ -513,6 +514,81 @@ export default function createSpaceApi ({
...
+   * @param {object} data.fields.file.[LOCALE].file - Can be a string, an ArrayBuffer or a Stream.
+   * @return {Promise<Asset.Asset>} Promise for the newly created Asset
+   * @example
+   * space.createAssetFromFiles({
+   *   fields: {
+   *     file: {
+   *       'en-US': {
+   *          contentType: 'image/jpeg',
+   *          fileName: 'filename_english.jpg',
+   *          file: readFileSync('path/to/filename_english.jpg')
```
johanneswuerbach: I wouldn't really encourage that as in general you don't want to buffer the entire file in memory, but instead use `createReadStream`

axe312ger: Axios accepts multiple formats, but yes, the example should use a less memory consuming version :) 👍

package.json
```diff
@@ -66,7 +66,7 @@
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-rewire": "^1.0.0",
     "babel-plugin-transform-object-assign": "^6.22.0",
-    "babel-plugin-transform-runtime": "^6.7.5",
+    "babel-plugin-transform-runtime": "^6.23.0",
```
Khaledgarbaya: any reason we have this back, we were only using it for `Promise` polyfill and we replaced it with `es6-promises`

lib/create-space-api.js
```diff
@@ -513,6 +514,81 @@ export default function createSpaceApi ({
...
+   * Creates a Upload.
+   * @memberof ContentfulSpaceAPI
+   * @param {object} locale - The locale of the file
+   * @param {object} data - Can be a string, an ArrayBuffer or a Stream.
+   * @return {Promise<Object>} File field value for the given locale.
+   * @example
+   * const fileToUpload = readFileSync('path/to/filename_english.jpg')
+   * space.createUpload('en-US', fileToUpload)
+   */
+  function createUpload (locale, data) {
```
Khaledgarbaya: this part assumes that all assets are localized and that is not the case, some assets are not.
can you swap the params order like `function createUpload (data, locale)` and check if the locale is undefined you do 

````js
return {
           contentType,
           fileName,
           uploadFrom: {
             sys: {
               type: 'Link',
               linkType: 'Upload',
               id
             }
           }
         }
```

lib/create-space-api.js
```diff
@@ -513,6 +514,81 @@ export default function createSpaceApi ({
...
+    const { file, fileName, contentType } = data
+    httpUpload.post('uploads', file, {
+      headers: {
+        'Content-Type': 'application/octet-stream'
+      }
+    })
+    .then((uploadResponse) => {
+      const { id } = uploadResponse.data.sys
+
+      return {
```
Khaledgarbaya: I think you are making so many changes to the original response, why not just returning the response object, because a user will expect to have an upload object and not an asset data.

lib/create-space-api.js
Khaledgarbaya: Have additional [comment](https://github.com/contentful/contentful-management.js/pull/84#discussion_r103325025) about this

lib/error-handler.js
```diff
@@ -40,19 +48,20 @@ export default function errorHandler (errorResponse) {
...
-      }
     }
-    if ('sys' in data && 'id' in data.sys) {
-      errorName = data.sys.id
+    if ('details' in data) {
+      errorData.details = data.details
+    }
+    if ('sys' in data) {
+      if ('id' in data.sys) {
+        errorName = data.sys.id
```
Khaledgarbaya: What if there is not `data.sys.id` what will be the value of errorName. you can have a default value like `UNKNOWN`

package.json
axe312ger: Old version of this PR had async/await, thats why I upgraded the dependency. But actually it should be removed completely.

lib/create-space-api.js
axe312ger: I can, just need to find a way to pass the locale/ensure the uploads end up in the correct locale.

lib/contentful-management.js
axe312ger: I will keep the naming since we are using `host` and `hostUpload` to have all hosts starting with the same characters. This ensures easier recognition of all the hosts in the passed configuration :)

lib/error-handler.js
```diff
@@ -40,19 +48,20 @@ export default function errorHandler (errorResponse) {
...
+    if ('message' in data) {
       errorData.message = data.message
-      if (data.details) {
-        errorData.details = data.details
-      }
     }
-    if ('sys' in data && 'id' in data.sys) {
-      errorName = data.sys.id
+    if ('details' in data) {
+      errorData.details = data.details
```
axe312ger: Then we may also add these defaults to the initial `errorData` object:

```
  const errorData = {
    status: response.status,
    statusText: response.statusText,
    message: '',
    details: {}
  }
```

What do u think?

lib/error-handler.js
axe312ger: We could, but thats breaking the old behaviour:

`error.name = errorName && errorName !== 'Unknown' ? errorName : `${response.status} ${response.statusText}`

axe312ger:
@Khaledgarbaya after travis is ready you can review again :)

Khaledgarbaya:
ok 🤞🏼 I will merge now

PR #89, opened 2017-03-20
feat(rateLimit): Adapt to the new ratelimit api

Khaledgarbaya:
The new sdk-core version which is [here](https://github.com/contentful/contentful-sdk-core/pull/14) will the remove the export of the ratelimit function as it will be inside the `creaHttpClient` so no need to use outside

### DIFF
--- a/lib/entities/asset.js
+++ b/lib/entities/asset.js
@@ -16,8 +16,8 @@ import {
   createArchivedChecker
 } from '../instance-actions'
 
-const ASSET_PROCESSING_CHECK_WAIT = 500
-const ASSET_PROCESSING_CHECK_RETRIES = 5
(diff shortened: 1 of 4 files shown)


lib/entities/asset.js
```diff
@@ -16,8 +16,8 @@ import {
   createArchivedChecker
 } from '../instance-actions'
 
-const ASSET_PROCESSING_CHECK_WAIT = 500
-const ASSET_PROCESSING_CHECK_RETRIES = 5
+const ASSET_PROCESSING_CHECK_WAIT = 2000
+const ASSET_PROCESSING_CHECK_RETRIES = 10
```
axe312ger: 👍 
Makes more sense and should be way more solid for bigger files

axe312ger:
LGTM

PR #103, opened 2017-05-16
feat(createClient): Add Contentful User Agent

Khaledgarbaya:
Will be changed when is merged https://github.com/contentful/contentful-sdk-core/pull/21


webpack.config.js
```diff
@@ -8,6 +8,7 @@ var plugins = [
     'caching': true
   }),
   new webpack.optimize.OccurrenceOrderPlugin(),
+  new webpack.IgnorePlugin(/vertx/),
```
axe312ger: It was time for that. Still unbelievable for me that they inject a environment related dependency into that polyfill ... nonono not good... at least we got it fixed now.

PR #105, opened 2017-05-24
fix(createUpload): Setting ContentType

### LINKED ISSUE #104: Content type not set for file upload from createAssetFromFiles
I'm using the space.createAssetFromFiles method to upload PDF documents. Have submitted a ticket (15797) through your support desk. 

The issue is uploading via the api [createAssetFromFiles](https://github.com/contentful/contentful-management.js/blob/8a23aa8ca679fefd5a04eb272ecfd42b60d52498/lib/create-space-api.js#L550) does not set the content type I pass through (application/pdf), so when I open the file (the link from CDN) in a browser it downloads the file rather than opening it inside the browser. This is because the content-type is set to `application/octet-stream` rather than what `application/pdf`. 

Looking at the code here `application/octet-stream` is hardcoded in [createUpload](https://github.com/contentful/contentful-management.js/blob/8a23aa8ca679fefd5a04eb272ecfd42b60d52498/lib/create-space-api.js#L594) - only the file property from data passed through is used. 

The contentType property is passed through and should be used

samjeffress:
ContentType for createUpload was previously hardcoded to
applicaion/octet-stream
This fix uses the contentType parameter for uploading a file

Closes #104

### DIFF
--- a/lib/create-space-api.js
+++ b/lib/create-space-api.js
@@ -585,13 +585,13 @@ export default function createSpaceApi ({
    * space.createUpload(uploadStream)
    */
   function createUpload (data) {
-    const { file } = data
+    const { file, contentType } = data
     if (!file) {
       return Promise.reject(new Error('Unable to locate a file to upload.'))
     }
     return httpUpload.post('uploads', file, {
       headers: {
-        'Content-Type': 'application/octet-stream'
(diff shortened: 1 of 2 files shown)


samjeffress:
the build looks like it had an issue running tests in windows 10 - i don't _think_ it was my changes that broke it :)

PR #106, opened 2017-05-29
feat(errHandler): make token obfuscation more obvious

Khaledgarbaya:
instead of `ACCESS_TOKEN` we show the last 5 characters of the token and pre-pend `...`


axe312ger:
i like it! <3

PR #111, opened 2017-06-13
[WIP] Refactor/migrate to jest

Khaledgarbaya:
Why this PR?

* Blue tape is very very limited interms of testing and it is very old
* Jest allows snapshotting testing and commit based commit which means very very fast
So this is attempt to migrate to jest


axe312ger:
Platform is switching to mocha/chai with all their projects.

We might consider doing the same, I'd love to have some consistency here to increase the number of people who are able and willing to contribute tests.

What do you think? The syntax is pretty much the same :)

Khaledgarbaya:
@axe312ger I find jest faster than mocha, but for our case, 200 test case won't be noticed and for the sake of consistency let's do it. I will close this PR

axe312ger:
@Khaledgarbaya as long Jest won't work with browsers, I guess we can close this PR?

Khaledgarbaya:
@axe312ger yep done

PR #112, opened 2017-06-15
fix(axios): fix proxy

Khaledgarbaya:
Axios was assuming always that the proxy protocol is the same as the URL it is requesting. This a problem when you use a http proxy when requesting data from https endpoint.

This is a temporary fix until axios maintainer merges this [PR](https://github.com/mzabriskie/axios/pull/959)

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "tonic-example.js"
   ],
   "dependencies": {
-    "axios": "contentful/axios#https-via-http-proxy",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "tonic-example.js"
   ],
   "dependencies": {
-    "axios": "~0.15.3",
+    "axios": "contentful/axios#https-via-http-proxy",
```
zcei: The branch is called `fix/https-via-http-proxy`

PR #118, opened 2017-07-10
feat: Add raw request method to client

jfr3000:


lib/create-contentful-api.js
```diff
@@ -141,10 +141,15 @@ export default function createClientApi ({ http }) {
     .then((response) => wrapOrganizationCollection(http, response.data), errorHandler)
   }
 
+  function rawRequest (opts) {
```
Khaledgarbaya: one more thing, can you add the function doc 🤓 . You can copy paste from other functions and change accordingly

axe312ger:
Shouldn't we add this feature also to https://github.com/contentful/contentful.js?

Khaledgarbaya:
@axe312ger maybe at some point but it's not high prio

PR #122, opened 2017-07-14
House keeping + Webpack 3

axe312ger:
Gives minor bundle size improvements plus potentual execution time improvements.

Also, cleans up the webpack config and separates the babel-preset-env config for browser and node


axe312ger:
Most relevant changes are in https://github.com/contentful/contentful-management.js/pull/122/commits/ec04c85b19c05db41a5268c5ed863e47254694ae

Khaledgarbaya:
@axe312ger can we avoid pushing changes related to linting next time, to make the PR more focus

webpack.config.js
```diff
@@ -31,44 +30,56 @@ if (process.env.NODE_ENV === 'production') {
...
+          exclude: /(node_modules|bower_components|dist)/,
+          loader: 'babel-loader',
+          options: {
+            env: 'browser'
+          }
+        }
+      ]
     },
-    devtool: 'source-map',
+    devtool: PROD ? false : 'source-map',
```
Khaledgarbaya: I think this will cause a crash when the user try to build his app, because of `es6-promise` source maps

axe312ger: Why? This should stripe out source maps. IMHO a prod release should not contain any source maps, it is just clutter which extends the bundle size by 2

axe312ger: Also I had (accidentally) the es6-promise the the bundle with this setting and it was working fine? (At least bundeling and tests went green)

PR #127, opened 2017-07-24
Next major version

axe312ger:
* [x] Upgrade dependencies
* [x] Reduce bundle size by removing/replacing dependencies
* [x] Improve dev tooling
* [x] Improve test tooling
* [x] Add snapshot endpoint
* [x] Introduce a legacy version for IE11 and other older browsers (instead of forcing user to have a promise polyfill)
* [x] Introduce an es-modules version
* [x] Mention in README how to effectively use webpack with our library
* [x] Revisit the few broken (?) integration tests
* [x] Find an alternative to [platform.js](https://github.com/bestiejs/platform.js) since it is huge and breaks react-native


axe312ger:
We are getting there. Here example screenshots from https://github.com/contentful-labs/file-upload-example with the new version

## Before
screen shot 2017-07-27 at 17 35 25

## After
screen shot 2017-07-27 at 17 30 55

## Conclusion:
Well, it reduces the size from 1.18mb to 1.13mb. That is not a lot, why?

1. The example is using lodash, the SDK not anymore. This causes some duplication. We might consider putting lodash back but use it only very limited
2. Buffer got in again, we need to provide some information how to avoid this and other issues. (When using webpack)
3. Webpack by default uses the `browser` field  the package.json. Since we use this one at the moment for the browser version, this might be an issue. Can be easily fixed by changing the `resolve.mainfields` config in webpack but this might not be ideal for everybody
4. the file-upload-example should be optimized. I am pretty sure we can get rid of some dependencies there also

I will investigate further :)

README.md
```diff
@@ -64,20 +67,26 @@ yarn add contentful-management
 If you'd like to use a standalone built file you can use the following script tag or download it from [unpkg](https://unpkg.com), under the `dist` directory:
 
 ``` html
-<script src="https://unpkg.com/contentful-management@latest/dist/contentful-management.min.js"></script>
+<script src="https://unpkg.com/contentful-management@latest/dist/contentful-management.browser.min.js"></script>
```
Khaledgarbaya: why Changing the name?

README.md
```diff
@@ -109,6 +118,56 @@ var client = contentful.createClient({
...
+
+We are constantly working on smaller bundle sizes for our customers to allow them to build faster and less bandwidth consuming apps. Since version 4 the contentful-management SDK is exported as modules version, also. This version is automatically picked by webpack to enable tree-shaking.
+
+For further optimization of your app size, we recommend the following:
+
+### Step 1: Transform generic lodash requires to cherry-picked ones:
+
+This can be achieved via [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) and can already help to reduce the bundle size.
+
+### Step 2: Remove unnecessary lodash functionality:
```
Khaledgarbaya: This is out of the scope of the library and it should not be part of the README. It is a common knowledge

README.md
```diff
@@ -148,7 +207,7 @@ You can try and change the above example at [Tonic](https://tonicdev.com/npm/con
 
 - **I can't Install the package via npm**
 	- Check your internet connection
-	- It is called `contentful-management` and not `contenful-management` ¯\_(ツ)_/¯
+	- It is called `contentful-management` and not `contenful-management` ¯\\\_(ツ)_/¯
```
Khaledgarbaya: can you bring back my `¯\_(ツ)_/¯`

karma.conf.saucelabs.js
```diff
@@ -24,6 +24,14 @@ var browsers = {
     browserName: 'MicrosoftEdge',
     platform: 'Windows 10',
     version: '13.10586'
+  },
+  sl_android: {
```
Khaledgarbaya: Why do we need this, either we have both mobile test or we don't also Android 4.4 is very old. IMHO this is an overkill for testing since Saucelab is not that stable also

package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "contentful-management",
-  "version": "3.7.6",
+  "version": "4.0.0",
```
Khaledgarbaya: this should be `4.0.0-rc1` `4.0.0-rc2` etc to not interfere with the `@latest` versions

README.md
axe312ger: Dude, no. Your version is broken :D I fixed it for correct output ;)

README.md
axe312ger: `common knowledge` - I doubt that. Configuring webpack is still magic for most of the web developers. Sharing the knowledge with less-webpack-skilled is IMHO a good service since it helps to squeeze the most out of our libs.

karma.conf.saucelabs.js
axe312ger: Very old but still used by a lot of people. I added it for the legacy version. With that argument, you could also kill IE11, also very old but still, we should support it for enterprise/corporate customers

README.md
axe312ger: To make it clear that this is a browser target. It won't work on node thanks to weird axios bundeling behavior. Since it is a major version, renaming is not a problem.

package.json
axe312ger: sure, will take care of it.

axe312ger:
# Size update:

| version | normal | minified | minified+gzipped |
|-------------------------------|--------|----------|------------------|
| v3 (with sdk-core v3.14.1)    | 342K   | 74K      | 23.14kB          |
| v3 (with sdk-core v3.14.2)    | 306K   | 63K      | 18.47kB          |
| v4 (with sdk-core v4.0.0-rc3) | 256K   | 53K      | 14.46kB          |

## v3 with sdk-core 3.14.1:
screen shot 2017-08-10 at 13 20 27

## v4 with sdk-core 4.0.0-rc3:
screen shot 2017-08-10 at 11 47 45

README.md
Khaledgarbaya: The description is detailed and out of the scope of the library, it should not be here

axe312ger:
With `contentful-management@4.0.0-rc2` the [file-upload-example](https://github.com/contentful-labs/file-upload-example) app is now down to `903.36KB` which is already a good step.

Ofc it needs more optimization, especially a vendor chunk is needed. But for now, I am pretty happy that I could effectively reduce the size of the app :)

screen shot 2017-08-11 at 11 42 06

axe312ger:
Further progress by introducing babel-preset-env v2, code splitting and lazy loading to the file-upload-example app.

Raw file size is now: 686.12KB (vendor) + 113.91KB (app) + 66.78KB (assets route) = `866.81KB`

This results in final file sizes on the disk:
assets.js: 31K
app.js: 85K
vendor.js: 233K

Which leads to 349K total size of the minified production app 👍 

screen shot 2017-08-11 at 17 42 08

Khaledgarbaya:
@axe312ger I think it's a good time to put all this knowledge in a blog post ;) cc @stefanjudis

axe312ger:
@Khaledgarbaya already in the worx

axe312ger:
@Khaledgarbaya: Please review the follwing diff:
https://github.com/contentful/contentful-management.js/compare/b5df9b58870783a34c8b85a8dba80aa0734bba3b...c41967139e9ac8482890f25e90edf09fff3d16f1

Also, rebasing should be the only thing missing here :)

axe312ger:
Rebased 💪

axe312ger:
@Khaledgarbaya I agree, we should have eslint updates separately. I just hope that the next versions of standard won't change the rules to have such insane impact as v10 did

PR #134, opened 2017-08-10
Fix docs about isUpdate, isPublished, etc.

madtrick:
Small fixes to the documentation. These make obvious that properties like `isPublished`, `isUpdated`, ... are functions


Khaledgarbaya:
@madtrick looks like this is an old version. The [docs](https://github.com/contentful/contentful-management.js/blob/master/lib/entities/asset.js#L348-L352) is correct on master

madtrick:
lol, I even pushed the code back then https://github.com/contentful/contentful-management.js/commit/a24a4dca8fc6dfedaf4436158531301141c9c110

PR #135, opened 2017-08-15
feat(endpoints): add UI Extensions endpoint

axe312ger:


lib/entities/ui-extension.js
```diff
@@ -0,0 +1,105 @@
...
+     * .then((uiExtension) => console.log(`UI Extension ${uiExtension.sys.id} updated.`))
+     * .catch(console.error)
+     */
+    update: function () {
+      const uiExtension = this
+      return createUpdateEntity({
+        http: http,
+        entityPath: 'extensions',
+        wrapperMethod: wrapUiExtension
+      }).call(uiExtension)
```
Khaledgarbaya: `.call(uiExtension)` ?

PR #139, opened 2017-09-18
feat(user): add `user/me` endpoint

Khaledgarbaya:
This PR adds the `/user/me/` endpoint to be able to get the currently authenticated user.

- [x] Unit test
- [x] Integration test ??

### DIFF
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -52,7 +52,8 @@ import entities from './entities'
  * @return {ClientAPI}
  */
 export default function createClientApi ({ http }) {
-  const {wrapSpace, wrapUser, wrapSpaceCollection} = entities.space
+  const {wrapSpace, wrapSpaceCollection} = entities.space
+  const {wrapUser} = entities.user
   const {wrapOrganizationCollection} = entities.organization
 
   /**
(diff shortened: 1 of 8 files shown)


lib/create-contentful-api.js
```diff
@@ -41,7 +41,7 @@
 import { createRequestConfig } from 'contentful-sdk-core'
 import errorHandler from './error-handler'
 import entities from './entities'
-
+import {wrapUser} from './entities/user'
```
axe312ger: so this was the solution to the import problem? :(

I have a bad feeling with having the entities imported twice, one general object with all entities as property plus wrapUser from the user separate? 👎

axe312ger: Should be at the beginning of `createClientApi` as `wrapSpace` and `wrapOrganizationCollection`.

I can give it a try? 🤷‍♂️

Khaledgarbaya: I won't merge the PR yet until I find out what is going on with the entities

axe312ger:
We should add tests for https://coveralls.io/builds/13334564/source?filename=lib%2Fcreate-contentful-api.js#L156

axe312ger:
But shouldn't https://github.com/contentful/contentful-management.js/pull/139/commits/2698e448659ece3a690291ddce2c625eca07f51d#diff-f8a01c6833d232cf4efc7d4e518a1d31 cover it? 🤔

Khaledgarbaya:
That was integration test

axe312ger:
@Khaledgarbaya rebase/squash and then commit? LGTM

PR #148, opened 2017-10-29
Fix code example in readme.md

sbezludny:
Fixed variable name in `README > Using ES6 import` to make code example usable.


axe312ger:
thx

PR #174, opened 2018-06-11
improve deprecation message

axe312ger:
Turns 

```
Deprecated: Space.getContentTypes() will be removed in version 6.0. Please migrate your code to use the new environments API.
```

into

```
Deprecated: Space.getContentTypes() will be removed in future major versions.

Please migrate your code to use Environment.getContentTypes():
https://contentful.github.io/contentful-management.js/contentful-management/latest/ContentfulEnvironmentAPI.html#.getContentTypes
```

### DIFF
--- a/lib/create-space-api.js
+++ b/lib/create-space-api.js
@@ -57,6 +57,18 @@ import entities from './entities'
  * @prop {function} getContentTypeSnapshots
  */
 
+function raiseDeprecationWarning (method) {
+  console.warn(
+    [
+      `Deprecated: Space.${method}() will be removed in future major versions.`,
+      null,
+      `Please migrate your code to use Environment.${method}():`,
+      `https://contentful.github.io/contentful-management.js/contentful-management/latest/ContentfulEnvironmentAPI.html#.${method}`,
+      null
+    ].join('\n')
+  )
+}
+
 /**
  * Creates API object with methods to access the Space API
  * @private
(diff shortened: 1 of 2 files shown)


lib/create-space-api.js
```diff
@@ -57,6 +57,18 @@ import entities from './entities'
  * @prop {function} getContentTypeSnapshots
  */
 
+function raiseDeprecationWarning (method) {
+  console.warn(
+    [
+      `Deprecated: Space.${method}() will be removed in version 6.0.0.`,
+      null,
+      `Please migrate your code to use the new environments API:`,
```
Khaledgarbaya: Please Change this to `Environment.${method}`

lib/create-space-api.js
```diff
@@ -57,6 +57,18 @@ import entities from './entities'
  * @prop {function} getContentTypeSnapshots
  */
 
+function raiseDeprecationWarning (method) {
+  console.warn(
+    [
+      `Deprecated: Space.${method}() will be removed in version 6.0.0.`,
```
Khaledgarbaya: Let's change this to future `major versions`

lib/create-space-api.js
axe312ger: done

lib/create-space-api.js
axe312ger: done

PR #176, opened 2018-07-13
docs(client): add correct membership to functions

axe312ger:
good catch @zeusdeux 

* adds missing 4 functions to the menu
* adds example code and more detailed docs for these 4 aswell


Khaledgarbaya:
👋 this won't trigger a rebuild for the docs, the commit message should have `fix` to trigger a new release and a doc build

axe312ger:
Yeah I know but it is not a fix to the plugin itself. We may redeploy the docs manually

PR #179, opened 2018-07-18
Feat/omit delete together

phoebeschmidt:


lib/entities/content-type.js
```diff
@@ -223,7 +227,38 @@ function createContentTypeApi (http) {
...
+
+    /**
+     * Omits and deletes a field if it exists on the contentType. This is a convenience method which does both operations at once and potentially less
+     * safe than the standard way. See note about deleting fields on the Update method.
+     * @memberof ContentType
+     * @func omitAndDeleteField
+     * @return {Promise<ContentType>} Object returned from the server with updated metadata.
+     */
+    omitAndDeleteField: function (id) {
+      this.changeField(id, 'omitted', true)
```
Khaledgarbaya: To avoid unnecessary requests, what do you think about adding a check if the field exists or not before calling `this.update`?

Khaledgarbaya: Maybe throw an error to inform the user that the field does not exist

phoebeschmidt: good call on doing a check first!

for throwing an error-- I was mimicking behavior of `update`, where if you try to update a field that doesn't exist it just does nothing. This is the first method we have that operates directly on fields though, so I could see why it makes sense to throw a specific error

Khaledgarbaya: for the `update` if the fields do not exist the API will create it if it is valid otherwise it will return an error

phoebeschmidt: ah I see my error in my test from before. thanks for checking on this

lib/entities/content-type.js
```diff
@@ -4,6 +4,7 @@
  */
 
 import cloneDeep from 'lodash/cloneDeep'
+import _ from 'lodash'
```
Khaledgarbaya: we don't need to load the entire `lodash` here because it maybe lead to an increase in bundle size.
it's better to do it like htis `import findIndex from 'lodash/findIndex'` so we can benefit from the webpack tree shaking

lib/entities/content-type.js
```diff
@@ -223,7 +228,39 @@ function createContentTypeApi (http) {
...
+
+    /**
+     * @private
+     * @param {string} id - unique ID of the field
+     * @param {string} key - the attribute on the field to change
+     * @param {string} value - the value to set the attribute to
+     * @return {Promise<ContentType>}
+     */
+    findAndUpdateField: function (id, key, value) {
+      const index = _.findIndex(this.fields, {'id': id})
```
Khaledgarbaya: any reason you are not using `find` instead? because it returns the object directly.

lib/entities/content-type.js
phoebeschmidt: 👍

lib/entities/content-type.js
phoebeschmidt: Just returning the object doesn't help me since I need to update/replace it in the list of fields. Or is there a nicer way to use `find` that i'm missing? :)

phoebeschmidt: NVM just had a temporary lapse in memory about JS evaluation (pass by reference for objects). fixing

lib/entities/content-type.js
```diff
@@ -223,7 +228,39 @@ function createContentTypeApi (http) {
...
+      if (index >= 0) {
+        const newField = this.fields[index]
+        newField[key] = value
+        this.fields.splice(index, 1, newField)
+        return this.update()
+      } else {
```
axe312ger: You won't need this else here due to the return in the line above. The parser will never reach the second return anyways.

This is called "return early" (http://blog.timoxley.com/post/47041269194/avoid-else-return-early)

PR #196, opened 2018-11-06
Feat/usages api

phoebeschmidt:
To do still:

- [x] Make sure documentation for new entities is working correctly (`usage_period` + collection, `usage` + collection)

### DIFF
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -42,6 +42,7 @@
  * @prop {function} getPersonalAccessToken
  * @prop {function} getOrganizations
  * @prop {function} rawRequest
+ * @prop {function} getUsagePeriods
  */
 
 import { createRequestConfig } from 'contentful-sdk-core'
@@ -61,6 +62,7 @@ export default function createClientApi ({ http }) {
   const {wrapUser} = entities.user
   const {wrapPersonalAccessToken, wrapPersonalAccessTokenCollection} = entities.personalAccessToken
   const {wrapOrganizationCollection} = entities.organization
+  const {wrapUsagePeriodCollection} = entities.usagePeriod
 
   /**
    * Gets all spaces
@@ -132,7 +134,7 @@ export default function createClientApi ({ http }) {
 
   /**
    * Gets a collection of Organizations
-   * @memberof ClientAPI
+   * @memberof ContentfulClientAPI
    * @return {Promise<OrganizationCollection>} Promise for a collection of Organizations
    * @example
    * const contentful = require('contentful-management')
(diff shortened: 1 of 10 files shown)


lib/create-contentful-api.js
```diff
@@ -153,6 +157,71 @@ export default function createClientApi ({ http }) {
...
+    }
+    return http.get('', {
+      baseURL,
+      headers
+    })
+      .then((response) => wrapUsagePeriodCollection(http, response.data), errorHandler)
+  }
+
+  /**
+   * Gets a collection of usages for the organization. Only available to Admin or owner of the organization.
```
phoebeschmidt: In general, I wonder if I'm giving too much context in these doc strings?

lib/create-contentful-api.js
```diff
@@ -153,6 +157,71 @@ export default function createClientApi ({ http }) {
...
+   * @memberof ContentfulClientAPI
+   * @param {string} organizationId
+   * @param {string} type - either 'organization' or 'space', type of usages to be returned
+   * @param {Object} query - Object with search params
+   * @param {string} query.filters[metric] - REQUIRED. Possible values are one of cda, cma, cpa, all_apis
```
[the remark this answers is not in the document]
phoebeschmidt: `query.filters[metric]` is also not valid jsdoc syntax and causing errors in compiling the docs 😅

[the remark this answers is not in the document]
Khaledgarbaya: the syntax for this would be 
`@param {Object<{filters:{metric: string}}>} query`

lib/create-contentful-api.js
```diff
@@ -153,6 +157,71 @@ export default function createClientApi ({ http }) {
       .then((response) => wrapOrganizationCollection(http, response.data), errorHandler)
   }
 
+  /**
+   * Gets a collection of usage periods for the organization. Only available to Admin or owner of the organization.
+   * @memberof ContentfulClientAPI
+   * @param organizationId
+   * @return {Promise<UsagePeriodCollection>} Promise for a collection of usage periods
```
Khaledgarbaya: `UsagePeriodCollection` needs to be define in the jsdoc comment on top similar to sys etc..

```
/**
 * @memberof UsagePeriodCollection
 * @typedef UsagePeriodCollection
 * @prop {number} total
 * @prop {number} skip
 * @prop {number} limit
 * @prop {Array<UsagePeriod.UsagePeriodCollection>} items
 * @prop {function(): Object} toPlainObject() - Returns this UsagePeriod collection as a plain JS object
 */
```

lib/create-contentful-api.js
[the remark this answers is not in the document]
phoebeschmidt: wow thank you i seriously was reading the manual bt obviously not closely enough

lib/create-contentful-api.js
Khaledgarbaya: Looks good enough having the code example will help the user getting more context

lib/create-contentful-api.js
phoebeschmidt: @Khaledgarbaya the definitions for these are in the entity files (`entities/usage.js` and `entities/usagePeriod.js`). Would you prefer they go into this file instead?

Also, I was previously missing the definitions of the namespace, but once I added those, links to types (Usage, UsageCollection, UsagePeriod, UsagePeriodCollection) is working now.

PR #204, opened 2019-02-08
config: remove travis config, add circle config

phoebeschmidt:
Counterpart to https://github.com/contentful/contentful.js/pull/309

Switch to circle. Do not run browser tests that require Sauce Labs for now, because of trouble with account. Keep config so we can quickly re-enable sauce lab tests in the future.


.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
+version: 2.1
+jobs:
+  unit:
+    docker:
+      - image: circleci/node:8
+    steps:
+      - run: node --version
+      - run: npm --version
```
Khaledgarbaya: Do we need `node --version` and `npm --version` here ?

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+      - run:
+          name: Build Success
+          when: on_success
+          command: ./node_modules/.bin/codecov
+  integration:
+    docker:
+      - image: circleci/node:8
+    steps:
+      - run: node --version
+      - run: npm --version
```
Khaledgarbaya: Same comment

.circleci/config.yml
phoebeschmidt: good q. I took this from a ci setup we had in another proj (maybe cli?). Fine to remove it if you also dont think it's necessary

PR #209, opened 2019-02-20
feat(environments): do not send source header if sourceEnvId is not provided

phoebeschmidt:
This PR updates the createEnvironmentWithId method to only send the source header if the source has been provided. This behavior is desired because of future feature flagging.

### DIFF
--- a/test/integration/integration-tests.js
+++ b/test/integration/integration-tests.js
@@ -161,7 +161,7 @@ test('Fails to get space', (t) => {
     })
 })
 
-test.only('Creates, updates and deletes a space', (t) => {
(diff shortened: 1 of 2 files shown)


test/integration/integration-tests.js
```diff
@@ -161,7 +161,7 @@ test('Fails to get space', (t) => {
     })
 })
 
-test('Creates, updates and deletes a space', (t) => {
+test.only('Creates, updates and deletes a space', (t) => {
```
Khaledgarbaya:
```suggestion
test('Creates, updates and deletes a space', (t) => {
```

PR #225, opened 2019-08-21
fix: Add typings folder in files section

hupe1980:
The typings are missing in the NPM package


Khaledgarbaya:
Thanks, @hupe1980,
I pushed that fixed directly to master.

PR #226, opened 2019-08-21
Feature/environment aliases

phoebeschmidt:
Add environment aliases as a new entity in the SDK.


lib/create-space-api.js
```diff
@@ -1607,6 +1657,8 @@ export default function createSpaceApi ({
     createUiExtension: createUiExtension,
     createUiExtensionWithId: createUiExtensionWithId,
     getEntrySnapshots: getEntrySnapshots,
-    getContentTypeSnapshots: getContentTypeSnapshots
+    getContentTypeSnapshots: getContentTypeSnapshots,
+    getEnvironmentAlias,
```
phoebeschmidt: style-wise, is it weird to use the ES6 convention here and not everywhere else? should I change all of them? should I make these last two match the others?

Khaledgarbaya:
Hey @phoebeschmidt I think this should be merged to the next branch.

lib/create-space-api.js
makinwab: Might be best to change all of them.

lib/entities/environment-alias.js
```diff
@@ -0,0 +1,90 @@
...
+     * .catch(console.error)
+     */
+    update: function () {
+      const environmentAlias = this
+      return createUpdateEntity({
+        http: http,
+        entityPath: 'environment_aliases',
+        wrapperMethod: wrapEnvironmentAlias,
+        headers: {'x-contentful-enable-alpha-feature': 'environment-aliasing'}
+      }).call(environmentAlias)
```
Khaledgarbaya: any reason we are calling it here?
looking at other entities we don't do that https://github.com/contentful/contentful-management.js/blob/master/lib/entities/content-type.js#L60

I might be missing something though

test/integration/environment-alias-integration.js
```diff
@@ -0,0 +1,46 @@
...
+        t.equals(updatedAlias.environment.sys.id, 'test-env')
+        // clean back up by reverting to the old aliased env
+        updatedAlias.environment.sys.id = 'previously-master'
+        return updatedAlias.update()
+      })
+      .then(twiceUpdatedAlias => {
+        t.equals(twiceUpdatedAlias.environment.sys.id, 'previously-master')
+        return space.getEnvironment('test-env')
+      })
+      .then((env) => env.delete())
```
Khaledgarbaya: I think the cleanup can be moved to an `after` callback

phoebeschmidt: @Khaledgarbaya do you mean `.finally()`? 

with some more thought, im not sure why it's important that i create a new environment in the test. that also makes the clean up easier (1 step instead of 2).

I'll rework the test and then put the cleanup in a `.finally` 👍

lib/entities/environment-alias.js
phoebeschmidt: eek good catch. i copy/pasted from `locale`, which is formulated slightly differently and uses `.call(...)`... 🤦‍♂

test/integration/environment-alias-integration.js
Khaledgarbaya: @phoebeschmidt I meant something like [this](https://github.com/contentful/contentful-management.js/blob/1211e020ad432d123eca1de9cd6809a7176205e9/test/integration/integration-tests.js#L209)

phoebeschmidt: 😯 got it. on it

PR #239, opened 2019-10-22
chore(ts): Add missing ContentFields properties

### LINKED ISSUE #238: ContentFields typings is missing some important fields
contentFields.d.ts, is missising a "validations" array and a few other fields
```
export interface ContentFields {
   id: string,
   name: string,
   required: boolean,
   localized: boolean,
   type: string
   //=> where are the validations, disabled, omitted, items, and linkType?
}
```

jroehl:
Add missing ContentFields properties to typescript definition file `contentFields.d.ts`

Resolves #238

### DIFF
--- a/typings/contentFields.d.ts
+++ b/typings/contentFields.d.ts
@@ -1,19 +1,49 @@
+export interface NumRange {
+  min: number
+  max: number
+}
+
+export interface DateRange {
+  min: string
+  max: string
+}
+
+export interface RegExp {
+  pattern: string
+  flags: string
+}
+
 export interface Validation {
-  linkContentType: string[],
+  linkContentType?: string[]
+  in?: string[]
+  linkMimetypeGroup?: string[]
+  enabledNodeTypes?: string[]
+  enabledMarks?: string[]
+  unique?: boolean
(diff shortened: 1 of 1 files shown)


typings/contentFields.d.ts
```diff
@@ -1,8 +1,49 @@
-export interface ContentFields {
-  id: string,
-  name: string,
-  required: boolean,
-  localized: boolean,
-  type: string,
+export interface NumRange {
+  min: number
```
Khaledgarbaya: I think we should allow null and not set option here `?`
because If I remember correctly you can set only min or max

typings/contentFields.d.ts
```diff
@@ -1,8 +1,49 @@
...
+  validations?: Validation[]
+}
 
+export interface ContentFields extends Item {
+  id: string
+  name: string
+  required: boolean
+  localized: boolean
+  disabled?: boolean
+  omitted?: boolean
```
realityking: This is interesting. How come `disabled` and `omitted` are nullable but `localized` and `required` are not? Is this an inconsistency in the API?

PR #243, opened 2019-11-01
fix: Added typings to the EntryProp to allow for typings on fields

snurby7:
Added typings to the EntryProps to bring it into line with the typings found in https://github.com/contentful/contentful.js/blob/9c4400a45580fd81a21121f12947e9ef8ad1ba14/index.d.ts#L80 for those of us that use the Management API to automate some tasks.

Should be compatible with previous versions as the typing isn't required but can be used if that's a way you all want to take it.


typings/entry.d.ts
```diff
@@ -18,20 +18,20 @@ export interface EntrySys extends MetaSysProps {
   publishedCounter?: number,
 }
 
-export interface Entry
-  extends EntryProp,
+export interface Entry<TEntryFields = any>
```
makinwab: Any reason why `TEntryFields` is `any` seeing as `EntryProp` has an explicit key-value type defined?

Some suggestions
- create a new type like:
```
type KVMap = {
  [key: string]: any
}
```

- update EntryProp
```
export interface EntryProp<TFields> { ...
```

- update Entry interface
```
export interface Entry<TEntryFields = KVMap> {
  extends EntryFields<TEntryFields>,
  DefaultElements<EntryFields<TEntryFields> & EntrySys>,
    MetaSys<EntrySys> {
  archive(): Promise<Entry<TEntryFields>>,
  delete(): Promise<void>,
  getSnapshot(id: string): Promise<any>,
  getSnapshots(): Promise<Collection<any>>,
  isArchived(): boolean,
  isDraft(): boolean,
  isPublished(): boolean,
  isUpdated(): boolean,
  publish(): Promise<Entry<TEntryFields>>,
  unarchive(): Promise<Entry<TEntryFields>>,
  unpublish(): Promise<Entry<TEntryFields>>,
  update(): Promise<Entry<TEntryFields>>,
}
```

typings/entry.d.ts
```diff
@@ -18,20 +18,20 @@ export interface EntrySys extends MetaSysProps {
...
   getSnapshots(): Promise<Collection<any>>,
   isArchived(): boolean,
   isDraft(): boolean,
   isPublished(): boolean,
   isUpdated(): boolean,
-  publish(): Promise<Entry>,
-  unarchive(): Promise<Entry>,
-  unpublish(): Promise<Entry>,
-  update(): Promise<Entry>,
+  publish<TFieldType = {}>(): Promise<Entry<TFieldType>>,
```
makinwab: I think the `publish` method returns the Entry based on `TEntryFields`. So, we can take out the generic template which defaults to an empty object.

You can reference my previous comment on the Entry interface for what this would look like.

typings/entry.d.ts
snurby7: my understanding was that if it wasn't an `any` it wouldn't be backwards compatible, but I  like your type suggestion, so I can certainly update that one.

snurby7:
@makinwab I updated it per your comments. 

Instead of doing this though
``` type KVMap = {
  [key: string]: any
}
```

I used a `Record<string, any>` which is effectively the same thing I believe.

typings/entry.d.ts
```diff
@@ -18,20 +18,20 @@ export interface EntrySys extends MetaSysProps {
...
-export interface Entry
-  extends EntryProp,
-    DefaultElements<EntryProp & EntrySys>,
-    MetaSys<EntrySys> {
-  archive(): Promise<Entry>,
+export interface Entry<TEntryFields = KeyValueMap>
+  extends EntryProp<TEntryFields>,
+  DefaultElements<EntryProp<TEntryFields> & EntrySys>,
+      MetaSys<EntrySys> {
+  archive<TFieldType>(): Promise<Entry<TFieldType>>,
```
TimBeyer: I think you can drop the generic `TFieldType` entirely here, and pass through the `TEntryFields` from the outer scope so that an entry of a certain type will always yield updated versions of the same type.
Otherwise, the type hints for the Entry do not propagate to its updated versions and you need to redefine the `FieldType` explicitly.

typings/environment.d.ts
```diff
@@ -23,8 +23,8 @@ export interface Environment extends DefaultElements<EnvironmentProps>, MetaSys<
   createAssetWithId(id: string, data: AssetProps): Promise<Asset>,
   createContentType(data: ContentTypeProps): Promise<ContentType>,
   createContentTypeWithId(id: string, data: ContentTypeProps): Promise<ContentType>,
-  createEntry(contentTypeId: string, data: EntryProp): Promise<Entry>,
-  createEntryWithId(contentTypeId: string, id: string, data: EntryProp): Promise<Entry>,
+  createEntry<TFieldType = any>(contentTypeId: string, data: EntryProp): Promise<Entry<TFieldType>>,
```
TimBeyer: Those `= any` will override the defaults of the `KeyValueMap` type if not specified.
Probably should put the `KeyValueMap` as a default instead.

suevalov:
@snurby7 Thanks for your PR, it definitely makes sense to improve this typings. We have started migrating the repo to Typescript and if you don't mind we'll grab this code and into a different branch and close your PR. The new version of CMA with a first-class Typescript typings is coming!

snurby7:
Hey thanks for the reply! No problem whatsoever, can't wait for that typescript support! 👍

Issue #246, opened 2020-01-13
getSpaces() interface missing parameter


nickgerig:
//typings/contentfulManagement.d.ts
ClientAPI.getSpaces(): Promise<Collection<Space>>

We cannot change query params as no parameter is defined in the interface.

ruderngespra:
Hi, thank you for raising this issue.
`getSpaces()` actually does not take any parameters.
Let us know if you have any other questions!

nickgerig:
Hi,

There is a default `limit` on the query of 25 items, whereas we have over 50 so I can't see how else we would query for items 26-50?

The underlying code supports standard Contentful query params. So for example I can do the following:

` //@ts-ignore
    const { total, skip, limit, items } = await client.getSpaces({ limit: 100 })`

ruderngespra:
Hi, thank you very much for pointing that out! I will update the `clientSpaces` method so that its interface supports these standard contentful query parameters.

ruderngespra:
I updated the typechecks so that `getSpaces()` should accept limit and skip as optional params now.

PR #248, opened 2020-01-20
fix(types): Add typecheck for getSpaces query params skip and limit

ruderngespra:
Make the params `limit` and `skip` available for TypeScript users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     },
     {
       "path": "./dist/contentful-management.browser.min.js",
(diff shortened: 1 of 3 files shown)


ruderngespra:
This pr is supposed to be a quick (not yet exhaustive) fix for #246 .
I had to change one test, because axios changed their handling of config.baseURL and config.url.
See e.g. https://github.com/axios/axios/issues/2365

PR #250, opened 2020-02-05
fix: ensure to return the most recent version when processing multiple asset locales

TimBeyer:
Currently when an asset has multiple locales, we trigger processing for them all individually in parallel, but when they finish we only return the result of the first triggered task.

This leads to a race condition where sometimes the version number is off, and subsequent publishing fails.

Change this logic to always return the last successfully processed asset state, which should then have the correct version for publishing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "contentful-management",
-  "version": "5.2.1",
+  "version": "5.12.0",
   "description": "Client for Contentful's Content Management API",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "contentful-management",
-  "version": "5.2.1",
+  "version": "5.12.0",
```
TimBeyer: I guess this is now maintained by semantic release, but I needed to bump this so that installing from git works correctly.

PR #256, opened 2020-02-20
add a way to translate entry object to SDK object

adamszeptycki:
Using Contentful SDK with ReactJS app has one problem.
We shouldn't store objects in React State (or Redux).
It would be nice to be able to get an Entry and store it as `plainObject` in the React state. But later be able to call `entry.update()`

To achieve that I would like to suggest adding an option of reconstructing SDK object with all the goodies (methods like update, delete and archive).


ruderngespra:
Hey, thank you very much for the contribution, this seems like a handy method to have and a very sensible use case!

As far as I understand, you want to be able to have the following workflow:

``` js

  space.getEntry('entryId').then(entry => {
    
    // Build a plainObject in order to make it usable for React:
    const plainObject = entry.toPlainObject();
    
    // The entry is being updated in some way as plainObject:
    const updatedPlainObject = {
      ...plainObject,
      fields: {
        ...plainObject.fields,
        title: {
          'en-US': 'updatedTitle'
        }
      }
    };
    
    // Rebuild an sdk object out of the updated plainObject:
    const entryWithMethodsAgain = space.getEntryFromData(updatedPlainObject);
    
    // Update with help of the sdk method:
    entryWithMethodsAgain.update();
    
  });

```

A few questions / comments:

- We are [deprecating](https://github.com/contentful/contentful-management.js/blob/master/lib/create-space-api.js#L62) methods that live on the space level. So I would suggest adding `getEntryFromData` and `getAssetFromData` to the environment api instead. Would you see any problems with that?

- Could you add a short documentation similar to the ones for other methods?

- Could you add simple [unit tests](https://github.com/contentful/contentful-management.js/blob/master/test/unit/create-environment-api-test.js) for success and failure cases of the two new methods?

For more information about docs and testing, see [Contributing](https://github.com/contentful/contentful-management.js/blob/master/CONTRIBUTING.md)

adamszeptycki:
@ruderngespra Thank you for delivering this awesome product!

As for the workflow you are almost correct:

```
function ExampleComponent() {
const [entry, setEntry] = React.useState<object>()

const loadData = React.useCallback( async () => {
    const plainObject = entry.toPlainObject();
// I want to store plain object (something that is serializable) in react state (in future in Redux)
    setEntry(plainObject) },
[setEntry])

const update = React.useCallback(() => {
    // Rebuild an sdk object out of the (possibly updated) plainObject (from state or Redux):
    const entryWithMethodsAgain = space.getEntryFromData(entry);
    
    // Update with help of the sdk method:
    entryWithMethodsAgain.update();
}, [entry])

React.useEffect(() => {
    loadData() // load data on mount
},[loadData])

return <button onClick={update}>Update</button>
}
```

> We are deprecating methods that live on the space level. So I would suggest adding getEntryFromData and getAssetFromData to the environment api instead. Would you see any problems with that?
I don't see a problem (usage wise on our end).
But I think it will not create correct url for all the methods (space ID is in update URL path) - I may be wrong here but I trust you know it would work correctly.


I will work on documentation and tests now.

adamszeptycki:
@ruderngespra I added docs and test, but only for success case, I don't if there is possible to write test for failing path.

adamszeptycki:
@ruderngespra  any feedback for this PR?

ruderngespra:
Hey @adamszeptycki, I am really sorry for the delay! I will address the new commits properly next week!

ruderngespra:
Hey, @adamszeptycki, thank you again for your contributions.

I think the tests and docs looks great!

But I don't fully understand your worries regarding switching to the environment api, maybe you can explain them a bit more? Are you saying that if we implemented `getEntryFromData` and `getAssetFromData` somewhere around [here](https://github.com/adamszeptycki/contentful-management.js/blob/c3d5c3370a448cee8dc5dd6af7939d7959ff4793/lib/create-environment-api.js#L900) in [create-environment-api.js](https://github.com/adamszeptycki/contentful-management.js/blob/c3d5c3370a448cee8dc5dd6af7939d7959ff4793/lib/create-environment-api.js) instead of where you suggested [create-space-api.js](https://github.com/adamszeptycki/contentful-management.js/blob/c3d5c3370a448cee8dc5dd6af7939d7959ff4793/lib/create-space-api.js), the methods of the objects that `getEntryFromData` and `getAssetFromData` return (like `update` for example) will not work properly anymore?

Could you point me maybe towards the update url path you are referring to?

adamszeptycki:
@ruderngespra I merged your suggestion and it works ok.
I've tried that with our project and it works perfectly 👍 
Thanks for feedback and awesome libreary.

Is there anything else you would like to see in this PR before merging?

ruderngespra:
Thank you, everything looks great! I just sent you a one-line-pr to update our max bundle size. (https://github.com/adamszeptycki/contentful-management.js/pull/2) As soon as this is merged, I will squash and merge your pr to master.

ruderngespra:
All merged after some more last minute hickups! Sorry again for the delay and thank you again very much for this valuable contribution!

PR #282, opened 2020-03-24
Add support for app framework

giotiskl:
Worked on by me and @FLoppix 

Target:

- [x] Turn `Organization` into a domain object
- [x] Add support for app definitions management (org level)
- [x] Add support for app installations management (env level)
- [x] Update `getOrganizations` to return instances of the our domain model instead of a JSON dump (which in the end was not a breaking change as initially thought)
- [x] Update TS definitions


lib/entities/app-definition.js
```diff
@@ -0,0 +1,100 @@
...
+     * .then((space) => space.getAppDefinition('<ui_extension_id>'))
+     * .then((appDefinition) => {
+     *   appDefinition.extension.name = 'New App Definition name'
+     *   return appDefinition.update()
+     * })
+     * .then((appDefinition) => console.log(`App Definition ${appDefinition.sys.id} updated.`))
+     * .catch(console.error)
+     */
+    update: createUpdateEntity({
+      http: http,
```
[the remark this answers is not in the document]
giotiskl: we can, but they are not used anywhere else in the code so we didn't for stylistic consistency

lib/create-contentful-api.js
```diff
@@ -63,9 +64,61 @@ export default function createClientApi ({ http }) {
...
+   * const client = contentful.createClient({
+   *   accessToken: '<content_management_api_key>'
+   * })
+   *
+   * client.getOrganizations()
+   * .then(result => console.log(result.items))
+   * .catch(console.error)
+   */
+  function getOrganizations () {
+    const baseURL = http.defaults.baseURL.replace('/spaces/', '/organizations/')
```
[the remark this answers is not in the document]
giotiskl: you can't... because the default base URL has `spaces` hardcoded in it, so you'd end up with `contentful.com/spaces/organizations`

This pattern just follows how we deal with creating personal access tokens, there they replace spaces for `/users` as well (see `lib/create-contentful-api.js` -> `createPersonalAccessToken`)

lib/create-environment-api.js
```diff
@@ -924,6 +973,8 @@ export default function createEnvironmentApi ({
     getUiExtensions: getUiExtensions,
     createUiExtension: createUiExtension,
     createUiExtensionWithId: createUiExtensionWithId,
+    getAppInstallation: getAppInstallation,
```
jelz: So we're missing `createAppInstallation(appDefinitionId, parameters)` method

typings/meta.d.ts
```diff
@@ -1,6 +1,7 @@
 export interface MetaSysProps {
   type: string,
   id: string,
+  appDefinition?: { sys: MetaLinkProps }
```
jelz: I don't think we should be adding it to the sys type. I'd extend.

lib/create-contentful-api.js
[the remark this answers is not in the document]
phoebeschmidt: true, we'd have to go into sdk-core and update the http client to allow for creating an org-scoped http client instead of space scoped.

I'm fine with this workaround (even prefer it)

lib/entities/app-definition.js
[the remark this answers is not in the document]
phoebeschmidt: I think we should start to slowly transform things as we maintain the sdk (we will never have a chance to re-write everything).

So especially in new files like "app-definition" I think it's more readable to use shorthands, string templates, etc. For me that's more important than stylistic consistency at this point.

typings/meta.d.ts
giotiskl: @FLoppix

lib/create-environment-api.js
giotiskl: @FLoppix

lib/create-organization-api.js
```diff
@@ -48,11 +49,8 @@ export default function createOrganizationApi ({
    * .catch(console.error)
    */
   function getOrganizationMembership (id) {
-    return http.get('organization_memberships/' + id, {
-      baseURL,
-      headers
-    })
-      .then(response => wrapOrganizationMembership(http, response.data, orgId), errorHandler)
+    return http.get('organization_memberships/' + id, { headers })
+      .then(response => wrapOrganizationMembership(http, response.data, this.sys.id), errorHandler)
```
phoebeschmidt: `this.sys.id` isn't correct actually, it's meant to be the organization id.

but since you made the intelligent move to simply pass an org-scoped http client into the organization API, could you please do a similar refactor to the organizationMembership entity and createOrganizationMembershipApi? basically removing the need to pass in an org id, since the http client it receives will already be org scoped (thanks to your smart refactor :) )

lib/entities/organization.js
```diff
@@ -13,26 +13,35 @@ import enhanceWithMethods from '../enhance-with-methods'
...
-
 export function wrapOrganization (http, data) {
-  const organization = toPlainObject(cloneDeep(data))
-
-  const organizationApi = createOrganizationApi({
-    http,
-    orgId: organization.sys.id
+  const org = toPlainObject(cloneDeep(data))
+  const baseURL = http.defaults.baseURL.replace('/spaces/', '/organizations/') + org.sys.id + '/'
+  const orgScopedHttpClient = http.cloneWithNewParams({ baseURL })
```
phoebeschmidt: nice 👍

test/unit/create-organization-api-test.js
```diff
@@ -29,14 +29,64 @@ function teardown () {
...
   })
 })
 
+test('API call getOrganizationMembership', (t) => {
+  t.plan(1)
+  const {api, entitiesMock} = setup(Promise.resolve({}))
+  entitiesMock.organizationMembership.wrapOrganizationMembership
+    .returns(organizationMembershipMock)
+
+  // needed as getOrganizationMembership reads the org's id from context
```
phoebeschmidt: not after the small refactor i suggested above :P

lib/entities/app-installation.js
```diff
@@ -0,0 +1,102 @@
...
+     * .then((space) => space.getEnvironment('<environment_id>'))
+     * .then((environment) => environment.getAppInstallation('<app_definition_id>'))
+     * .then((appInstallation) => {
+     *    appInstallation.parameters.someParameter = 'New Value'
+     *    return appInstallation.update()
+     * })
+     * .then((appInstallation) => console.log(`App installation ${appInstallation.sys.id} was updated`))
+     * .catch(console.error)
+     */
+    update: function () {
```
phoebeschmidt: why can't we use `createUpdateEntity` here?

test/unit/entities/app-installation-test.js
```diff
@@ -0,0 +1,65 @@
...
+  return entity.update()
+    .then((response) => {
+      t.ok(response.toPlainObject, 'response is wrapped')
+      t.equals(httpMock.put.args[0][1].name, 'updatedname', 'data is sent')
+      return {httpMock, entityMock, response}
+    })
+})
+
+test('AppInstallation update fails', (t) => {
+  return failingVersionActionTest(t, setup, {
```
phoebeschmidt: I think there's a better "failingActionTest" here that's better than failingVersionActionTest (especially if appInstallation isn't minding version numbers in the same way? did I understand that right?)

typings/environment.d.ts
```diff
@@ -44,5 +45,7 @@ export interface Environment extends DefaultElements<EnvironmentProps>, MetaSys<
   getLocales(): Promise<Collection<Locale>>,
   getUiExtension(id: string): Promise<UIExtension>,
   getUiExtensions(): Promise<Collection<UIExtension>>,
+  getAppInstallation(id: string): Promise<AppInstallation>,
+  getAppInstallations(): Promise<Collection<AppInstallation>>,
```
phoebeschmidt: are you also going to add createAppInstallation? It's unclear to me from Jakub's comment above

test/unit/entities/organization-test.js
```diff
@@ -1,28 +1,38 @@
...
 }
 
 test('Organization is wrapped', (t) => {
-  entityWrappedTest(t, setup, {
-    wrapperMethod: wrapOrganization
-  })
+  setup()
+  const wrappedOrg = wrapOrganization(httpMock, organizationMock)
+  t.looseEqual(wrappedOrg.toPlainObject(), organizationMock)
+  t.equal(httpMock.cloneWithNewParams.args[0][0].baseURL, 'http://foo.bar/organizations/id/', 'adjust the baseURL to match organizations')
```
phoebeschmidt: I'll say it again: pleased about this update :D

test/unit/entities/app-installation-test.js
giotiskl: @FLoppix

typings/environment.d.ts
giotiskl: @FLoppix

lib/entities/app-installation.js
FLoppix: Because createUpdateEntity expects an `id` on the root sys object. However the payload of an `appInstallation` does not have an `id` on the `sys` object. It uses the `appDefinition id` which is deeper nested inside the response. So the HTTP call will fail if you use the `createUpdateEntity`

giotiskl: @FLoppix

typings/environment.d.ts
FLoppix: Yes we will !

lib/entities/app-definition.js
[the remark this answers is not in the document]
giotiskl: Done

lib/create-contentful-api.js
```diff
@@ -63,9 +64,61 @@ export default function createClientApi ({ http }) {
...
+   *   accessToken: '<content_management_api_key>'
+   * })
+   *
+   * client.getOrganization('<org_id>')
+   * .then((org) => console.log(org))
+   * .catch(console.error)
+   */
+  function getOrganization (id) {
+    const baseURL = http.defaults.baseURL.replace('/spaces/', '/organizations/')
+    return http.get(baseURL)
```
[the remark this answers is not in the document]
giotiskl: Done

lib/create-organization-api.js
giotiskl: `this.sys.id` was actually correct, as this is `create-organization-api` the `.getOrganization` method would be invoked with the org object as the `this` context so it'd refer to the org's id.

However, what I didn't realize is that there's no need to even refactor anything but rather just remove orgId from Organization Membership API, the HTTP client we're passing down is already scoped to the org itself as it is the `create-organization-api` after all. So I simplified it and everything seems to be working great!

lib/entities/organization.js
giotiskl: thanks :dancer:

lib/create-organization-api.js
phoebeschmidt: that's exactly what I meant, cool thanks for doing that!

lib/create-environment-api.js
```diff
@@ -32,6 +32,8 @@ import entities from './entities'
  * @prop {function} getUiExtensions
  * @prop {function} createUiExtension
  * @prop {function} createUiExtensionWithId
+ * @prop {function} getAppInstallation
+ * @prop {function} getAppInstallations
```
phoebeschmidt: we're missing `createAppInstallation` here :)

lib/create-environment-api.js
```diff
@@ -831,6 +834,75 @@ export default function createEnvironmentApi ({
...
+   * @return {Promise<AppInstallation.AppInstallation>} Promise for an App Installation
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * const client = contentful.createClient({
+   *   accessToken: '<content_management_api_key>'
+   * })
+   *
+   * client.getSpace('<space_id>')
+   * .then((space) => space.getEnvironment('<environment-id>'))
```
phoebeschmidt: some formatting in the dev docs is not required, but nice

lib/create-environment-api.js
giotiskl: @FLoppix

lib/create-environment-api.js
giotiskl: @FLoppix

lib/create-environment-api.js
FLoppix: nice catch @phoebeschmidt  :D

PR #290, opened 2020-04-01
Hejo 1146/add space members endpoint

vbeleuta:
We want to expose two new endpoints that are being moved to GA:
[GET] /spaces/{spaceId}/space_members
[GET] /spaces/{spaceId}/space_members/{spaceMemberId}

Note: the tests are failing because the alpha header removal on GK was not deployed yet. Will not merge before those changes are live.


lib/entities/space-member.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+/**
+ * @private
+ * @param {Object} http - HTTP client instance
+ * @param {Object} data - Raw space member data
+ * @return {SpaceMember} Wrapped space member data
+ */
+export function wrapSpaceMember (http, data) {
+  const SpaceMember = toPlainObject(cloneDeep(data))
```
burakukula, line 27: just a little typing - lower case here probably. 🧐

test/integration/space-members-integration.js
```diff
@@ -0,0 +1,19 @@
...
+      t.ok(response.sys, 'sys')
+      t.ok(response.sys.type, 'Array')
+      t.ok(response.items, 'items')
+    })
+  })
+  t.test('Gets spaceMember', t => {
+    t.plan(4)
+    return space.getSpaceMember('0PCYk22mt1xD7gTKZhHycN').then(response => {
+      t.ok(response.sys, 'sys')
+      t.ok(response.sys.type, 'SpaceMember')
```
burakukula, line 14: I think you would be able to use `t.equal` in this tests.

test/integration/integration-tests.js
```diff
@@ -181,6 +182,13 @@ test('Gets space for read only tests', (t) => {
       assetReadOnlyTests(t, space)
     })
 })
+test('Gets v2 space for read only tests', (t) => {
+  return v2Client.getSpace('w6xueg32zr68')
+    .then(space => {
+      userTests(t, space)
```
burakukula, line 188: I am not sure why do you need `userTests` apart of the `spaceMembersTests` ?

lib/entities/space-member.js
vbeleuta: :D yep, thanks

test/integration/space-members-integration.js
vbeleuta: oh thanks that sounds like a good idea!

test/integration/integration-tests.js
vbeleuta: oops, I wanted to do a separate test suite for userTests and spaceMembers and leave the other one only for environment tests since sometimes they were failing when combined, so to keep them less flaky; i will remove lines 203 and 204

lib/create-space-api.js
```diff
@@ -43,6 +43,8 @@ import entities from './entities'
  * @prop {function} createRoleWithId
  * @prop {function} getSpaceUser
  * @prop {function} getSpaceUsers
+ * @prop {function} getSpaceMember
```
phoebeschmidt, line 46: Let's also add this to the typings pls :)

test/integration/space-members-integration.js
```diff
@@ -0,0 +1,19 @@
+export default function spaceMembersTests (t, space) {
+  t.test('Gets spaceMembers', t => {
+    t.plan(3)
+    return space.getSpaceMembers().then(response => {
+      t.ok(response.sys, 'sys')
+      t.ok(response.sys.type, 'Array')
+      t.ok(response.items, 'items')
```
phoebeschmidt, line 7: this can be true even if it returns nothing, right? should we check that it returns something or is that overkill

lib/create-space-api.js
vbeleuta: Oh definitely! Thanks for reminding me.

test/integration/space-members-integration.js
vbeleuta: hmm good point, I can test that returns the correct type for one of the items? could also test the length, will add

vbeleuta:
> One more thing-- typings for the new methods: `getSpaceMembers` and `getSpaceMember` need to be added to the `space.d.ts` type :)

Ooops didnt know, will do! thanks

typings/space.d.ts
```diff
@@ -76,6 +77,8 @@ export interface ContentfulSpaceAPI {
   getPreviewApiKeys(): Promise<Collection<PreviewApiKey>>,
   getWebhook(id: string): Promise<WebHooks>,
   getWebhook(): Promise<Collection<WebHooks>>,
+  getSpaceMember(id: string): Promise<SpaceMember>,
```
phoebeschmidt, line 80: 👏

phoebeschmidt:
Remember semantic release commit message format :D

vbeleuta:
> Remember squash and merge and semantic release commit message format :D

Thank you!

PR #293, opened 2020-04-03
Invitation endpoint for organization

burakukula:
We would like to add support two endpoints for invitations on the organization level:

[GET] /organization/{organizationId}/invitations/{invitationId}
[POST] /organization/{organizationId}/invitations


lib/create-organization-api.js
```diff
@@ -116,6 +117,56 @@ export default function createOrganizationApi ({ http }) {
...
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * const client = contentful.createClient({
+   *   accessToken: '<content_management_api_key>'
+   * })
+   *
+   * client.getOrganizations()
+   *  .then((organization) => organization.createInvitation({
+   *    'email': 'user.email@example.com'
```
phoebeschmidt, line 154: I think the keys do not need to be quoted here :)

lib/create-organization-api.js
```diff
@@ -192,6 +243,8 @@ export default function createOrganizationApi ({ http }) {
     getOrganizationMembership,
     getOrganizationMemberships,
     getOrganizationSpaceMembership,
-    getOrganizationSpaceMemberships
+    getOrganizationSpaceMemberships,
+    getInvitation,
```
phoebeschmidt, line 247: I think these need to be more specific-- what is it an invitation for?

Then again, `getSpaceMembershipInvitation` is maybe too long... WDYT?

phoebeschmidt: Let's also add these to the type definitions :) (I know I didn't ask for this before but I think it's a good idea)

burakukula: @phoebeschmidt thx for feedback! :) maybe I could say something like: `getOrganizationInvitation()` ? WDYT?

burakukula:
@phoebeschmidt I made changes that you mentioned, update in the namespaces, comments and type definitions. If you would have moment, could you have a look? thx a lot! 🙂

lib/create-organization-api.js
burakukula: @phoebeschmidt yeah, totally, updated, thx :)

lib/create-organization-api.js
guilebarbosa: Although I don't disagree that `getOrganizationInvitation` is a good name, it is in fact the only kind of invitation we have in Contentful. You have to invite people to your org but once they are members you can simply add them to teams or spaces.

lib/create-organization-api.js
```diff
@@ -15,6 +15,8 @@ import entities from './entities'
  * @prop {function} getAppDefinitions
  * @prop {function} getOrganizationMembership
  * @prop {function} getOrganizationMemberships
+ * @prop {function} getOrganizationInvitation
```
phoebeschmidt, line 18: @guilebarbosa made a good point about these being the only types of invitations we have in cf, but I still like this better. Sometimes the js sdk is the first way people interact with contentful, so it's good to be clear about what the invitation is for :)

phoebeschmidt: so nice job! :)

lib/create-organization-api.js
```diff
@@ -111,6 +116,56 @@ export default function createOrganizationApi ({ http }) {
...
 
+  /**
+   * Gets an Invitation in Organization
+   * @memberof ContentfulOrganizationAPI
+   * @param {string} id - Invitation ID
+   * @return {Promise<OrganizationInvitation>} Promise for a OrganizationInvitation in an organization
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganizations()
```
phoebeschmidt, line 127: this would have to be the singular `getOrganization(id)` for this example to work, I think :) (this method for getting just 1 was added by the extensibility team's PR )

lib/create-organization-api.js
```diff
@@ -111,6 +116,56 @@ export default function createOrganizationApi ({ http }) {
...
+   * @param {string} id - Invitation ID
+   * @return {Promise<OrganizationInvitation>} Promise for a OrganizationInvitation in an organization
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * const client = contentful.createClient({
+   *   accessToken: '<content_management_api_key>'
+   * })
+   *
+   * client.getOrganizations()
```
phoebeschmidt, line 151: same here, just `getOrganization(orgID)`

lib/create-organization-api.js
```diff
@@ -111,6 +116,56 @@ export default function createOrganizationApi ({ http }) {
...
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganizations()
+   * .then((organization) => organization.getOrganizationInvitation('invitation_id'))
+   * .then((invitation) => console.log(invitation))
+   * .catch(console.error)
+   */
+  function getOrganizationInvitation (id) {
+    return http.get('invitations/' + id, {
+      headers
```
phoebeschmidt, line 134: so this one needs a different header than createOrganizationInvitation?

typings/organizationInvitation.d.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface OrganizationInvitationProps {
+  firstName: string,
+  lastName: string,
+  email: string,
+  role: string,
+}
+
+export interface OrganizationInvitation extends MetaSys<MetaSysProps> {
+  organizationMembership: { sys: MetaLinkProps },
+  user: null,
```
phoebeschmidt, line 12: is `null` correct here?

lib/create-organization-api.js
burakukula: @phoebeschmidt yeah, seems like it.

typings/organizationInvitation.d.ts
burakukula: As far as I understand I think so. 
Until user not accept invitation, there is no `user` associated with this invitation and we will always get null, I think.

lib/create-organization-api.js
burakukula: on my way! 🚐

lib/create-organization-api.js
burakukula: on my way! 🚀

lib/create-organization-api.js
burakukula: We gonna remove the alpha header at the end I think from both though.

lib/create-organization-api.js
burakukula: solved 🙂

lib/create-organization-api.js
burakukula: solved 🙂

typings/organizationInvitation.d.ts
burakukula: @phoebeschmidt ok, seems like it can be object or null. Updated. Does this make sense? 🧐

lib/create-organization-api.js
burakukula: @phoebeschmidt sorry, just verified it - we would like to keep invitations in alpha for now.

typings/organizationInvitation.d.ts
guilebarbosa: This is correct.
When you create an invitation it's addressed to an e-mail address. The invitee then has to assign a user to it to be able to accept.
Part of me thinks that this could be omitted from the response but it is in fact part of the invitation object.

phoebeschmidt: but is it always null? so when you "getInvitation(id)", after the invitation has been accepted, it will be an object, won't it?

guilebarbosa: You're right @phoebeschmidt. It would be a user object then.

PR #296, opened 2020-04-03
Hejo 1296/add teams team memberships endpoints

vbeleuta:
This PR adds the following endpoints for `/organization/<orgId>/teams/:team_id/team_memberships`
- GET
- POST
- PUT
- DELETE

**TODO:**
- add createTeamMembershipByTeam
- add integration tests
- add typings

**Note for reviewers:**
Please let me know if the structure of having `*TeamMembershipByTeam` makes sense for everyone. This is a particular case and in my opinion this was the best implementation. However I look forward to your feedback. Additionally I did not use the test util functions for organization api tests because the functions needed two arguments, and again, since this is a particular case wasn't sure if it made sense to complicate the existing ones or add new ones. Thank you!

**Task:**
https://contentful.atlassian.net/browse/HEJO-1145

### DIFF
--- a/lib/create-organization-api.js
+++ b/lib/create-organization-api.js
@@ -28,6 +28,9 @@ import entities from './entities'
 export default function createOrganizationApi ({ http }) {
   const {wrapAppDefinition, wrapAppDefinitionCollection} = entities.appDefinition
   const {wrapOrganizationMembership, wrapOrganizationMembershipCollection} = entities.organizationMembership
+  const {wrapSpaceMembership, wrapSpaceMembershipCollection} = entities.spaceMembership
+
+  const headers = {'x-contentful-enable-alpha-feature': 'organization-user-management-api'}
 
   /**
    * Organization instances.
@@ -70,6 +73,47 @@ export default function createOrganizationApi ({ http }) {
       query
     })
       .then(response => wrapOrganizationMembershipCollection(http, response.data), errorHandler)
+  }
+
+  /**
+   * Gets an Space Membership in Organization
+   * @memberof ContentfulOrganizationAPI
+   * @param {string} id - Organiztion Space Membership ID
+   * @return {Promise<SpaceMembership>} Promise for a Space Membership in an organization
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganizations()
+   * .then((organization) => organization.getOrganizationSpaceMembership('organizationSpaceMembership_id'))
+   * .then((organizationMembership) => console.log(organizationMembership))
+   * .catch(console.error)
+   */
+  function getOrganizationSpaceMembership (id) {
+    return http.get('space_memberships/' + id, {
+      headers
+    })
+      .then(response => wrapSpaceMembership(http, response.data), errorHandler)
+  }
+
+  /**
+   * Gets a collection Space Memberships in organization
+   * @memberof ContentfulOrganizationAPI
+   * @param {object=} query - Object with search parameters. Check the <a href="https://www.contentful.com/developers/docs/javascript/tutorials/using-js-cda-sdk/#retrieving-entries-with-search-parameters">JS SDK tutorial</a> and the <a href="https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters">REST API reference</a> for more details.
+   * @return {Promise<SpaceMembership[]>} Promise for a Space Membership collection across all spaces in the organization
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganizations()
+   * .then((organization) => organization.getOrganizationSpaceMemberships()) // you can add queries like 'limit': 100
+   * .then((response) => console.log(response.items))
+   * .catch(console.error)
+   */
+  function getOrganizationSpaceMemberships (query = {}) {
+    return http.get('space_memberships', {
+      headers,
+      query
+    })
+      .then((response) => wrapSpaceMembershipCollection(http, response.data), errorHandler)
   }
 
   /**
@@ -146,6 +190,8 @@ export default function createOrganizationApi ({ http }) {
     getAppDefinitions,
     createAppDefinition,
     getOrganizationMembership,
-    getOrganizationMemberships
+    getOrganizationMemberships,
+    getOrganizationSpaceMembership,
+    getOrganizationSpaceMemberships
   }
 }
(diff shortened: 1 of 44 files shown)


lib/create-organization-api.js
```diff
@@ -116,6 +120,42 @@ export default function createOrganizationApi ({ http }) {
...
+   * @return {Promise<TeamMembership>} Promise for an Team Membership
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganizations()
+   * .then((organization) => organization.getTeamMembershipByTeam('teamId', 'teamMembership_id'))
+   * .then((teamMembership) => console.log(teamMembership))
+   * .catch(console.error)
+   */
+  function getTeamMembershipByTeam (teamId, teamMembershipId) {
```
phoebeschmidt, line 136: I don't like the name `getTeamMembershipByTeam`. Would there ever be another way to get a team membership? I think prob not.

let's drop the `byTeam` part from the methods

lib/create-organization-api.js
```diff
@@ -15,6 +15,8 @@ import entities from './entities'
  * @prop {function} getAppDefinitions
  * @prop {function} getOrganizationMembership
  * @prop {function} getOrganizationMemberships
+ * @prop {function} getTeamMembershipByTeam
```
phoebeschmidt, line 18: Let's also add these to the typescript definitions

lib/entities/team-membership.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+}
+/**
+ * @private
+ * @param {Object} http - HTTP client instance
+ * @param {Object} data - Raw team membership data
+ * @return {TeamMembership} Wrapped team membership data
+ */
+export function wrapTeamMembershipByTeam (http, data, teamId) {
```
phoebeschmidt, line 98: let's drop `byTeam` here too

lib/entities/team-membership.js
```diff
@@ -0,0 +1,137 @@
...
+  return freezeSys(teamMembership)
+}
+
+/**
+ * @private
+ * @param {Object} http - HTTP client instance
+ * @param {Object} data - Raw team membership collection data
+ * @return {TeamMembershipCollection} Wrapped team membership collection data
+ */
+export function wrapTeamMembershipCollectionByTeam (http, data, teamId) {
```
phoebeschmidt, line 110: and here

lib/entities/team-membership.js
```diff
@@ -0,0 +1,137 @@
...
+  return freezeSys(teamMemberships)
+}
+
+/**
+ * @private
+ * @param {Object} http - HTTP client instance
+ * @param {Object} data - Raw team membership data
+ * @return {TeamMembership} Wrapped team membership data
+ */
+export function wrapTeamMembership (http, data) {
```
phoebeschmidt, line 122: I don't see why we have "teamMembership" entity, but then we also have ways to wrap `teamMembershipByTeam` and `teamMembership`. 

Isn't the entity that's being returned always a `teamMembership`? And shouldn't it always have update and delete methods on it?

I think these wrap methods can be consolidated to just `wrapTeamMembership` and `wrapTeamMembershipCollection`, and the former will wrap the entity and create the teamMembershipApi (with update/delete)

lib/create-organization-api.js
vbeleuta: Well we also have another endpoint to be added soon which is get all `/organization/orgId/team_memberships` and for that we do not want the update and delete which is why I have two wrapper methods;
However maybe I can make these wrapTeamMemberships and for the one case of all org team memberships I will name it like that: `wrapOrganizationTeamMemberships` WDYT?

lib/entities/team-membership.js
vbeleuta: Yes, I know its a bit confusing, as I mentioned before I needed the wrapper methods because we will also have the `/organization/orgId/team_memberships` and for that we do not want the update and delete.

vbeleuta: Sorry should have also specified that in the PR description. I stared with the org team memberships but then priorities shifted.

lib/create-organization-api.js
phoebeschmidt: this whole api is very confusing to me.

I think your proposal is good: organizationTeamMemberships and teamMemberships. They're separate entities, one has update/delete methods and one returns an "organizationTeamMembership" which is readOnly.

Maybe also elaborate in the jsdoc strings a little more about this api, because for me as a reviewer, it's super confusing. (Also spacemembers, spacememberships, etc)

vbeleuta: I see, makes sense, I will separate the organizationTeamMembership.
Ill add more documentation too. Thank you.

lib/entities/team-membership.js
```diff
@@ -0,0 +1,116 @@
...
+     * .then(org => org.getTeamMembership('teamId', 'teamMembership_id'))
+     * .then((teamMembership) => {
+     *  teamMembership.delete();
+     * })
+     * .catch(console.error)
+     */
+
+    delete: function () {
+      return http.delete('teams/' + teamId + '/team_memberships/' + this.sys.id,
+        {headers: {'x-contentful-enable-alpha-feature': 'teams-api'}}
```
phoebeschmidt, line 88: only delete needs the alpha header?

test/integration/team-membership-integration.js
```diff
@@ -0,0 +1,34 @@
...
+  t.test('Create and delete teamMembership', (t) => {
+    t.plan(3)
+    return organization.createTeamMembership('7pIEx2fMx53SSR1jd7C46M', {
+      admin: true, organizationMembershipId: '5mDCHGHvePI8NOzw0WaY2Z'
+    })
+      .then((teamMembership) => {
+        t.ok(teamMembership.sys, 'sys')
+        t.equal(teamMembership.admin, true)
+        t.ok(teamMembership.sys.type, 'TeamMembership')
+        teamMembership.delete()
```
phoebeschmidt, line 31: I think we need to wait on this async function and maybe t.ok it?

The way it's written now, if it fails, your tests will have some error outputs but I don't think the test will actually "fail", if that makes sense.

lib/entities/team-membership.js
vbeleuta: yes, looks like the update does need it; in fact I will probably remove the alpha header in this pr too since was removed on GK side today

test/integration/team-membership-integration.js
vbeleuta: hmm yes I see what you mean

lib/entities/team-membership.js
vbeleuta: just fyi, not everything was deployed on GK side, so I will remove the alpha header in my next task that is also related (the getOrganizationTeamMemberships endpoint)

PR #303, opened 2020-04-08
feat(organizationTeamMemberships): add getOrganizationTeamMemberships function to organization api

vbeleuta:
Added `/organization/orgId/team_memberships` endpoint

### DIFF
--- a/lib/create-organization-api.js
+++ b/lib/create-organization-api.js
@@ -52,7 +52,7 @@ export default function createOrganizationApi ({ http }) {
    * @example
    * const contentful = require('contentful-management')
    *
-   * client.getOrganizations()
+   * client.getOrganization('organization_id')
    * .then((organization) => organization.getOrganizationMembership('organizationMembership_id'))
    * .then((organizationMembership) => console.log(organizationMembership))
    * .catch(console.error)
@@ -70,7 +70,7 @@ export default function createOrganizationApi ({ http }) {
    * @example
    * const contentful = require('contentful-management')
    *
-   * client.getOrganizations()
+   * client.getOrganization('organization_id')
    * .then((organization) => organization.getOrganizationMemberships({'limit': 100})) // you can add more queries as 'key': 'value'
    * .then((response) => console.log(response.items))
    * .catch(console.error)
@@ -90,7 +90,7 @@ export default function createOrganizationApi ({ http }) {
    * @example
    * const contentful = require('contentful-management')
    *
-   * client.getOrganizations()
+   * client.getOrganization('organization_id')
    * .then((organization) => organization.getOrganizationSpaceMembership('organizationSpaceMembership_id'))
    * .then((organizationMembership) => console.log(organizationMembership))
    * .catch(console.error)
@@ -108,7 +108,7 @@ export default function createOrganizationApi ({ http }) {
    * @example
    * const contentful = require('contentful-management')
    *
-   * client.getOrganizations()
+   * client.getOrganization('organization_id')
    * .then((organization) => organization.getOrganizationSpaceMemberships()) // you can add queries like 'limit': 100
    * .then((response) => console.log(response.items))
    * .catch(console.error)
--- a/lib/create-organization-api.js
+++ b/lib/create-organization-api.js
@@ -32,7 +32,7 @@ import entities from './entities'
 export default function createOrganizationApi ({ http }) {
   const {wrapAppDefinition, wrapAppDefinitionCollection} = entities.appDefinition
   const {wrapOrganizationMembership, wrapOrganizationMembershipCollection} = entities.organizationMembership
-  const {wrapTeamMembership, wrapTeamMembershipCollection} = entities.teamMembership
+  const {wrapTeamMembership, wrapTeamMembershipCollection, wrapOrganizationTeamMembershipCollection} = entities.teamMembership
   const {wrapSpaceMembership, wrapSpaceMembershipCollection} = entities.spaceMembership
   const {wrapOrganizationInvitation} = entities.organizationInvitation
 
@@ -133,7 +133,7 @@ export default function createOrganizationApi ({ http }) {
    *   accessToken: '<content_management_api_key>'
    * })
    *
-   * client.getOrganization('<org_id>')
+   * client.getOrganization('organizationId')
    * .then((org) => org.createTeamMembership('teamId', {
    *    admin: true,
    *    organizationMembershipId: "3ugleZJgHKk89I1P5MSDuY"
@@ -147,14 +147,15 @@ export default function createOrganizationApi ({ http }) {
   }
 
   /**
-   * Gets an Team Membership
+   * Gets an Team Membership from the team with given teamId
    * @memberof ContentfulOrganizationAPI
+   * @param {string} teamId - Team ID
    * @param {string} id - Team Membership ID
    * @return {Promise<TeamMembership>} Promise for an Team Membership
    * @example
    * const contentful = require('contentful-management')
    *
-   * client.getOrganizations()
+   * client.getOrganization('organizationId')
    * .then((organization) => organization.getTeamMembership('teamId', 'teamMembership_id'))
    * .then((teamMembership) => console.log(teamMembership))
    * .catch(console.error)
@@ -165,14 +166,14 @@ export default function createOrganizationApi ({ http }) {
   }
 
   /**
-   * Gets an Team Membership
+   * Get all Team Memberships for the team
    * @memberof ContentfulOrganizationAPI
    * @param {string} id - Team Membership ID
    * @return {Promise<TeamMembership[]>} Promise for a Team Membership Collection
    * @example
    * const contentful = require('contentful-management')
    *
-   * client.getOrganizations()
+   * client.getOrganization('organizationId')
    * .then((organization) => organization.getTeamMemberships('teamId'))
    * .then((teamMemberships) => console.log(teamMemberships))
    * .catch(console.error)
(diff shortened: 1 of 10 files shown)


lib/create-organization-api.js
```diff
@@ -142,46 +141,64 @@ export default function createOrganizationApi ({ http }) {
...
+   * @return {Promise<TeamMembership[]>} Promise for a Team Membership Collection
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganization('organizationId')
+   * .then((organization) => organization.getOrganizationTeamMemberships())
+   * .then((teamMemberships) => console.log(teamMemberships))
+   * .catch(console.error)
+   */
+  function getOrganizationTeamMemberships (query = {}) {
```
phoebeschmidt, line 197: I personally think `getOrganizationTeamMemberships` is confusing alongside `getTeamMemberships`. 

I think I'd prefer one method, `getTeamMemberships(opts)` where opts is an object that may or may not contain team ID. If it contains team ID, it gets team memberships for that team. If it doesnt, it gets team memberships for the whole org.

This is a new pattern for the SDK, but we also dont have any other entities like this, and I think it's better than having two different methods.

I'm open to reasons to keep it this way, though, if you all feel strongly.

lib/entities/team-membership.js
```diff
@@ -114,3 +110,28 @@ export function wrapTeamMembershipCollection (http, data, teamId) {
...
+  return freezeSys(teamMembership)
+}
+
+/**
+ * @private
+ * @param {Object} http - HTTP client instance
+ * @param {Object} data - Raw teamMembership data
+ * @return {TeamMembershipCollection} Wrapped TeamMembershipCollection data
+ */
+export function wrapOrganizationTeamMembershipCollection (http, data) {
```
phoebeschmidt, line 131: Can you explain to me again why the teamMemberships returned from `/organization/<orgId>/team_memberships` shouldn't have update/delete, but the teamMemberships returned from `organization/<orgId>/teams/<teamId>/team_memberships` should?

lib/create-organization-api.js
phoebeschmidt: ex: opts can include `{query: {blah}, teamId: <teamId}`

vbeleuta: I personally think they should be different because of the different wrap teamMembership methods. These are different use cases that I tried to explain with the docs and additional info in the doc-app. The endpoint `organization/orgid/team_memberships` only has [GET] implemented and there is no story to implement show, update, delete. To manage particular teamMemberships one should use `getTeamMembership(teamId, teamMembershipId)`.

lib/entities/team-membership.js
vbeleuta: The endpoint `organization/orgid/team_memberships` only has [GET] implemented and there is no story to implement show, update, delete. To manage particular teamMemberships one should use `getTeamMembership(teamId, teamMembershipId)`. https://app.stoplight.io/wk/cduzvQ7Fbbsj2mRk2/7knfjqR4Y5n8oj7hq/6vEAKeemm9QaKW2Pq/design/endpoints/phK5PdTGRPfLukiJq

lib/create-organization-api.js
vbeleuta: maybe naming can be different
`getAllOrganizationTeamMemberships()`
`getTeamMemberships(teamId)` (I think the arg and docs specifies that it is by team since it the name is not nice if getTeamMembershipsByTeam)
`getTeamMembership(teamId, teamMembershipId)`

lib/entities/team-membership.js
phoebeschmidt: I get that you cant PUT or DEL to `/organization/orgid/team_memberships`, but wont the objects returned from that API include a teamID?

From the perspective of the user of the SDK, if what I get back is a collection of team membership entities (regardless of the path from which I requested it), I would expect to be able to update/delete each of them. Plus, if the objects in the `items` returned from `/organization/orgid/team_memberships` have a teamID attached to them, we can easily wrap them using the existing method.

this is the power of the SDK :) it doesnt have to directly mimic the API design, it should be straight forward and easy to use. For me, it's confusing that there would be two types of teamMembership entities with different APIs defined on them.

lib/create-organization-api.js
phoebeschmidt: see my comment below, wrapping the same entity with a different method also seems like the wrong approach to me.

lib/entities/team-membership.js
phoebeschmidt: the only reason that I would see to keep it with a different fetching method and a different wrap method would be:
the actual items returned from `/organization/orgid/team_memberships` are so different, contain different information, do not include the team id.

vbeleuta: hmmm, yes I see what you mean by that

vbeleuta: ok let me try to update and see if the tests pass and cover the cases! thank you!

phoebeschmidt: A compromise I could live with:

The method to fetch is different, but it's still wrapped with the same wrapper function

vbeleuta: its ok, I thought the issue was the other way around :D i didnt understand the concern from before

lib/create-organization-api.js
```diff
@@ -76,9 +75,7 @@ export default function createOrganizationApi ({ http }) {
    * .catch(console.error)
    */
   function getOrganizationMemberships (query = {}) {
-    return http.get('organization_memberships', {
-      query
-    })
+    return http.get('organization_memberships', createRequestConfig({query: query}))
```
phoebeschmidt, line 78: nitpick: we can just use the "property value shorthand" here and on line 114, line 179 (we can slowly try to improve the sdk 😅 )

lib/create-organization-api.js
```diff
@@ -142,44 +137,50 @@ export default function createOrganizationApi ({ http }) {
...
-      .then(response => wrapTeamMembership(http, response.data, teamId), errorHandler)
+    return http.get('teams/' + teamId + '/team_memberships/' + teamMembershipId)
+      .then(response => wrapTeamMembership(http, response.data), errorHandler)
   }
 
   /**
-   * Gets an Team Membership
+   * Get all Team Memberships for the team
    * @memberof ContentfulOrganizationAPI
    * @param {string} id - Team Membership ID
```
phoebeschmidt, line 166: This is now "opts" config object, with optional `teamID` key :) (let's include both those pieces of info in the docstring)

lib/create-organization-api.js
```diff
@@ -142,44 +137,50 @@ export default function createOrganizationApi ({ http }) {
...
   function getTeamMembership (teamId, teamMembershipId) {
-    return http.get('teams/' + teamId + '/team_memberships/' + teamMembershipId, {headers: teamAlphaHeaders})
-      .then(response => wrapTeamMembership(http, response.data, teamId), errorHandler)
+    return http.get('teams/' + teamId + '/team_memberships/' + teamMembershipId)
+      .then(response => wrapTeamMembership(http, response.data), errorHandler)
   }
 
   /**
-   * Gets an Team Membership
+   * Get all Team Memberships for the team
```
phoebeschmidt, line 164: maybe let's also be specific here. Sth like: "Get all Team Memberships. If teamID is provided in the optional config object, it will return all Team Memberships in that team. By default, returns all team memberships for the organization."

lib/create-organization-api.js
```diff
@@ -142,44 +137,50 @@ export default function createOrganizationApi ({ http }) {
...
-   * client.getOrganizations()
+   * client.getOrganization('organizationId')
    * .then((organization) => organization.getTeamMemberships('teamId'))
    * .then((teamMemberships) => console.log(teamMemberships))
    * .catch(console.error)
    */
-  function getTeamMemberships (teamId, query = {}) {
-    return http.get('teams/' + teamId + '/team_memberships', {headers: teamAlphaHeaders, query})
-      .then(response => wrapTeamMembershipCollection(http, response.data, teamId), errorHandler)
+  function getTeamMemberships (opts = {}) {
```
phoebeschmidt, line 176: 👏

lib/entities/team-membership.js
```diff
@@ -27,7 +27,7 @@ import errorHandler from '../error-handler'
  * @prop {function(): Object} toPlainObject() - Returns the collection as a plain JS object
  */
 
-function createTeamMembershipApi (http, teamId) {
+function createTeamMembershipApi (http) {
```
phoebeschmidt, line 30: awesome simplification!

test/integration/team-membership-integration.js
```diff
@@ -1,7 +1,7 @@
 export default function teamMembershipTests (t, organization) {
   t.test('Gets teamMemberships', (t) => {
```
phoebeschmidt, line 2: another nit: maybe describe this test as being "get team memberships for team" or something like that, and the one below could be "get team memberships for org"

test/integration/team-membership-integration.js
```diff
@@ -32,4 +32,25 @@ export default function teamMembershipTests (t, organization) {
         t.pass('teamMembership was deleted')
       })
   })
+
+  t.test('Gets organizationTeamMemberships', (t) => {
```
phoebeschmidt, line 36: nit: this and the next test could also have more specific test descriptions :)

typings/organization.d.ts
```diff
@@ -10,14 +10,16 @@ export interface OrganizationProp {
...
-  getTeamMembership(teamId: string, id: string): Promise<TeamMembership>,
-  getTeamMembership(teamId: string): Promise<Collection<TeamMembership>>,
-  getOrganizationInvitation(id: string): Promise<OrganizationInvitation>,
-  createOrganizationInvitation(data: OrganizationInvitationProps): Promise<OrganizationInvitation>
+  createAppDefinition(data: AppDefinitionProps): Promise<AppDefinition>;
+  getAppDefinition(id: string): Promise<AppDefinition>;
+  getAppDefinitions(): Promise<Collection<AppDefinition>>;
+  createTeamMembership(data: TeamMembershipProps): Promise<TeamMembership>;
+  getTeamMembership(teamId: string, id: string): Promise<TeamMembership>;
+  getTeamMemberships(opts: object): Promise<Collection<TeamMembership>>;
```
phoebeschmidt, line 18: this opts object should be optional and could reference a more specific type (opts might contain `teamID` and/or `query`, right? but both optional?)

typings/teamMembership.d.ts
```diff
@@ -1,6 +1,11 @@
-import { MetaSys, MetaSysProps } from "./meta";
-import { DefaultElements } from "./defaultElements";
+import { MetaSys, MetaSysProps } from './meta'
+import { DefaultElements } from './defaultElements'
+import { QueryOptions } from './queryOptions'
 
+export interface Options {
```
phoebeschmidt, line 5: yay! nice :)

PR #308, opened 2020-04-16
feat(teamSpaceMemberships): Team space memberships endpoints

burakukula:
Adding support for teamSpaceMemberships - part of the https://contentful.atlassian.net/browse/HEJO-1148

### DIFF
--- a/lib/entities/team-space-membership.js
+++ b/lib/entities/team-space-membership.js
@@ -1,5 +1,5 @@
 /**
- * Space Membership instances
+ * Team Space Membership instances
  * @namespace SpaceMembership
  */
 
--- a/lib/entities/team-space-membership.js
+++ b/lib/entities/team-space-membership.js
@@ -21,8 +21,6 @@ import errorHandler from '../error-handler'
  */
 
 function createTeamSpaceMembershipApi (http) {
-  const teamAlphaHeaders = {'x-contentful-enable-alpha-feature': 'teams-api'}
-
   return {
 
     /**
@@ -60,7 +58,6 @@ function createTeamSpaceMembershipApi (http) {
       return http.put('team_space_memberships/' + this.sys.id, data, {
         headers: {
           'X-Contentful-Version': this.sys.version || 0,
-          'x-contentful-enable-alpha-feature': 'teams-api',
           'x-contentful-team': this.sys.team.sys.id
         }
       })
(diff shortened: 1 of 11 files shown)


lib/entities/team-space-membership.js
```diff
@@ -0,0 +1,114 @@
+/**
+ * Space Membership instances
+ * @namespace SpaceMembership
```
vbeleuta, line 3: These should be TeamSpaceMembership?

test/integration/team-space-membership-integration.js
```diff
@@ -0,0 +1,42 @@
...
+        teamMembership.roles = [
+          {
+            sys: {
+              type: 'Link',
+              linkType: 'Role',
+              id: '6YFc9mw7PREXh5FQ1lKSN6' // change role to author
+            }
+          }
+        ]
+        return teamMembership
```
vbeleuta, line 28: do we want to add a call to update the teamMembership after the role change?

lib/entities/team-space-membership.js
```diff
@@ -0,0 +1,114 @@
...
+     *        }
+     *      }
+     *    ]
+     *  })
+     *  .then((spaceMembership) => console.log(`spaceMembership ${spaceMembership.sys.id} updated.`))
+     *  .catch(console.error)
+     */
+    update: function () {
+      const raw = this.toPlainObject()
+      const data = cloneDeep(raw)
```
vbeleuta, line 56: i am not sure if it is required or not,  but last endpoint I did I had to add `delete data.sys` (after cloneDeep) to make the update endpoint work since we do not allow to update sys data I think

burakukula: @vbeleuta yeah, I think so too. fixed.

test/integration/team-space-membership-integration.js
burakukula: @vbeleuta good catch! fixed 🙂

lib/entities/team-space-membership.js
burakukula: @vbeleuta fixed 🙂

burakukula:
@vbeleuta thx a lot for feedback! 🙂I have fixed issues and add missing tests. If you would have a moment, please have a look again if everything is ok from your perspective 🙇‍♀️

PR #309, opened 2020-04-17
feat(organizationUsers): add organization users endpoints

vbeleuta:
add new organization api endpoints
[GET] /organization/orgId/users
[GET] /organization/orgId/users/userId
removed alpha header from getUser (tests will pass when GK change is deployed)

### DIFF
--- a/lib/create-organization-api.js
+++ b/lib/create-organization-api.js
@@ -15,6 +15,10 @@ import entities from './entities'
  * @prop {function} getAppDefinitions
  * @prop {function} getOrganizationMembership
  * @prop {function} getOrganizationMemberships
+ * @prop {function} createTeam
+ * @prop {function} getTeam
+ * @prop {function} getTeams
+ * @prop {function} createTeamMembership
  * @prop {function} getTeamMembership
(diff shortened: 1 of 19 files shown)


lib/create-organization-api.js
```diff
@@ -371,9 +374,48 @@ export default function createOrganizationApi ({ http }) {
...
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganization('<organization_id>')
+   * .then((organization) => organization.getUser('id'))
+   * .then((user) => console.log(user))
+   * .catch(console.error)
+   */
+  function getUser (id) {
+    return http
+      .get('users/' + id, {headers})
```
burakukula, line 392: @vbeleuta out of curiosity - what headers are here?

vbeleuta: there is a const headers defined at the beginning of the file:
const headers = {'x-contentful-enable-alpha-feature': 'organization-user-management-api'}

vbeleuta:
> looks great, only wonder about typings? 🙏

of course!! thank you for reminding me! will add right away

PR #310, opened 2020-04-20
fix(SpaceMemberships): Depracation warning for space memberships

burakukula:


burakukula:
@phoebeschmidt yeah, good idea. on my way.

burakukula:
@phoebeschmidt I updated PR, something like this would be ok from your perspective?

PR #311, opened 2020-04-23
feat(teamSpaceMembership): implement org level team space memberships

guilebarbosa:
Add support for GET requests for Team Space Membership at the organization level

### DIFF
--- a/lib/create-organization-api.js
+++ b/lib/create-organization-api.js
@@ -194,12 +194,11 @@ export default function createOrganizationApi ({ http }) {
       .then(response => wrapTeamMembershipCollection(http, response.data), errorHandler)
   }
 
-   /**
+  /**
    * Get a Team Space Membership with given teamSpaceMembershipId
    * @memberof ContentfulOrganizationAPI
-   * @param {string} teamSpaceMembershipId - Team ID
-   * @param {string} id - Team Membership ID
-   * @return {Promise<TeamMembership>} Promise for an Team Membership
+   * @param {string} teamSpaceMembershipId
+   * @return {Promise<TeamMembership>} Promise for a Team Space Membership
    * @example
    * const contentful = require('contentful-management')
    *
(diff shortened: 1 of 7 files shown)


lib/create-organization-api.js
```diff
@@ -193,6 +194,50 @@ export default function createOrganizationApi ({ http }) {
...
+   * @return {Promise<TeamMembership>} Promise for an Team Membership
+   * @example
+   * const contentful = require('contentful-management')
+   *
+   * client.getOrganization('organizationId')
+   * .then((organization) => organization.getTeamSpaceMembership('teamSpaceMembershipId'))
+   * .then((teamSpaceMembership) => console.log(teamSpaceMembership))
+   * .catch(console.error)
+   */
+  function getTeamSpaceMembership (teamSpaceMembershipId) {
```
burakukula, line 211: @guilebarbosa I am not able to comment not changed code, but you could add this function in the top of this file (line 26 or so), if you would like this ot appear in the overview of the functions the docs.

burakukula:
@guilebarbosa what do you think about adding some integration tests ?

lib/create-organization-api.js
```diff
@@ -193,6 +194,50 @@ export default function createOrganizationApi ({ http }) {
       .then(response => wrapTeamMembershipCollection(http, response.data), errorHandler)
   }
 
+   /**
+   * Get a Team Space Membership with given teamSpaceMembershipId
+   * @memberof ContentfulOrganizationAPI
+   * @param {string} teamSpaceMembershipId - Team ID
+   * @param {string} id - Team Membership ID
```
vbeleuta, line 201: i think we do not need line 201 here and in 200 we should either describe it as Team Space Membership Id or just remove the Team ID part since probably the argument is specific enough

guilebarbosa: Busted!

vbeleuta:
@guilebarbosa we also need to add the new function typings to the `organization.d.ts` then I think we are all good

lib/create-organization-api.js
guilebarbosa: Can you explain what you mean?

PR #315, opened 2020-05-13
fix: correct toPlainObject typings for content type and app definition

suevalov:
There are two errors in type definitions for `.toPlainObject` functions.

`MetaSys` scopes MetaSysProps to be in `sys` object.


suevalov:
@makinwab does it happen automatically here? or should I merge and it gets published?

makinwab:
> @makinwab does it happen automatically here? or should I merge and it gets published?

It should be released automatically, you'll have to ensure the commit message format is semantic release compliant. You can use the `Squash and merge`. @suevalov

PR #317, opened 2020-05-15
fix: allow methods to be configurable

suevalov:
Where there any good reasons why we disable `configurable` attribute for all additional methods? 

> Configurable. If false, any attempts to delete the property or change its attributes (Writable, Configurable, or Enumerable) will fail.

We're using https://overmindjs.org/ with contentful-management and it requires wrapping sdk object with `Proxy` object. 

`configurable: false` restricts us from doing this.

### DIFF
--- a/lib/enhance-with-methods.js
+++ b/lib/enhance-with-methods.js
@@ -13,7 +13,7 @@ export default function enhanceWithMethods (baseObject, methodsObject) {
   return Object.keys(methodsObject).reduce((enhancedObject, methodName) => {
     Object.defineProperty(enhancedObject, methodName, {
       enumerable: false,
-      configurable: false,
+      configurable: true,
       writable: false,
       value: methodsObject[methodName]
     })


makinwab:
> Where there any good reasons why we disable `configurable` attribute for all additional methods?

There's this part of the [comment for that method](https://github.com/contentful/contentful-management.js/blob/master/lib/enhance-with-methods.js#L4) that addresses your question, @suevalov 
>
> All the added methods are set as non enumerable, non configurable, and non writable properties. This ensures that if we try to clone or stringify the base object, we don't have to worry about these additional methods.

suevalov:
@makinwab yep, I've seen this comment but `configurable` property doesn't affect "if we try to clone or stringify the base object, we don't have to worry about these additional methods." 

`enumerable` and `writable` affect this behavior, `configurable` does not.

suevalov:
@makinwab assets integration tests seem to fail consistently across all branches. where should I  look to find the root cause? Is it an option to disable them temporarily if it's not easily fixable now?

PR #319, opened 2020-05-22
chore: allow gradual Typescript migration

suevalov:
- [x] Update to Babel 7
- [x] Add Typescript support to Babel files
- [x] Add one Typescript file and make sure that output stays the same
- [x] Make sure that types are generated correctly
- [x] Make sure unit and integration tests work properly
- [x] Make sure that documentation generates properly

*Learnings*:

It's impossible to use JSDoc with a gradual Typescript adoption and essentially JSDoc site generator was the main pain point in this PR.

After spending 5h trying to leave the current documentation website as it is I gave up and started exploring alternatives.

End up with using https://typedoc.org/. I set it up to pick up documentation from `typings` folder instead of the actual codebase. Once we fully migrate from JS to TS, we can point typedoc to generate docs from the code itself.

The switch from JSDoc to Typedoc changes UI from the documentation and requires us to move code examples from `lib` JSDoc comments to `typings`.

*Follow-ups*:

* Set up ESLint for Typescript
* Fix ESLint errors (there are 100+ in master)
* Prettier for the whole codebase


typings/create-http-client.d.ts
```diff
@@ -0,0 +1,82 @@
+/// <reference types="node" />
```
suevalov, line 1: This file is generated automatically based on an actual implementation file.

lib/create-cma-http-client.ts
```diff
@@ -0,0 +1,128 @@
...
+import {createHttpClient, getUserAgentHeader} from 'contentful-sdk-core'
+import cloneDeep from 'lodash/cloneDeep'
+
+import { Agent as httpAgent } from 'http'
+import { Agent as httpsAgent } from 'https'
+import { AxiosProxyConfig } from 'axios'
+
+export type ClientParams = {
+  /**
+   * Contentful CDA Access Token
```
phoebeschmidt, line 17: but this is a CMA token, right?

lib/create-cma-http-client.ts
```diff
@@ -0,0 +1,128 @@
...
+import axios from 'axios'
+// @ts-ignore
+import {createHttpClient, getUserAgentHeader} from 'contentful-sdk-core'
+import cloneDeep from 'lodash/cloneDeep'
+
+import { Agent as httpAgent } from 'http'
+import { Agent as httpsAgent } from 'https'
+import { AxiosProxyConfig } from 'axios'
+
+export type ClientParams = {
```
phoebeschmidt, line 15: there's some other client params that weren't reflected inthe jsdocs 😅  (but they're in the readme! hah)

* requestLogger
* responseLogger
* basePath

https://github.com/contentful/contentful-sdk-core/blob/master/lib/create-http-client.js

package.json
```diff
@@ -18,23 +18,25 @@
   "tonicExampleFilename": "tonic-example.js",
   "scripts": {
     "clean": "rimraf dist",
-    "build": "npm run clean && npm run build:modules && npm run build:standalone",
+    "build": "npm run clean && npm run build:modules && npm run build:standalone && npm run build:types",
     "build:modules": "BABEL_ENV=modules babel lib -d dist/es-modules/",
     "build:standalone": "webpack && NODE_ENV=production webpack",
     "build:standalone:log": "NODE_ENV=production WEBPACK_MODE=log webpack --json --profile --progress > webpack-build-log.json && webpack-bundle-analyzer webpack-build-log.json",
-    "docs:build": "jsdoc -c jsdoc.json",
+    "build:types": "yarn tsc --declaration --noEmit false --emitDeclarationOnly true --allowJs false",
```
marcolink, line 25: Maybe clear the `generated` folder before emitting to ensure up-to-date state of type declarations.

suevalov: good idea!

lib/create-cma-http-client.ts
suevalov: @phoebeschmidt I added requestLogger and reponseLogger. There no JSDoc in `contentful-sdk-core` about basePath, so I'm not sure how to describe the purpose. Are we sure that we want to expose it here?

package.json
suevalov: @marcolink added

lib/create-space-api.js
```diff
@@ -173,325 +74,72 @@ export default function createSpaceApi ({
       .then((response) => wrapSpace(http, response.data), errorHandler)
   }
 
-  /**
```
suevalov, line 176: All these JSDoc comments are moved to typings files. When we refactor this file to TS the comments should be moved back to be collocated with an actual implementation again.

lib/create-cma-http-client.ts
phoebeschmidt: you're right maybe let's not expose it

package.json
```diff
@@ -18,23 +18,25 @@
   "tonicExampleFilename": "tonic-example.js",
   "scripts": {
     "clean": "rimraf dist",
-    "build": "npm run clean && npm run build:modules && npm run build:standalone",
+    "build": "npm run clean && npm run build:modules && npm run build:standalone && npm run build:types",
     "build:modules": "BABEL_ENV=modules babel lib -d dist/es-modules/",
     "build:standalone": "webpack && NODE_ENV=production webpack",
     "build:standalone:log": "NODE_ENV=production WEBPACK_MODE=log webpack --json --profile --progress > webpack-build-log.json && webpack-bundle-analyzer webpack-build-log.json",
-    "docs:build": "jsdoc -c jsdoc.json",
+    "build:types": "rimraf typings/generated && yarn tsc --declaration --noEmit false --emitDeclarationOnly true --allowJs false",
```
Khaledgarbaya, line 25: Any reason we have a mix of `yarn` and `npm` ways of doing things?

suevalov: oops, a good catch. it's just a habit of using yarn 🤣

PR #322, opened 2020-05-27
chore: rewrite helpers and api-key entity to Typescript

suevalov:


lib/instance-actions.ts
```diff
@@ -0,0 +1,181 @@
+import cloneDeep from 'lodash/cloneDeep'
+import { AxiosInstance } from 'axios'
+import type { MetaSysProps, DefaultElements } from './common-types'
```
marcolink, line 3: is this `type` meant to be here?

```suggestion
import { MetaSysProps, DefaultElements } from './common-types'
```

suevalov: yep, it's meant to be here. It's a new thing in the latest typescript version to differentiate type imports from actual code imports

lib/instance-actions.ts
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * @private
+ */
+export function createUpdateEntity<T = unknown>({
+  http,
+  entityPath,
+  wrapperMethod,
+  headers,
+}: {
```
taye, line 17: Might be worth using `@internal` and setting [`compilerOptions.stripInternal`](https://www.typescriptlang.org/v2/en/tsconfig#stripInternal) to true so that they aren't available in the generated declaration files.

suevalov: nice tip!

lib/create-cma-http-client.ts
```diff
@@ -20,41 +19,41 @@ export type ClientParams = {
...
    * @default false
    */
-  insecure?: boolean | null
+  insecure?: boolean
   /**
    * If we should retry on errors and 429 rate limit exceptions
    * @default true
    */
-  retryOnError?: boolean | null
+  retryOnError?: boolean
```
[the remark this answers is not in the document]
suevalov: I'd consider it as a fix of wrong typings. We have many of them in place and each proper fix will mean a breaking change for typescript typings. It's just an outcome of manual typings for a long time.

lib/instance-actions.ts
marcolink: @suevalov ah cool. Now i have to teach my editor about this new feature 👍

PR #323, opened 2020-05-28
chore: app management typescript

suevalov:
* app-installation and app-definition to typescript
* fix building ESM modules (there was no release yet, so it's not a problem)


lib/entities/app-definition.ts
```diff
@@ -0,0 +1,155 @@
...
+
+interface ArrayFieldType {
+  type: 'Array'
+  items: SingleFieldType | LinkFieldType
+}
+
+type FieldType = SingleFieldType | LinkFieldType | ArrayFieldType
+
+// Locations
+type AppLocation = 'app-config' | 'entry-sidebar' | 'entry-editor' | 'dialog'
```
giotiskl, line 41: this definition of `AppLocation` options looks incomplete :thinking: there's also location 'page' now https://www.contentful.com/developers/docs/extensibility/app-framework/locations/

typings/generated/entities/app-definition.d.ts
```diff
@@ -0,0 +1,106 @@
...
+interface LinkFieldType {
+    type: 'Link';
+    linkType: LinkType;
+}
+interface ArrayFieldType {
+    type: 'Array';
+    items: SingleFieldType | LinkFieldType;
+}
+declare type FieldType = SingleFieldType | LinkFieldType | ArrayFieldType;
+declare type AppLocation = 'app-config' | 'entry-sidebar' | 'entry-editor' | 'dialog';
```
giotiskl, line 17: same here 'page' missing

lib/entities/app-definition.ts
suevalov: good spot, I copied from `typings`, so it was missing from the beginning :) 

I'll add it.

PR #324, opened 2020-05-28
chore: webhook to typescript

anho:
* webhook to typescript


lib/entities/webhook.ts
```diff
@@ -0,0 +1,215 @@
...
+   */
+  transformation?: {
+    method?: string
+    contentType?: string
+    includeContentLength?: boolean
+    body?: object
+  }
+}
+
+export interface WebHooks extends WebhookProps, DefaultElements<WebhookProps> {
```
phoebeschmidt, line 60: I think everywhere else we do "Webhooks" with a lowercase H? Could be wrong

anho: It's also the only place where an entity is in plural form 🤷‍♂️ Didn't want to break BC so I sticked with it.

suevalov: I also avoid doing breaking changes for type definitions. Some names are random but I'd collect all these things for the time being a fix them later.

PR #336, opened 2020-06-02
increase timeout parameter for asset processing

ruderngespra:


phoebeschmidt:
once CI is green then go for it

PR #350, opened 2020-06-05
feat: Async iterator for paginated responses

anho:
Something that is convenient for node users specifically but can benefit web apps too:

```
import { createPlainClient, asIterator } from 'contentful-management';
import { from } from 'ix/asynciterable';

const client = createPlainClient({});

const getTitle = (fields, sys) => '¯\\_(ツ)_/¯';

(async () => {
  const query = { limit: 26, order: 'sys.updatedAt' };
  const result = asIterator(client.entry.getMany, { query });
  const source = from(result).pipe(
    map(async ({ fields, sys }) => ({ sys, title: getTitle(fields, sys) })),
    // filter(startsWith('u')),
    take(5), // will stop after taking 5, no more network requests
  )

  let i = 1;
  for await (const { title } of source) {
    console.log(`${i++}: ${title}`);
  }
})();

### DIFF
--- a/lib/contentful-management.js
+++ b/lib/contentful-management.js
@@ -8,6 +8,7 @@ import createContentfulApi from './create-contentful-api'
 import { createCMAHttpClient } from './create-cma-http-client'
 export { createCMAHttpClient } from './create-cma-http-client'
 export { createPlainClient } from './plain/plain-client'
+export { asIterator } from './plain/as-iterator'
 
 /**
  * Create a client instance
--- a/lib/plain/as-iterator.ts
+++ b/lib/plain/as-iterator.ts
@@ -1,0 +1,42 @@
+import { CollectionProp } from '../types/common-types'
+import { QueryParams } from './index'
+import cloneDeep from 'lodash/cloneDeep'
+import merge from 'lodash/merge'
+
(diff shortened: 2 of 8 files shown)


marcolink:
Pretty cool stuff 👏 
Any chance to add some tests for it?

suevalov:
I wonder if it will be compatible with a legacy browsers build. By looking at the code it seems that it can be an optional wrapper.

suevalov:
If it can be composed on top - I'd prefer to have this separate from the main client to make API as simple as possible

anho:
> Pretty cool stuff 👏
> Any chance to add some tests for it?

It's already on my list. I should have made a draft PR 👍 

> I wonder if it will be compatible with a legacy browsers build. By looking at the code it seems that it can be an optional wrapper.

Probably not compatible: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator#Browser_compatibility - might check how github is doing that with their Rest SDK.

> If it can be composed on top - I'd prefer to have this separate from the main client to make API as simple as possible

It can potentially be something on top of the main methods, but I prefer the simplicity on the consumer side for now. I might take a look in the next couple of days on how to make this bearable on the other side of the fence without intruding the main code.

PR #353, opened 2020-06-05
chore: convert-organization-membership-to-typescript

burakukula:


lib/entities/organization-membership.ts
```diff
@@ -0,0 +1,127 @@
...
+          {
+            headers: {
+              'X-Contentful-Version': this.sys.version || 0,
+            },
+          }
+        )
+        .then((response) => wrapOrganizationMembership(http, response.data), errorHandler)
+    },
+
+    delete: createDeleteEntity({
```
burakukula, line 93: I updated this method to use generic solution. Before it was done like this:

```
delete: function () {	
      return http	
        .delete('organization_memberships' + '/' + this.sys.id)	
        .then(() => {}, errorHandler)	
    },
```
I think it will work the same way now.

lib/entities/organization-membership.ts
```diff
@@ -0,0 +1,127 @@
+import cloneDeep from 'lodash/cloneDeep'
+import { freezeSys, toPlainObject } from 'contentful-sdk-core'
+import enhanceWithMethods from '../enhance-with-methods'
+import errorHandler from '../error-handler'
+import { createDeleteEntity } from '../instance-actions'
+import { AxiosInstance } from 'axios'
+import { MetaSysProps, QueryOptions, DefaultElements, CollectionProp } from '../common-types'
+
+export interface Options {
```
suevalov, line 9: I wouldn't create a special type for this and rather use `QueryOptions` in place

lib/entities/organization-membership.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export interface OrganizationMembership
+  extends OrganizationMembershipProps,
+    DefaultElements<OrganizationMembershipProps> {
+  /**
+   * Sends an update to the server with any changes made to the object's properties
+   * @memberof OrganizationMembership
+   * @func update
+   * @return {Promise<OrganizationMembership>} Object returned from the server with updated changes.
+   * @example
```
suevalov, line 37:
```suggestion
   * @return Object returned from the server with updated changes.
   * @example ```javascript
```

lib/entities/organization-membership.ts
```diff
@@ -0,0 +1,127 @@
...
+   * .catch(console.error)
+   * ```
+   */
+  delete(): Promise<void>
+}
+
+function createOrganizationMembershipApi(http: AxiosInstance) {
+  return {
+    update: function () {
+      const raw = this.toPlainObject()
```
suevalov, line 77: In order to make `this` typed, I suggest using the following approach: https://github.com/contentful/contentful-management.js/blob/master/lib/entities/app-installation.ts#L67-L73

typings/organization.d.ts
```diff
@@ -5,6 +5,7 @@ import { Options as TeamMembershipOptions, TeamMembership, TeamMembershipProps }
 import { Options as TeamSpaceMembershipOptions, TeamSpaceMembership } from './generated/entities/team-space-membership'
 import { Team, TeamProps } from './generated/entities/team'
 import { User } from './generated/entities/user'
+import { Options as OrganizationMembershipOptions, OrganizationMembership } from './generated/entities/organization-membership'
```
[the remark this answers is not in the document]
suevalov: It's good to have them now :)

lib/entities/organization-membership.ts
burakukula: @suevalov thx for feedback! yeah, make much more sense. It's updated, please have a look if this is how we would like to solve it.

lib/entities/organization-membership.ts
burakukula: @suevalov done :)

lib/entities/organization-membership.ts
burakukula: @suevalov done! :)

PR #354, opened 2020-06-08
chore: convert organization entity to ts

burakukula:


lib/entities/organization.ts
```diff
@@ -1,27 +1,21 @@
 import cloneDeep from 'lodash/cloneDeep'
 import { freezeSys, toPlainObject } from 'contentful-sdk-core'
 import enhanceWithMethods from '../enhance-with-methods'
+// @ts-ignore
```
burakukula, line 4: This is temporary. As soon as we convert create-org-api into ts we would be able to remove it? WDYT about this solution? or maybe it would be better to do it some other way?

suevalov: Let's use ``// @ts-expect-error`` and remove changes `eslint` file

lib/entities/organization.ts
```diff
@@ -33,9 +27,12 @@ import createOrganizationApi from '../create-organization-api'
  * @param  {Object} data - API response for a Organization
```
suevalov, line 27: Could you please delete `{ }` from all comments? 

```suggestion
 * @param data - API response for an Organization
```

lib/entities/organization.ts
burakukula: For some reason it does not work when it comes to imports 😕 that's why I used ignore here. But it's possible that I do something wrong.

suevalov: Then

```
// eslint-disable-next-line
// @ts-ignore
```

without making changes in .eslint file

lib/entities/organization.ts
burakukula: @suevalov sure! done! ✅

lib/entities/organization.ts
burakukula: @suevalov ✅

typings/generated/entities/organization.d.ts
```diff
@@ -0,0 +1,33 @@
...
+     */
+    name: string;
+};
+/**
+ * This method creates the API for the given organization with all the methods for
+ * reading and creating other entities. It also passes down a clone of the
+ * http client with an organization id, so the base path for requests now has the
+ * organization id already set.
+ * @private
+ * @param  {Object} http - HTTP client instance
```
suevalov, line 19: Looks like it needs to be regenerated

suevalov: @burakukula please, run `build:types` one more time

burakukula: upsi, forgot about it. ✅

PR #355, opened 2020-06-08
chore: convert organization invitation to ts

burakukula:


lib/entities/organization-invitation.ts
```diff
@@ -0,0 +1,30 @@
+import cloneDeep from 'lodash/cloneDeep'
+import { freezeSys, toPlainObject } from 'contentful-sdk-core'
+import { AxiosInstance } from 'axios'
+import { MetaLinkProps, DefaultElements } from '../common-types'
+
+export type OrganizationInvitationProps = {
+  firstName: string
```
suevalov, line 7: Is there a `sys`?

burakukula: I am not really sure how to handle this one...
Screenshot 2020-06-08 at 11 50 57

suevalov:
```
sys: MetaSysProps & { user: User | null, invitationUrl: string  }
```

burakukula: @suevalov updated - one thing that we talked about is user type - we don't have any `User` in common types, maybe I would create one? I am not really sure if we gonna need it in many places though..

suevalov: if the shape is consistent for all usages then it probably makes sense to create, otherwise, let's leave it here

PR #357, opened 2020-06-10
chore: convert locale to typescript

suevalov:


typings/generated/entities/locale.d.ts
```diff
@@ -0,0 +1,94 @@
+import { AxiosInstance } from 'axios';
+import { Except, SetOptional } from 'type-fest';
```
marcolink, line 2: I'm just curious, what is the difference between `setOptional` and the builtin `Partial`?
Same for `Except`  and `Exclude`

marcolink: Nevermind, looked at the docs 👍

suevalov: The `Omit` type does not restrict the omitted keys to be keys present on the given type, while `Except` does.

`Partial` marks all properties optional while `SetOptional` only those that mentioned keeping other untouched.

PR #364, opened 2020-06-15
feat: all typescript definitions are generated from sources

suevalov:
* Finishes Typescript migration
* Removes typings from the repository: they will be distributed in `dist/typings` 

* All the code shuffling is needed only for a proper documentation generation.
* No logic changed, no tests changed, all tests pass


package.json
```diff
@@ -32,8 +32,8 @@
     "pretest": "rimraf coverage && npm run lint",
     "test": "npm run test:cover && npm run test:integration && npm run test:browser-local && npm run test:size",
     "test:ci": "./node_modules/contentful-sdk-core/bin/test-ci.sh",
-    "test:cover": "BABEL_ENV=test babel-node  --extensions \".ts\" --extensions \".js\" ./node_modules/istanbul/lib/cli.js cover ./test/runner.js | tap-dot",
-    "test:unit": "BABEL_ENV=test babel-node --extensions \".ts\" --extensions \".js\" ./test/runner.js | tap-dot",
+    "test:cover": "BABEL_ENV=test babel-node  --extensions \".ts\" --extensions \".js\" ./node_modules/istanbul/lib/cli.js cover ./test/runner.js | tap-spec",
```
anho, line 35: What that change back to `tap-spec`?

lib/contentful-management.js
```diff
@@ -1,26 +0,0 @@
...
- * with access to the Contentful Content Management API.
- * @packageDocumentation
- */
-
-import createContentfulApi from './create-contentful-api'
-import { createCMAHttpClient } from './create-cma-http-client'
-
-/**
- * Create a client instance
- * @param params - Client initialization parameters
```
anho, line 12: Intended that this went missing?

lib/entities/content-type.ts
```diff
@@ -195,7 +196,7 @@ export interface ContentType
 
 function createContentTypeApi(http: AxiosInstance): ContentTypeApi {
   return {
-    update: createUpdateEntity({
+    update: createUpdateEntity<ContentType>({
```
anho, line 199: Can `ContentType` be inferred from the wrapperMethod?

suevalov: I wouldn't get it to work easily but I'll give it another try.

package.json
suevalov: I simply couldn't correlate the error and the test which failed. IMO, `spec` works better. Could you elaborate on why `dot` is better?

lib/contentful-management.js
suevalov: This comment is not used in the documentation generated from sources but I can bring it back.

anho: Isn't it this one? https://contentful.github.io/contentful-management.js/contentful-management/5.26.5/globals.html#createclient

package.json
anho: If it's not reporting errors than the reporter is just plain bad :) Usually I see dot reporters superior because it reduces the "noise" of all the working tests and simply highlights the ones failing so you CAN actually see them easily. But again, if tap-dot is not living up to that promise for you, just use tap-spec.

lib/entities/content-type.ts
anho: Something along these lines:

```
type WrapperMethod<E, A extends AxiosInstance = AxiosInstance> = (http: A, data: any) => E;

export function createUpdateEntity<T>({}: {
  wrapperMethod: WrapperMethod<T>
})
```

package.json
suevalov: so, do you think we should revert changes here and leave `tab-spec` installed so we can debug using it?

lib/contentful-management.js
suevalov: 👀

lib/entities/content-type.ts
suevalov: done! a good suggestion 👍

package.json
anho: Just move forward for now and maybe you can show me 1on1 what the actual problem is with tab-spec.

suevalov:
@marcolink yep, everything seems to be correct :)

Issue #363, opened 2020-06-15
Typing for 'Role' seems to be incorrect


fabbru-s2:
Hi,

it looks like the type for `Role` does not seem to be correct. When comparing the [type](https://github.com/contentful/contentful-management.js/blob/master/lib/entities/role.ts) to the [documentation](https://www.contentful.com/developers/docs/tutorials/general/roles-via-api/)  you can make a few observations, e.g.:
-  `Environments` and `EnvironmentAliases` are keys of the `permissions` object but missing in the [type](https://github.com/contentful/contentful-management.js/blob/master/lib/entities/role.ts)
- `actions` in `policies` can actually be an array of strings but the [type](https://github.com/contentful/contentful-management.js/blob/master/lib/entities/role.ts) is defined as string only.

Maybe issue [360](https://github.com/contentful/contentful-management.js/issues/360) is related  to this malformed typing as well.

Cheers,
Fabian

suevalov:
Hi @fabbru-s2! Thanks for reporting the issue. I can confirm that typings are not correct.

We'll fix it and release a new version soon

suevalov:
Hey @fabbru-s2 👋 

The fix is available in v5.27.1: https://github.com/contentful/contentful-management.js/releases/tag/v5.27.1

Issue #360, opened 2020-06-10
Updating space roles permissions and policies gives 503 error


rohit-010:
_client.getSpace('<space_id>')
.then((space) => space.getRole('<roles_id>'))
.then((roles) => {
  roles.name = 'New role name'
  return roles.update()
})_

Above method works fine to update name and description of role
But throws 503 Server error when we try to update policies and permissions 

Is there another way to update policies and permissions  using API ?

vbeleuta:
Hi @rohit-010, 

Could you try and see if you can update policies and permissions by using v5.27.1: https://github.com/contentful/contentful-management.js/releases/tag/v5.27.1?

rohit-010:
Hello @vbeleuta 
Thanks for update now  able to update policies .

Cheers

PR #366, opened 2020-06-16
fix: role typings

vbeleuta:
I believe these are the correct typings for space roles.
Should fix issue https://github.com/contentful/contentful-management.js/issues/363 and potentially https://github.com/contentful/contentful-management.js/issues/360

### DIFF
--- a/lib/entities/role.ts
+++ b/lib/entities/role.ts
@@ -6,6 +6,21 @@ import { wrapCollection } from '../common-utils'
 import { createUpdateEntity, createDeleteEntity } from '../instance-actions'
 import { MetaSysProps, DefaultElements } from '../common-types'
 
+export type ActionType =
+  | 'read'
+  | 'create'
+  | 'update'
+  | 'delete'
+  | 'publish'
+  | 'unpublish'
+  | 'archive'
+  | 'unarchive'
+
+type ConditionType = 'and' | 'or' | 'not' | 'equals'
+export type ConstraintType = {
+  [key in ConditionType]: ConstraintType[] | any
+}
+
 export type RoleProps = {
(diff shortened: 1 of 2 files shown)


lib/entities/role.ts
```diff
@@ -12,13 +12,15 @@ export type RoleProps = {
    * Permissions for application sections
    */
   permissions: {
-    ContentDelivery: string
+    ContentDelivery: string[] | string
     ContentModel: string[]
-    Settings: any[]
+    EnvironmentAliases: string[] | string
+    Environment: string
```
suevalov, line 18: Looks like it should be plural - Environment**s**

Also, by looking at the example all of them should be `string[] | string`

lib/entities/role.ts
```diff
@@ -12,13 +12,15 @@ export type RoleProps = {
...
+    Settings: any[] | string
   }
   policies: {
     effect: string
-    actions: string
+    actions: string[] | string
```
suevalov, line 23: Looks like actions list is a limited set of actions:

```
"read",
"create",
"update",
"delete",
"publish",
"unpublish",
"archive",
"unarchive"
```

so it can be 'all' | ActionType[]

lib/entities/role.ts
```diff
@@ -12,13 +12,15 @@ export type RoleProps = {
...
     constraint: any
```
suevalov, line 24: Can we be more specific than `any`?

suevalov:
@vbeleuta could you please merge master branch? 

Changes in `typings/generated/entities/role.d.ts` are not needed anymore. Types will be automatically generated from Typescript sources.

PR #370, opened 2020-06-22
chore: spread plain endpoints into distinct files

anho:


lib/create-environment-api.ts
```diff
@@ -1,7 +1,8 @@
 import { createRequestConfig } from 'contentful-sdk-core'
 import errorHandler from './error-handler'
 import entities from './entities'
-import * as endpoints from './plain'
+import * as endpoints from './plain/endpoints'
+import { locale } from './plain/endpoints'
```
suevalov, line 5: Is it intended to change? We already have all endpoints imported one line above

PR #371, opened 2020-06-25
feat: remove all deprecated methods [experimental]

suevalov:
Removes all deprecated methods from an experimental branch to simplify refactoring to a flat API. Once we release the new optional flat API it will be a breaking change due to deleted deprecated methods, but it's not a problem for all dependant CLI because they don't use deprecated methods.

A breaking change for us is a chance to rename some typings.


suevalov:
@marcolink nope, it's an experimental branch, and docs are not published for this one. it's solely for our purpose at the moment. we'll do proper communication and docs update once we merge experimental to the main branch.

Issue #365, opened 2020-06-15
Typings of Webhook is not correct


rohit-010:
Hello Team,
I am using createWebhookWithId method and it is accepting two params 
 1) webhook id (string)
 2) data (WebhookProps) 
So problem is with WebhookProps 
WebhookProps has mandatory attributes like : 
      sys: MetaSysProps;
      httpBasicUsername: string;
      httpBasicPassword: string;

As I am creating webhook I wont have detail of **sys**
**httpBasicUsername** and **httpBasicPassword** -- this I dont want to use for my webhook

Can you please fix this typescript issue:
Also if I refer to your API docs example given is : 
_client.getSpace('<space_id>')
.then((space) => space.createWebhookWithId('<webhook_id>', {
  'name': 'My webhook',
  'url': 'https://www.example.com/test',
  'topics': [
    'Entry.create',
    'ContentType.create',
    '*.publish',
    'Asset.*'
  ]
}))_

So this issue is related to typings , can you please fix this asap.
Thanks

suevalov:
HI @rohit-010,

We rewrote the core of the library to Typescript and now type definitions should be the way better.
Could you please verify that the issue is gone in `5.27.0`?

rohit-010:
Hello @suevalov ,

Thanks for change but still few other's are not fixed 
example I want to pass secret header then it is not allowing or API is throwing error while creating new webhook

How can I pass header like : 
  _"headers": [
        {
            "key": "key",
			"value: "key-value"
            "secret": true
        }
    ]_
	
Also if I don't want to pass httpBasicUsername and httpBasicPassword, this is mandatory in current WebhookProps so can you please handle this as well.

Thanks

rohit-010:
Hello @suevalov ,

Also I think URL is accepting as string in Typings but string is not accepted by server 
So can you please change typings of url from string to URL.

Thanks

rohit-010:
@suevalov  any update on this ??

Issue #306, opened 2020-04-13
createWebhook with secret Authorization-Bearer Token


jfarris587:
https://contentful.github.io/contentful-management.js/contentful-management/5.21.1/Webhook.html#.Webhook

You can create a custom webhook with the module and have httpBasicUsername and httpBasicPassword. However, there is no option to `add a secret header` like there is in the web interface

suevalov:
Hi @jfarris587 

We updated our type definitions for webhooks and you can pass an optional `secret: true` value to any header

https://github.com/contentful/contentful-management.js/blob/master/lib/entities/webhook.ts#L36

PR #373, opened 2020-06-25
fix: incorrect webhook types

### LINKED ISSUE #372: Webhook Props --Typescript headers
Hello Team,

I am working on typescript project , I can see webhook props typings has some issue in sdk.

Example: How can I pass header like :
"headers": [
{
"key": "key",
"value: "key-value"
"secret": true
}
] -- This throws typescript error

Also inside below properties are mandatory webhook props :
         httpBasicUsername: string;
          httpBasicPassword: string;

Can you please update sdk or let me how to fix this issue.

Thanks

suevalov:
Closes #372 #365 #306

### DIFF
--- a/lib/entities/webhook.ts
+++ b/lib/entities/webhook.ts
@@ -1,4 +1,5 @@
 import { AxiosInstance } from 'axios'
+import { JsonValue } from 'type-fest'
 import cloneDeep from 'lodash/cloneDeep'
 import { freezeSys, toPlainObject } from 'contentful-sdk-core'
 import enhanceWithMethods from '../enhance-with-methods'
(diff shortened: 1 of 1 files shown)


lib/entities/webhook.ts
```diff
@@ -28,33 +28,36 @@ export type WebhookProps = {
   /**
    * Username for basic http auth
    */
-  httpBasicUsername: string
+  httpBasicUsername: string | null
```
Khaledgarbaya, line 31: I think this should be `httpBasicUsername?: string` because the property will omitted from the payload if it does not exist

lib/entities/webhook.ts
```diff
@@ -28,33 +28,36 @@ export type WebhookProps = {
...
-  httpBasicPassword: string
+  httpBasicPassword: string | null
 
   /**
    * Headers that should be appended to the webhook request
    */
-  headers: {
-    [key: string]: string
-  }
+  headers: Array<{ key: string; value: unknown; secret?: boolean }>
```
marcolink, line 41: I'm always uncertain with this, but aren't header values always strings? or, get transformed to string?

jelz: Only ASCII strings are accepted for both key and value

lib/entities/webhook.ts
```diff
@@ -28,33 +28,36 @@ export type WebhookProps = {
...
+
   /**
    * Transformation to apply
    */
   transformation?: {
     method?: string
     contentType?: string
     includeContentLength?: boolean
-    body?: object
+    body?: Record<string, any>
```
jelz, line 60: `body` can be any valid JSON, so `true`, `['test']`, or `123` are valid too

lib/entities/webhook.ts
```diff
@@ -28,33 +28,36 @@ export type WebhookProps = {
...
 
   /**
    * Topics the webhook wants to subscribe to
    */
   topics: string[]
 
+  /**
+   * Webhook filters
+   */
+  filters?: Array<{ [key: string]: any }>
```
jelz, line 51: I'm fine with that but could be typed better. Will send a link to aschema

lib/entities/webhook.ts
suevalov: updated

lib/entities/webhook.ts
suevalov: updated

lib/entities/webhook.ts
suevalov: updated

lib/entities/webhook.ts
marcolink: shouldn't that be of type 

```ts
body?: JsonValue
```

marcolink: I obviously commented on an outdated version 😆 👍

PR #374, opened 2020-06-25
feat: add asset endpoint and update headers to plain cma client

djagya:
Adds the `asset` endpoint (similar to `entry`) to the plain CMA client. Adds a way to specify headers for `update` calls.

### DIFF
--- a/lib/entities/space.ts
+++ b/lib/entities/space.ts
@@ -12,7 +12,7 @@ type SdkHttpClient = AxiosInstance & {
 }
 
 export type SpaceProps = {
-  sys: BasicMetaSysProps
+  sys: BasicMetaSysProps & { organisation: { sys: { id: string } } }
   name: string
 }
 
--- a/lib/entities/space.ts
+++ b/lib/entities/space.ts
@@ -12,7 +12,7 @@ type SdkHttpClient = AxiosInstance & {
 }
 
 export type SpaceProps = {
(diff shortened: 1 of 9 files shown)


lib/entities/space.ts
```diff
@@ -12,7 +12,7 @@ type SdkHttpClient = AxiosInstance & {
 }
 
 export type SpaceProps = {
-  sys: BasicMetaSysProps
+  sys: BasicMetaSysProps & { organisation: { sys: { id: string } } }
```
suevalov, line 15:
```suggestion
  sys: BasicMetaSysProps & { organization: { sys: { id: string } } }
```

lib/plain/endpoints/asset.ts
```diff
@@ -0,0 +1,131 @@
+import { AxiosInstance } from 'axios'
+import * as raw from './raw'
+import { CreateAssetProps, AssetProps } from '../../entities/asset'
+import { normalizeSelect } from './utils'
+import cloneDeep from 'lodash/cloneDeep'
+import { GetEnvironmentParams } from './environment'
+import { QueryParams, CollectionProp } from './common-types'
+
+export type GetManyEntriesParams = GetEnvironmentParams & QueryParams
```
suevalov, line 9: Could you rename it to `GetManyAssetsParams`?

PR #377, opened 2020-06-26
feat: add more plain methods

anho:
Slightly deviated from the current naming schema for organizations because its not paginated.


lib/plain/plain-client.ts
```diff
@@ -23,6 +26,7 @@ export const createPlainClient = (params: ClientParams, defaults?: DefaultParams
     },
     user: {
       getManyForSpace: wrap(wrapParams, endpoints.user.getManyForSpace),
+      getCurrent: wrapHttp(http, endpoints.user.getCurrent),
```
suevalov, line 29: I think we can still use `wrap` and it's up to a method use or not to use params?

suevalov:
@anho you need to use "feat: " prefix for the PR if you want to trigger publish once it's merged

lib/plain/plain-client.ts
anho: There would be a mandatory `params` popping up in the typings.

suevalov: got it

PR #383, opened 2020-06-30
feat: move all environment methods to plain client

suevalov:
Exposes all environment-related endpoints to a plain CMA client and reuses endpoints in the original CMA client internals.

### DIFF
--- a/lib/create-space-api.ts
+++ b/lib/create-space-api.ts
@@ -195,14 +195,19 @@ export default function createSpaceApi({ http }: { http: AxiosInstance }) {
      * .catch(console.error)
      * ```
      */
-    createEnvironmentWithId(id: string, data: CreateEnvironmentProps, sourceEnvironmentId: string) {
+    createEnvironmentWithId(
+      id: string,
+      data: CreateEnvironmentProps,
+      sourceEnvironmentId?: string
+    ) {
       const raw = this.toPlainObject() as SpaceProps
       return endpoints.environment
         .createWithId(
           http,
           {
             spaceId: raw.sys.id,
-            environmentId: sourceEnvironmentId,
+            environmentId: id,
+            sourceEnvironmentId,
           },
           data
         )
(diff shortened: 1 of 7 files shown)


lib/plain/endpoints/environment.ts
```diff
@@ -13,6 +14,10 @@ export const get = (http: AxiosInstance, params: GetEnvironmentParams) => {
   )
 }
 
+export const getAll = (http: AxiosInstance, params: GetSpaceParams) => {
```
anho, line 17: That endpoint is paginated, so it should be getMany with additional query params.

lib/entities/environment.ts
```diff
@@ -22,6 +22,10 @@ export type EnvironmentProps = {
   name: string
 }
 
+export type CreateEnvironmentProps = {
```
anho, line 25: What about `type CreateEnvironmentProps  = Partial<Omit<EnvironmentProps, 'sys'>>`?

lib/create-space-api.ts
```diff
@@ -184,18 +195,18 @@ export default function createSpaceApi({ http }: { http: AxiosInstance }) {
...
-      sourceEnvironmentId?: string
-    ) {
-      return http
-        .put('environments/' + id, data, {
-          headers: sourceEnvironmentId
-            ? { 'X-Contentful-Source-Environment': sourceEnvironmentId }
-            : {},
-        })
-        .then((response) => wrapEnvironment(http, response.data), errorHandler)
+    createEnvironmentWithId(id: string, data: CreateApiKeyProps, sourceEnvironmentId: string) {
```
anho, line 198: The type of data looks wront.

lib/plain/endpoints/environment.ts
suevalov: it's not paginated according to the documentation and the current implementation of the main client.

anho: Interestingly, the API does answer with a paginated request and I am able to utilise skip and limit.

lib/create-space-api.ts
suevalov: fixed

lib/plain/endpoints/environment.ts
suevalov: pricing limits are always below 100 envs, so all our API clients expose `getAll` instead of `getMany`

PR #384, opened 2020-07-01
Feat/tags management entries assets

phoebeschmidt:


lib/create-environment-api.ts
```diff
@@ -1017,6 +1017,7 @@ export default function createEnvironmentApi({
         .put(`tags/${id}`, {
           name,
           sys: {
+            type: 'Tag',
```
marcolink, line 1020: @phoebeschmidt great catch :)

PR #389, opened 2020-07-02
feat: move content type methods

anho:
Add all existing endpoints for content types to plain client and makes current client used these endpoints.

### DIFF
--- a/lib/create-environment-api.ts
+++ b/lib/create-environment-api.ts
@@ -178,7 +178,7 @@ export default function createEnvironmentApi({
 
     /**
      * Gets a Content Type
-     * @param id - Content Type ID
+     * @param contentTypeId - Content Type ID
      * @return Promise for a Content Type
      * @example ```javascript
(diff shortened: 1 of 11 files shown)


lib/create-environment-api.ts
```diff
@@ -292,10 +307,28 @@ export default function createEnvironmentApi({
      * .catch(console.error)
      * ```
      */
-    createContentTypeWithId(id: string, data: CreateContentTypeProps) {
-      return http
-        .put('content_types/' + id, data)
-        .then((response) => wrapContentType(http, response.data), errorHandler)
+    createContentTypeWithId(contentTypeId: string, data: CreateContentTypeProps) {
+      const raw = this.toPlainObject() as EnvironmentProps
+      console.log(
```
suevalov, line 312: You forgot to remove `console.log`

suevalov: @anho please delete before merging :)

PR #395, opened 2020-07-07
feat: add query to `getUiExtensions`

Jwhiles:
We want it possible to pass query parameters to the getExtensions method.
I've tested this with the properties we care about: `stripSrcdoc` and `sys.id[in]`. Both of these work.

I wonder if I should create a new more specific type that just lists these properties? Or whether the existing `QueryOptions` type is the appropriate one to use. 

I also wonder whether I should be adding a new integration test to cover these cases?

### DIFF
--- a/lib/entities/ui-extension.ts
+++ b/lib/entities/ui-extension.ts
@@ -5,14 +5,19 @@ import enhanceWithMethods from '../enhance-with-methods'
 import { createUpdateEntity, createDeleteEntity } from '../instance-actions'
 import { EntryFields } from './entry-fields'
 import { wrapCollection } from '../common-utils'
-import { DefaultElements, MetaSysProps, QueryOptions } from '../common-types'
+import { DefaultElements, MetaSysProps } from '../common-types'
 
-export interface GetUiExtensionsQuery extends QueryOptions {
+export interface GetUiExtensionsQuery {
   /**
    * Whether to exclude srcdoc from the response. This can significantly reduce
    * the response size.
    */
   skipSrcdoc?: boolean
+  /**
+   * A comma seperated list of extension IDs. Only extensions with IDs that are
+   * in the list will be fetched.
+   */
+  'sys.id[in]'?: string
 }
 
 export type UIExtensionProps = {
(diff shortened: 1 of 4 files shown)


test/integration/ui-extension-integration.js
```diff
@@ -70,4 +70,35 @@ export default function uiExtensionTests(t, space) {
...
+      extension: {
+        name: 'Another awesome extension!',
+        src: 'https://anotherawesome.extension',
+        fieldTypes: [{ type: 'Text' }],
+      },
+    })
+
+    const environment = await space.getEnvironment('master')
+
+    const extensions = await environment.getUiExtensions({ 'sys.id[in]': idTwo })
```
jelz, line 96: What do we run it against? Don't we have to mock the API behavior somewhere?

test/integration/ui-extension-integration.js
```diff
@@ -70,4 +70,35 @@ export default function uiExtensionTests(t, space) {
...
+
+    t.equals(extensions.items.length, 1)
+    t.equals(extensions.items.name, 'Another awesome extension!', 'name')
+
+    extensionOne.delete()
```
jelz, line 101: don't we have to await to ensure the test exits with no rubbish laying around?

lib/entities/ui-extension.ts
```diff
@@ -5,7 +5,15 @@ import enhanceWithMethods from '../enhance-with-methods'
 import { createUpdateEntity, createDeleteEntity } from '../instance-actions'
 import { EntryFields } from './entry-fields'
 import { wrapCollection } from '../common-utils'
-import { DefaultElements, MetaSysProps } from '../common-types'
+import { DefaultElements, MetaSysProps, QueryOptions } from '../common-types'
+
+export interface GetUiExtensionsQuery extends QueryOptions {
```
jelz, line 10: I wouldn't extend `QueryOptions`. Only `stripSrcdoc` and `sys.id[in]` are supported in `/extensions`

jelz:
Why did you remove the test?

Jwhiles:
> Why did you remove the test?

Sorry didn't see your comments before I removed it. I realised that I needed to make some other changes to the test.

test/integration/ui-extension-integration.js
Jwhiles: So far as I can tell these are actually end-to-end tests. It doesn't appear that we do any mocking.

suevalov:
@Jwhiles I'm closing this PR cause it seems out of date.

PR #402, opened 2020-07-09
feat(getTags): send query params object

phoebeschmidt:
We should send a query params object in getTags (for consistency with our other get-collection endpoints in the SDK).

### DIFF
--- a/lib/create-environment-api.ts
+++ b/lib/create-environment-api.ts
@@ -1023,9 +1023,9 @@ export default function createEnvironmentApi({
         })
         .then((response) => wrapTag(http, response.data), errorHandler)
     },
-    getTags(skip?: number, limit?: number) {
+    getTags(query: QueryOptions = {}) {
       return http
-        .get('tags', { params: { skip, limit } })
+        .get('tags', createRequestConfig({ query }))
         .then((response) => wrapTagCollection(http, response.data), errorHandler)
     },
     getTag(id: string) {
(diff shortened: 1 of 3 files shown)


lib/create-environment-api.ts
```diff
@@ -1023,9 +1023,9 @@ export default function createEnvironmentApi({
         })
         .then((response) => wrapTag(http, response.data), errorHandler)
     },
-    getTags(skip?: number, limit?: number) {
+    getTags(query: QueryOptions = {}) {
```
TimBeyer, line 1026: The type isn't entirely applicable here.
It includes `content_type` and `include`.

See https://github.com/contentful/contentful-management.js/blob/28483c2cb8c36f874837a7de15724c9ab7d6315f/lib/common-types.ts#L46-L54

Probably we should have a separate type with all the allowed params.

TimBeyer:
This is a breaking change so probably it should be a `feat!`

lib/create-environment-api.ts
phoebeschmidt: great catch! yep will do

phoebeschmidt:
@TimBeyer if it should be a new major version then it should be a `perf` ([relevant sem-rel docs](https://github.com/semantic-release/semantic-release#commit-message-format)). But this shouldn't actually break for anybody, since no one is using it. 

Feat seems right to me. WDYT?

PR #405, opened 2020-07-13
feat: expose usage to a flat api

suevalov:


lib/plain/endpoints/usage.ts
```diff
@@ -0,0 +1,30 @@
+import { AxiosInstance } from 'axios'
+import * as raw from './raw'
+import { UsageProps } from '../../entities/usage'
+import { CollectionProp, QueryParams } from './common-types'
+
+export const getSpaceUsage = (
+  http: AxiosInstance,
+  params: { organizationId: string } & QueryParams
```
anho, line 8: What about a `GetUsageParams = GetOrganizationParams & QueryParams`?

lib/plain/endpoints/common-types.ts
```diff
@@ -1,5 +1,9 @@
 import { QueryOptions, PaginationQueryOptions } from '../../common-types'
 
+export type GetSpaceParams = { spaceId: string }
+export type GetEnvironmentParams = { environmentId: string }
```
anho, line 4: I ALWAYS see it used in conjunction with `GetSpaceParams` so I'd suggest to fall back to what was there earlier: `GetEnvironmentParams = GetSpaceParams & ...`. Environment never ever works without a space so I'd bake that into the types.

lib/plain/endpoints/usage.ts
suevalov: I'm actually not sure if query params are exactly the same for these endpoints but that's that we have now. I'd rather keep them inline if it's used only in 2 places.

lib/plain/plain-client.ts
```diff
@@ -85,6 +87,10 @@ export const createPlainClient = (params: ClientParams, defaults?: DefaultParams
       },
       revoke: wrap(wrapParams, endpoints.personalAccessToken.revoke),
     },
+    usage: {
+      getSpaceUsage: wrap(wrapParams, endpoints.usage.getSpaceUsage),
```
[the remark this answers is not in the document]
suevalov: 👍

lib/plain/endpoints/common-types.ts
suevalov: I actually separated them cause the name of this type was misleading and confusing to me. It's either should be `GetSpaceEnvironmentParams` or have only `{ environmentId: string }` as I implemented it here. What do you think @anho ?

anho: That would make it more explicit which I would be fine with. But for me environments don't exist without a space semantically so I'm not confused 😄

lib/plain/endpoints/usage.ts
anho: As we talked ages ago I always prefer one single type for the params so that they can be reused on the consumer side without inferring the function signature or the need to duplicate that intersection. But lets stick with it for time being.

PR #413, opened 2020-08-04
Feat/scheduled actions

ronaldronson:
Add SDK support for scheduled actions:
- create scheduled action on the space level
- fetch scheduled actions on the space level
- delete scheduled action on scheduled action entity level

### DIFF
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -1,7 +1,13 @@
 import { AxiosInstance } from 'axios'
 import { freezeSys, toPlainObject } from 'contentful-sdk-core'
 import { cloneDeep } from 'lodash'
-import { DefaultElements, MetaSysProps, Link, ISO8601Timestamp } from '../common-types'
+import {
+  DefaultElements,
+  MetaSysProps,
+  Link,
+  ISO8601Timestamp,
+  BasicCursorPaginationOptions,
+} from '../common-types'
 import { wrapCollection } from '../common-utils'
 import enhanceWithMethods from '../enhance-with-methods'
 import errorHandler from '../error-handler'
@@ -22,7 +28,7 @@ enum ScheduledActionStatus {
   canceled = 'canceled',
 }
 
-type SchedulableEntityType = 'Entry' | 'Release'
+type SchedulableEntityType = 'Entry'
 type SchedulableActionType = 'publish' | 'unpublish'
 
 export type ScheduledActionSysProps = Pick<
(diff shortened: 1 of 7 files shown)


lib/entities/scheduled-action.ts
```diff
@@ -0,0 +1,117 @@
...
+  /** action was canceled by a user (terminal state) */
+  canceled = 'canceled',
+}
+
+type SchedulableEntityType = 'Entry'
+type SchedulableActionType = 'publish' | 'unpublish'
+
+export type ScheduledActionSysProps = Pick<
+  MetaSysProps,
+  'id' | 'space' | 'createdAt' | 'createdBy'
```
suevalov, line 36: Some of these properties are being overwritten by the type below. Do we need to use `Pick<MetaSysProps>` at all?

lib/entities/scheduled-action.ts
```diff
@@ -0,0 +1,117 @@
...
+> & {
+  type: 'ScheduledAction'
+  id: string
+  space: Link<'Space'>
```
suevalov, line 40: Usually, we don't specify a concrete literal and leave as `string`. There should be `MetaLinkProps ` type.

lib/entities/scheduled-action.ts
```diff
@@ -0,0 +1,117 @@
...
+  scheduledFor: {
+    datetime: ISO8601Timestamp
+  }
+}
+
+export interface ScheduledActionCollection {
+  sys: {
+    type: 'Array'
+  }
+  pages: {
```
suevalov, line 63: `BasicCursorPaginationOptions`?

lib/entities/scheduled-action.ts
ronaldronson: Thanks, replaced here and in other places with `MetaLinkProps`.

lib/entities/scheduled-action.ts
ronaldronson: Indeed! Fixed.

lib/entities/scheduled-action.ts
ronaldronson: Agree, removed.

ronaldronson:
Hi @suevalov, the scheduled actions entity is space scoped on purpose, and there are no plans to move it to the environment scope. As for the flat API I want to backport those new methods, is it only the matter of merging it to the `experimental` branch or some additional afford needs to be done?

suevalov:
@ronaldronson there's some additional work to be done, but it's rather small I would say. 

An example PR: https://github.com/contentful/contentful-management.js/pull/416

PR #423, opened 2020-08-11
feat: exposes new editor_interface collection endpoint

suevalov:
* Exposes new method to the main client
* Exposes new method to the flat client

* Refactored the code to used `endpoints` in the main client implementation.

### DIFF
--- a/lib/plain/endpoints/editor-interface.ts
+++ b/lib/plain/endpoints/editor-interface.ts
@@ -1,9 +1,8 @@
 import { AxiosInstance } from 'axios'
 import * as raw from './raw'
-import { EnvironmentProps } from '../../entities/environment'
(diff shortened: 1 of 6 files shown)


lib/plain/endpoints/editor-interface.ts
```diff
@@ -14,6 +13,13 @@ export const get = (http: AxiosInstance, params: GetEditorInterfaceParams) => {
   return raw.get<EditorInterfaceProps>(http, getBaseUrl(params))
 }
 
+export const getMany = (http: AxiosInstance, params: GetSpaceEnvironmentParams & QueryParams) => {
```
anho, line 16: The query params are not used, so I'd say either remove it from the signature or we actually use it.

PR #428, opened 2020-08-18
chore: test setup

marcolink:
## Migrate tests to mocha
This PR aims to provide a more maintainable and readable test setup. 
It also removes some legacy test setup that hasn't been used in a while, using [saucelabs](https://saucelabs.com/) (I couldn't find any calls in `circleci` config file).

#### Unit tests
- will run in a node environment by default
- will be written in [mocha](https://mochajs.org/) mostly as async tests using [chai](https://www.chaijs.com/) assertions. 
- will run against source files filed under `./lib` (no need to build in advance)

#### Integration tests
- will run in a node environment by default
- will be written in [mocha](https://mochajs.org/) mostly as async tests using [chai](https://www.chaijs.com/) assertions. 
- will run against source files filed under `./lib` (no need to build in advance)
- will create dedicated space per suite for `write` operations (no clashing of stateful tests anymore).

#### Browser tests
- ~~will run in a browser environment~~
- ~~will run `unit` AND `integration` tests~~
- ~~uses a karma plugin to populate env vars to the `window` object~~
> With the typescript migration browser tests stopped working. there seems to be a problem with `@babel/preset-typescript` and `babel-plugin-rewire-ts`. Since the rewire plugin seems not maintained well anymore, I recommend switching over to `sinon` completely or eventually replace everything with `jest`.

### 🎉 Benefits 🎉  
- better support of test hooks (better test structure)
- full async support
- create dedicated space for all (stateful) `write` tests, and clean up after run.
- faster (no precompile step needed)
- possibility of test pattern matching and watch mode
- remove old bluebird/tap setup
- as a freebie, the [codecov reports](https://codecov.io/gh/contentful/contentful-management.js/branch/feat%2Fkarma-mocha-test-setup) are fixed with separation between different test types (see report below)

### ToDo
- [x] ~~currently, we can only test on one browser. any additional browser will result in a parallel run which will mess up our stateful test~~
- [x] validate if testing source files for node been migrated ~~and bundles code for browser~~ is sufficient.
- [x] add test coverage reports
- [x] ~~run browser tests on circel ci~~
- [x] the little task of migrating all tests 😺 
- [x] add a script to delete all old test spaces  (for now only manual execution)


marcolink:
Screenshot 2020-08-18 at 15 58 49

marcolink:
Screenshot 2020-08-18 at 17 00 59

Example setup for integration tests (using contenttags api)

marcolink:
Screenshot 2020-08-18 at 20 40 51
...and karma browser tests

suevalov:
Great to see that you’re simplifying this complicated test setup 👍

.circleci/config.yml
```diff
@@ -20,8 +20,8 @@ jobs:
           key: dependency-cache-{{ checksum "package.json" }}
       - run: npm install
       - run: npm run build
-      - run: npm run test:cover
-      - run: bash <(curl -s https://codecov.io/bash)
+      - run: npm run test:cover-unit
+      - run: bash <(curl -s https://codecov.io/bash) -F unittests
```
TimBeyer, line 24: Not really a new change, but I'm always a bit concerned seeing curl to bash scripts.  
Maybe we could eventually use the node reporter? Looks like we do that anyway in the `Build Success` step for the integration tests.

test/integration/tag-integration.js
```diff
@@ -1,107 +1,92 @@
...
-  const env = await space.getEnvironment('master')
-  entry = await env.createEntry('layout', { fields: {} })
-  await writeEntityTagsTest(t, entry, env)
-  let asset = await env.getAsset('1YK5kwroV6UEGS64mQs0Eo')
-  await writeEntityTagsTest(t, asset, env)
-}
+  afterEach(async () => {
+    const tags = await environment.getTags(0, 1000)
+    for (let index = 0; index < tags.total; index++) {
+      await tags.items[index]['delete']()
```
TimBeyer, line 28: How about
```
for (const tag of tags.items) {
  await tag.delete()
}
```

or even

```
const deleting = tags.items.map((tag) => tag.delete())
await Promise.all(deleting)
```

TimBeyer (approved):
Overall really happy to see the change away from `tape` and to see proper cleanup and setup steps everywhere.

test/integration/tag-integration.js
marcolink: 👍

suevalov:
@marcolink what's left here? could we wrap it up?

suevalov (approved):
A monumental work!

P.S. Should it be just `chore: change test setup` so it doesn't trigger new release?

test/integration/team-space-membership-integration.js
```diff
@@ -1,7 +1,21 @@
...
+    space = await createTestSpace(client(), 'TSM')
+  })
+
+  after(async () => {
+    if (space) {
+      return space.delete()
+    }
+  })
+
+  test.skip('Create, update and delete teamSpaceMembership', async () => {
```
anho, line 18: Mistake or intentional?

anho (approved):
This work is 💯 ! Just left a question in regards to a skipped test.

test/integration/team-space-membership-integration.js
marcolink: @anho 
Unfortunately intended. This setup expects a certain space state, and would leave the space in a different state in case it breaks. We have to find a better setup with a dedicated space in a follow up PR.

anho: Then maybe leave a comment so other people have a starting point.

marcolink: @anho good point - added.

phoebeschmidt (approved):
way way way better nice work!

do we have tickets in the backlog for the remaining todos? like asset integration "write" tests?

marcolink:
> way way way better nice work!
> 
> do we have tickets in the backlog for the remaining todos? like asset integration "write" tests?

Yes, I created them :)

PR #445, opened 2020-08-31
feat: add support for extensibility entities [EXT-2094]

FLoppix:
Add support to the flat client for the following entities:

- `AppDefinition`
- `AppInstallation`
- `UIExtension`
- `Webhook`


suevalov (changes requested):
A great first step of exposing all endpoints to a new API. 👏 

A part of the refactoring also includes using these new functions in the original CMA client to avoid the duplicated the source of truth for all endpoints.

Please refer to any PR that was merged to `experimental` branch last month to find an example of such refactoring.

P.S. Could you also update PROGRESS.md and mark all these endpoints as ✅ ?

FLoppix:
Sure @suevalov I will !

lib/entities/ui-extension.ts
```diff
@@ -33,6 +34,16 @@ export type UIExtensionProps = {
   }
 }
 
+export type UpdateUIExtenionProps = SetOptional<
+  UIExtensionProps['extension'],
+  'name' | 'fieldTypes' | 'sidebar'
+>
+
+export type CreateUIExtensionProps = RequireExactlyOne<
```
jelz, line 42: Again, why do they differ? Validations are the same...

lib/plain/endpoints/ui-extension.ts
```diff
@@ -0,0 +1,84 @@
...
+      })
+}
+
+export const update = async (
+  http: AxiosInstance,
+  params: GetUiExtensionParams,
+  rawData: UpdateUIExtenionProps,
+  headers?: Record<string, unknown>
+) => {
+  const extenstionToUpdate = await raw.get<UIExtensionProps>(http, getUIExtensionUrl(params))
```
jelz, line 62: Now I get why types differ but is this really what we want to do (fetch -> apply change -> save)

lib/plain/endpoints/ui-extension.ts
```diff
@@ -0,0 +1,84 @@
...
+  const data = {
+    extension: {
+      ...extenstionToUpdate.extension,
+      ...rawData,
+    },
+  }
+
+  return raw.put<UIExtensionProps>(http, getUIExtensionUrl(params), data, {
+    headers: {
+      'X-Contentful-Version': sys.version ?? 0,
```
jelz, line 76: Here we basically hack around optimistic locking

lib/plain/endpoints/ui-extension.ts
```diff
@@ -0,0 +1,84 @@
...
+      ...headers,
+    },
+  })
+}
+
+export const del = (http: AxiosInstance, params: GetUiExtensionParams) => {
+  return raw.del(http, getUIExtensionUrl(params))
```
jelz, line 83: Doesn't require a version?

lib/plain/endpoints/webhook.ts
```diff
@@ -0,0 +1,101 @@
...
+    : raw.post<WebhookProps>(http, getBaseUrl(params), rawData)
+}
+
+export const update = async (
+  http: AxiosInstance,
+  params: GetWebhookParams,
+  rawData: UpdateWebhookProps,
+  headers?: Record<string, unknown>
+) => {
+  const webhookToUpdate = await raw.get<WebhookProps>(http, getWebhookUrl(params))
```
anho, line 80: What is the rationale behind that kind of implementation? It's deviating from the non-plain implementation and its not really in line with one of the main drivers for creating the plain version of the CMA which is reducing network requests.

lib/entities/app-definition.ts
```diff
@@ -71,6 +85,16 @@ export type AppDefinitionProps = {
   locations: LocationDefinition[]
 }
 
+export type CreateAppDefinitionProps = SetOptional<
+  Except<AppDefinitionProps, 'sys'>,
+  'src' | 'locations'
+>
+
+export type UpdateAppDefinitionProps = SetOptional<
```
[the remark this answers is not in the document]
FLoppix: Depends how you see it. If I only want to update one `property` of an entity I dont want to provide the rest of the data that would be mandatory for the `create` of the entity. But that goes hand in hand with @anho's comment. If we say we always pass in the `required` data and lower the network request then these properties will be the same and I need to change it

lib/plain/endpoints/ui-extension.ts
FLoppix: No it does not.

lib/plain/endpoints/ui-extension.ts
FLoppix: If thats what we want then it should be same. I agree.

lib/plain/endpoints/webhook.ts
FLoppix: I think I saw that from a different perspective and that is for sure up for a discussion. If the cycle of an entity is fetch -> change -> update then it does not make sense to have this extra network call. But it seems like we have always want that cycle of fetch the entity -> change the data -> update the data.

lib/plain/endpoints/ui-extension.ts
jelz: Can we check with people building `experimental` branch to understand what model do they want to follow?

My reservation is that if we don't request version from a user then we defeat the purpose of optimistic locking.

cc @suevalov @anho

lib/entities/app-definition.ts
```diff
@@ -68,9 +76,14 @@ export type AppDefinitionProps = {
   /**
    * Locations where the app can be installed
    */
-  locations: LocationDefinition[]
+  locations: Location[]
```
suevalov, line 79: Do you mind renaming this type to be a bit more specific and indicating that this is actually the location of an app?

lib/entities/webhook.ts
```diff
@@ -176,7 +265,7 @@ export interface WebHooks extends WebhookProps, DefaultElements<WebhookProps> {
    * .catch(console.error)
    * ```
    */
-  getCall(id: string): Promise<Record<string, unknown>>
+  getCall(id: string): Promise<WebhookCallDetailsProps>
```
suevalov, line 268: 🎖️

lib/create-environment-api.ts
```diff
@@ -1026,9 +1061,14 @@ export default function createEnvironmentApi({
...
     getAppInstallation(id: string) {
-      return http
-        .get('app_installations/' + id)
-        .then((response) => wrapAppInstallation(http, response.data), errorHandler)
+      const raw = this.toPlainObject() as EnvironmentProps
+      return endpoints.appInstallation
+        .get(http, {
+          spaceId: raw.sys.space.sys.id,
+          environmentId: raw.sys.id,
+          appDefinitionId: id,
```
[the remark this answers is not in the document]
FLoppix: The appInstallationId is in fact the appDefinitionId.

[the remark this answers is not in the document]
suevalov: Got it

suevalov (approved):

jelz (approved):

anho (approved):

PR #478, opened 2020-10-07
feat: expose tag endpoints in plain client

anho:


suevalov (approved):

marcolink (approved):
> Looks like changing of `update` and `delete` methods for tag instance is still missing here:
> 
> https://github.com/contentful/contentful-management.js/blob/experimental/lib/entities/tag.ts

Hi @suevalov what exactly do you mean by "changing"?
Is there something we're missing on the master branch?

This PR looks good btw 🎉 🚢

anho:
@marcolink I just missed to use the added plain endpoint in one specific place which was the `createTagApi` inside `lib/entities/tag.ts`.

PR #506, opened 2020-10-20
feat: expose environment aliases

suevalov:
* Exposes environment aliases endpoints to a new plain CMA client
* Adds new environment sys properties


lib/plain/endpoints/environment-alias.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as raw from './raw'
+import cloneDeep from 'lodash/cloneDeep'
+import {
+  EnvironmentAliasProps,
+  CreateEnvironmentAliasProps,
+} from '../../entities/environment-alias'
+import { CollectionProp, PaginationQueryParams, GetSpaceParams } from './common-types'
+
+type GetSpaceEnvAliasParams = GetSpaceParams & { environmentAliasId: string }
+
```
anho, line 11: I like the `getBaseUrl` pattern we started to introduce and would love to see it in here too.

lib/create-space-api.ts
```diff
@@ -940,10 +940,14 @@ export default function createSpaceApi({ http }: { http: AxiosInstance }) {
      * .catch(console.error)
      * ```
      */
-    createEnvironmentAliasWithId(id: string, data: Omit<EnvironmentAliasProps, 'sys'>) {
-      return http
-        .put('environment_aliases/' + id, data)
-        .then((response) => wrapEnvironmentAlias(http, response.data), errorHandler)
+    createEnvironmentAliasWithId(
+      environmentAliasId: string,
+      data: Omit<EnvironmentAliasProps, 'sys'>
```
anho, line 945: Could reuse `CreateEnvironmentAliasProps` here.

anho (approved):
Just added minor consistency ideas.

PR #533, opened 2020-11-05
fix(test): on cleanup explicitly target environment instead of alias

ruderngespra:
On cleanup explicitly target environment instead of alias.

To unblock https://github.com/contentful/contentful-management.js/pull/530


marcolink (approved):
:shipit:

PR #537, opened 2020-11-06
feat: extend scheduled actions type

Spring3:


lib/entities/scheduled-action.ts
```diff
@@ -28,12 +28,13 @@ enum ScheduledActionStatus {
...
 }
 
-type SchedulableEntityType = 'Entry'
+type SchedulableEntityType = 'Entry' | 'Asset' | 'Release'
 type SchedulableActionType = 'publish' | 'unpublish'
+type ScheduledActionType = 'ScheduledAction' | 'ReleaseAction'
 
 export type ScheduledActionSysProps = {
   id: string
-  type: 'ScheduledAction'
```
freakyfelt, line 36: This should remain ScheduledAction. ReleaseAction should be a separate shape for when a release is executed

freakyfelt (changes requested):

lib/entities/scheduled-action.ts
Spring3: Oh I see. Ok, thanks for noting that

freakyfelt (approved):
🚀

suevalov (approved):

PR #540, opened 2020-11-11
feat: add editors and sidebar to EditorInterface type

Jwhiles:


FLoppix:
So it was just a type issue ? And the API calls are working as expected?

Jwhiles:
> So it was just a type issue ? And the API calls are working as expected?

Yep! I guess this was only an issue for people using TypeScript

suevalov (approved):

PR #544, opened 2020-11-13
Expose space member and membership

suevalov:


anho (approved):
Looks about right.

vbeleuta (approved):
This looks great! 👍  The only thought I had is that we might be missing the SpaceMemberships getMany by orgId (in the create-organization-api), which probably also need to be added in the PROGRESS file. Ignore me if you wanted to handle them in a different PR.

suevalov:
@vbeleuta We're going to tackle this in a separate PR

PR #545, opened 2020-11-16
feat: expose team and membership endpoints

anho:


lib/create-organization-api.ts
```diff
@@ -181,12 +181,15 @@ export default function createOrganizationApi({ http }: { http: AxiosInstance })
      * ```
      */
     getTeams(query: QueryOptions = {}) {
-      return http
-        .get('teams', createRequestConfig({ query }))
-        .then((response) => wrapTeamCollection(http, response.data), errorHandler)
+      const raw = this.toPlainObject() as OrganizationProp
+
+      return endpoints.team
+        .getMany(http, { organizationId: raw.sys.id, query })
```
suevalov, line 187: Shouldn't we pass query thought `query: createRequestConfig({ query }).params,` as we do in many other places?

lib/create-organization-api.ts
```diff
@@ -245,16 +252,20 @@ export default function createOrganizationApi({ http }: { http: AxiosInstance })
...
-      if (opts.teamId) {
-        return http
-          .get('teams/' + opts.teamId + '/team_memberships', createRequestConfig({ query }))
-          .then((response) => wrapTeamMembershipCollection(http, response.data), errorHandler)
+      const { teamId, query } = opts
+      const raw = this.toPlainObject() as OrganizationProp
+
+      if (teamId) {
+        return endpoints.teamMembership
+          .getManyForTeam(http, { organizationId: raw.sys.id, teamId, query })
```
suevalov, line 260: the same as above

lib/create-organization-api.ts
```diff
@@ -318,6 +329,7 @@ export default function createOrganizationApi({ http }: { http: AxiosInstance })
      * ```
      */
     getOrganizationSpaceMembership(id: string) {
+      // TODO need to move
```
suevalov, line 332: Do we need to move this method somewhere or just add it to plain client?

lib/create-organization-api.ts
anho: Good catch 👍

lib/create-organization-api.ts
anho: It's not present yet in one of the plain endpoint objects. Let's implement on what we agreed upon and then do another round of checking whats left.

vbeleuta: yes, i think this can do under space memberships, getForOrganization maybe?

lib/plain/endpoints/common-types.ts
```diff
@@ -3,6 +3,7 @@ import { QueryOptions, PaginationQueryOptions } from '../../common-types'
 export type GetSpaceParams = { spaceId: string }
 export type GetSpaceEnvironmentParams = { spaceId: string; environmentId: string }
 export type GetOrganizationParams = { organizationId: string }
+export type GetTeamParams = { organizationId: string; teamId: string }
```
vbeleuta, line 6: not exactly related to this PR, but why do we name these types as Get*Params? I would suggest just SpaceParams for ex
Update: So I think you name it like that because they are params for the GET calls yes? 🤔

vbeleuta (approved):
nice! 💯  just a question about some types naming

lib/plain/endpoints/common-types.ts
anho: @vbeleuta correct, types are present to ease re-use and naming is because the sole purpose is for typing the many get calls.

suevalov (approved):

PR #546, opened 2020-11-16
feat: expose all left org related endpoints

suevalov:
Finishes exposing all existing endpoints to a plain client.


lib/plain/endpoints/team-space-membership.ts
```diff
@@ -0,0 +1,86 @@
...
+    `/organizations/${params.organizationId}/team_space_memberships/${params.teamSpaceMembershipId}`
+  )
+}
+
+export const getManyForOrganization = (
+  http: AxiosInstance,
+  params: GetOrganizationParams & QueryParams & { teamId?: string }
+) => {
+  const query = params.query || {}
+  if (params.teamId) {
```
anho, line 40: I'd rather leave this magic in the organization api of the old client.

lib/entities/team-membership.ts
```diff
@@ -69,25 +73,23 @@ export interface TeamMembership extends TeamMembershipProps, DefaultElements<Tea
...
-        })
-        .then((response) => wrapTeamMembership(http, response.data), errorHandler)
+      return endpoints.teamMembership
+        .update(http, getParams(raw), raw)
+        .then((data) => wrapTeamMembership(http, data))
     },
 
-    delete: function () {
+    delete: function del() {
       const raw = this.toPlainObject()
```
vbeleuta, line 91: i believe should be: `const raw = this.toPlainObject() as TeamMembershipProps`

vbeleuta (approved):
looks great 👍  just a small comment on team membership delete

anho (commented):
Thanks for fixing `entities/team*` stuff.

lib/plain/endpoints/team-space-membership.ts
suevalov: I wouldn't say it's magic but rather a part of the how API works. We didn't leave any implementation details in the official client in other methods, so I'd rather be consistent here.

lib/entities/team-membership.ts
suevalov: good call. fixed.

lib/plain/endpoints/team-space-membership.ts
anho: I think it's something different to what we provide in other places. But no strong opinion on this one.

anho (approved):

PR #555, opened 2020-11-17
feat!: drop all deprecated methods and introduce a new optional plain client

suevalov:
## External changes:

BREAKING CHANGE: All deprecated methods are removed. Improved Typescript support. New types have breaking several small changes that are breaking the compatibility with the previous version.

The following deprecated methods were ❗ **removed** ❗ :

```javascript
space.getContentType(...)
space.getContentTypes(...)
space.createContentType(...)
space.createContentTypeWithId(...)
space.getEditorInterfaceForContentType(...)
space.getEntry(...)
space.getEntries(...)
space.createEntry(...)
space.createEntryWithId(...)
space.createUpload(...)
space.createAsset(...)
space.getAsset(...)
space.getAssets(...)
space.createAssetId(...)
space.createAssetFromFiles(...)
space.getUpload(...)
space.getLocale(...)
space.getLocales(...)
space.createLocale(...)
space.getUiExtension(...)
space.getUiExtensions(...)
space.createUiExtension(...)
space.createUiExtensionWithId(...)
space.getEntrySnapshots(...)
space.getContentTypeSnapshots(...)
```

These methods were deprecated since `contentful-management@5` and now we're fully removing them.

Note that all listed methods are still 🥳  available 🥳  for `environment` namespace:

```javascript
const space = await client.getSpace('space-id');
const environment = await space.getEnvironment('master');
const entry = await environment.getEntry('entry-id');
```

### New alternative plain API

Starting `contentful-management@7` this library provides an alternative plain SDK which exposes all CMA endpoints in a simple flat manner oppose to a default waterfall structure.

```javascript
const contentful = require('contentful-management')
const plainClient = contentful.createClient(
  {
    // This is the access token for this space. Normally you get the token in the Contentful web app
    accessToken: 'YOUR_ACCESS_TOKEN',
  },
  { type: 'plain' }
)

const environment = await plainClient.environment.get({
  spaceId: '<space_id>',
  environmentId: '<environment_id>',
})

const entries = await plainClient.entry.getMany({
  spaceId: '123',
  environmentId: '',
  query: {
    skip: 10,
    limit: 100,
  },
})

// With scoped space and environment
const scopedPlainClient = contentful.createClient(
  {
    // This is the access token for this space. Normally you get the token in the Contentful web app
    accessToken: 'YOUR_ACCESS_TOKEN',
  },
  {
    type: 'plain',
    defaults: {
      spaceId: '<space_id>',
      environmentId: '<environment_id>',
    },
  }
)

// entries from '<space_id>' & '<environment_id>'
const entries = await scopedPlainClient.entry.getMany({
  query: {
    skip: 10,
    limit: 100,
  },
})
```

The benefits of using the plain version of SDK are:

- The ability to reach any possible CMA endpoint without the necessity to call any async functions beforehand.
  - It's especially important if you're using this CMA client for non-linear scripts (for example, a complex Front-end application)
- All returned objects are simple Javascript objects without any wrappers. They can be easily serialized without an additional `toPlainObject` function call.

## Internal changes

A complete revamp of internal implementation of the client, which allows to expose `plain` version of the CMA client.


lib/contentful-management.ts
```diff
@@ -17,10 +27,34 @@ import { createCMAHttpClient, ClientParams } from './create-cma-http-client'
...
-  const api = createContentfulApi({
-    http: http,
-  })
-  return api
+
+function createClient(params: ClientParams): ClientAPI
+function createClient(
+  params: ClientParams,
+  opts: {
+    plainClient?: true
```
anho, line 35: It will only work if `planclient: true` is given, so I'd remove the question mark.

anho (commented):
I'd add a bare minimum paragraph to the README as part of this PR.

lib/contentful-management.ts
suevalov: Change to `type: 'plain'`

suevalov:
@anho Updated README.md

anho (approved):
Awesome 🚀

djagya (approved):
Congrats everyone working on it!

phoebeschmidt (changes requested):
Overall I approve the changes, nice work team and the several people involved who helped prepare the SDK for this!

I'd maybe add a bit of context in the README (and the release notes?) for our motivation for these changes. Why did we decide to go this route? advantages or scoped or unscoped clients?

suevalov:
Good point @phoebeschmidt!

I added a new paragraph below the example both in README and the description of the PR. Once the new package is published I'll update the description of the release with the description of this pull request.

phoebeschmidt (approved):

PR #558, opened 2020-11-20
Expose entity typings on package entrypoint

zcei:
[As mentioned in the Community](https://www.contentfulcommunity.com/t/using-typings-from-typescript/4932), with new v7 it's a bit annoying to import types for entities.

What you need to do right now:

```typescript
import { Entry } from 'contentful-management/dist/typings/entities/entry';
```

What we should be able to do: 

```typescript
import { Entry } from 'contentful-management'
// ... or ...
import { Entry } from 'contentful-management/entities'
```

Currently the former is implemented, but if the latter is preferred this can be done with a static file in repo root that re-exports the built `entity-types.ts`.


lib/entities/entity-types.ts
```diff
@@ -0,0 +1,30 @@
+export type { ApiKey } from './api-key'
```
freakyfelt, line 1: Any reason not to/is it possible to just export * types from each file?

freakyfelt: I'm picturing scenarios where I need to dig in and use a specific inner type for an entity (e.g. the EntrySys)

zcei: Wasn't sure about potential clashes between all those files and `export type` didn't give me the asterisk notation. So I thought a sensible selection might make more sense. But you're right, this makes it harder to obtain shapes like `ApiKeyProps`.

Giving it a shot with `export * from` right now. But might make sense to revisit the `index.ts` then as it exports everything under the default export.

@suevalov is there a reason for `entities/index.ts` default export behavior?

anho: @zcei That was just convenience on the internal consumption inside the `lib/create-*-api.ts` files. Changing the `lib/entities/index.ts` exports would require some additional adjustments internally but it can be done with relative ease.

suevalov:
I'd probably go with the latter approach and expose it as `'contentful-management/types'`

lib/entities/entity-types.ts
suevalov: We'd also need to export all `*Props` types, cause they are actually describing the actual API responses without any additional wrappers on it.

suevalov:
Maybe even `'contentful-management/types'` & `'contentful-management/plain-types'`

lib/entities/entity-types.ts
aleksd-dfds: From the user point of view, we don't have a big prefrence, or know the distinction of "inner types" / "outer types". 
We mainly want import {Entries, EntryProps} from 'contentful-managment/types' or something like that. If you have groupings, that is also fine => anything is better than the current way.

zcei:
@suevalov  is taking this one over in a new PR from within this repo than a fork (for CI) and to add the plain-types idea he mentioned above

PR #559, opened 2020-11-20
feat: a single entry for all external type definitions

suevalov:
Adds ability to consume all external Typescript types using `contentful-management/types`

For example:

```typescript
import type { SpaceProps } from 'contentful-management/types';
```


lib/entities/content-type-fields.ts
```diff
@@ -1,19 +1,19 @@
...
 }
 
-export interface RegExp {
+interface RegExp {
   pattern: string
   flags: string
 }
 
-export interface Validation {
+export interface ContentFieldValidation {
```
zcei, line 16: How about `ContentTypeFieldValidation` or do we use use `ContentField` uniform to reference a content type's field?

suevalov: I'm fine with `ContentTypeFieldValidation` 👍

anho (approved):

zcei (approved):

PR #561, opened 2020-11-22
feat: additional raw endpoint for easy access to all HTTP verbs

anho:
`contentful-migration` exposes a way to access to all HTTP verbs easily to make the shift easier this is reflected in the plain endpoint.


marcolink:
@anho should we reflect this addition in documentation in some way?

anho:
@marcolink we haven't documented anything about the plain client so far, so I have no clue where you want to put that kind of information :)

marcolink (approved):
:shipit:

suevalov (approved):

PR #568, opened 2020-11-23
chore: clean experimental configuration

suevalov:
Removes semantic release configuration for `experimental` branch


.circleci/config.yml
```diff
@@ -64,4 +64,3 @@ workflows:
                 - master
                 - next
```
anho, line 65: We want to keep `next` and `next-major` around?

suevalov: Yep, they were there before we started to work on `experimental`, so it's fair to leave them here

anho (approved):

PR #570, opened 2020-11-23
feat: extensiblity types

jelz:
## Summary

Sync types with API schemas

When adding a new method:


lib/entities/app-definition.ts
```diff
@@ -72,11 +43,11 @@ export type AppDefinitionProps = {
   /**
    * URL where the root HTML document of the app can be found
    */
-  src: string
+  src?: string
   /**
    * Locations where the app can be installed
    */
-  locations: AppLocation[]
+  locations?: AppLocation[]
```
Jwhiles, line 50: It looks like locations is always defined, but can be empty.

Jwhiles: Or at least that's the case when you update through the user interface

lib/entities/editor-interface.ts
```diff
@@ -8,24 +8,58 @@ import * as endpoints from '../plain/endpoints'
...
-  widgetId: string
-  widgetNamespace: string
   settings?: Record<string, any>
 }
 
 export interface Editor {
+  /**
+   * Type of the widget used
+   */
+  widgetNamespace: string
```
Jwhiles, line 32: Should we limit this to the specific values that are possible?

suevalov (approved):
Splendid!

P.S. Should it be `fix: extensibility types` or `feat: extensibility types`?

Jwhiles (approved):

jelz:
@suevalov done, but I still cannot merge, can you?

PR #583, opened 2020-11-27
fix: correctly pass params to plain client for getSpaces

anho:
## Summary

Pass down params properly.


TimBeyer (approved):
I feel like we should just get rid of `createRequestConfig` because the contents look sort of nonsensical from https://github.com/contentful/contentful-sdk-core/blob/fe7d2f73e02e61b43762d75601bffee64ef6e223/lib/create-request-config.js

But for now looks good.

Issue #576, opened 2020-11-25
Incorrect Typescript definitions


nekosaur:
1. `createSpaceMembership` on `Space` incorrectly requires a `name` property when it should be `email`
```js
// CURRENT
space.createSpaceMembership({
  admin: true,
  name: '???',
  roles: [],
});

// CORRECT
space.createSpaceMembership({
  admin: true,
  email: 'hello@world.com',
  roles: [],
});
```
Additionally, `roles` is marked as required when it is not if using REST API.

2. `createLocale` on `Environment` incorrectly requires `internal_code`
```js
env.createLocale({
  code: locale.code,
  name: locale.name,
  fallbackCode: null,
  internal_code: '???' // This is not required through REST API, and things work correctly if ignoring the TS error
})
```

suevalov:
Thanks for reporting @nekosaur, we'll release a fix shortly

PR #586, opened 2020-11-30
fix: proper types for create space-membership and create locale methods

suevalov:
## Summary

Fixes https://github.com/contentful/contentful-management.js/issues/576

When adding a new method:

### DIFF
--- a/lib/entities/space-membership.ts
+++ b/lib/entities/space-membership.ts
@@ -3,20 +3,19 @@ import cloneDeep from 'lodash/cloneDeep'
 import { freezeSys, toPlainObject } from 'contentful-sdk-core'
 import enhanceWithMethods from '../enhance-with-methods'
 import { wrapCollection } from '../common-utils'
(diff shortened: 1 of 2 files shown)


lib/entities/space-membership.ts
```diff
@@ -3,20 +3,19 @@ import cloneDeep from 'lodash/cloneDeep'
...
+  sys: MetaSysProps & { space: SysLink; user: SysLink }
+  user: SysLink
   admin: boolean
-  roles: { sys: MetaLinkProps }[]
+  roles: SysLink[]
 }
 
-export type CreateSpaceMembershipProps = Omit<SpaceMembershipProps, 'sys'>
+export type CreateSpaceMembershipProps = Omit<SpaceMembershipProps, 'sys' | 'user'> & {
+  email: string
```
anho, line 17: `email` should be in SpaceMembershipProps too?

suevalov: apparently, not :)

TimBeyer (approved):

Issue #584, opened 2020-11-29
Given v7.* when createUpload() or createAssetFromFiles() then 404 


TjenWellens:
It seems a lot like what they mentioned in this thread:
[createAssetFromFiles upload bug #165](https://github.com/contentful/contentful-management.js/issues/165)

When I try
```
const fs = require('fs');
const uploadStream = fs.createReadStream("some-file.png");

async function main() {
  const space = await client.getSpace(process.env.CONTENTFUL_SPACE)
  const environment = await space.getEnvironment(process.env.CONTENTFUL_ENVIRONMENT)
  console.log(environment)
  environment.createUpload({file: uploadStream})
	.then(upload => {
		console.log(upload)
	})
	.catch(error => {
		console.error(error)
	})
}
main()
```
on `v7.3.1` I get the error
```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'ClientRequest'
    |     property 'socket' -> object with constructor 'TLSSocket'
    --- property '_httpMessage' closes the circle
    at JSON.stringify (<anonymous>)
    at errorHandler (.../node_modules/contentful-management/dist/contentful-management.node.js:13750:24)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```
When I go in to `contentful-management.node.js:13750`
add a line of logging before it crashes on the JSON.stringify cycle
```
error.name = errorName && errorName !== 'Unknown' ? errorName : `${response === null || response === void 0 ? void 0 : response.status} ${response === null || response === void 0 ? void 0 : response.statusText}`;
console.error(errorData) // I added this
error.message = JSON.stringify(errorData, null, '  ');
```
Then I see the actual error message:
```
{
  status: 404,
  statusText: 'Not Found',
  message: 'Not Found',
  details: {},
  request: {
    url: '/spaces/<my-space-id>/environments/master/uploads', // anonymized my-space-id
    headers: {
      Accept: 'application/json, text/plain, */*',
      'Content-Type': 'application/octet-stream',
      'X-Contentful-User-Agent': 'sdk contentful-management.js/7.3.1; platform node.js/v14.7.0; os macOS/19.6.0;',
      Authorization: 'Bearer ...<my-token>', // anonymized
      'user-agent': 'node.js/v14.7.0',
      'Accept-Encoding': 'gzip'
    },
    method: 'post',
    payloadData: ReadStream {
      ... // removed some stuff
    }
  }
}
```

So thats a problem...

[...]

PR #589, opened 2020-11-30
fix: correct file uploads

suevalov:
## Summary

Fixes https://github.com/contentful/contentful-management.js/issues/584

## Motivation and Context

Apparently, we did a mistake while migrating to a new plain API and scoped uploads to environments while upload are bound only to a space. Integration tests don't cover this part so it was never caught.

When adding a new method:

### DIFF
--- a/lib/error-handler.ts
+++ b/lib/error-handler.ts
@@ -1,4 +1,5 @@
 import isPlainObject from 'lodash/isPlainObject'
+import get from 'lodash/get'
 import { AxiosError } from 'axios'
 
 /**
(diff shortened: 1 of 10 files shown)


lib/error-handler.ts
```diff
@@ -67,6 +67,11 @@ export default function errorHandler(errorResponse: AxiosError): never {
   const error = new Error()
   error.name =
     errorName && errorName !== 'Unknown' ? errorName : `${response?.status} ${response?.statusText}`
-  error.message = JSON.stringify(errorData, null, '  ')
+
+  try {
```
anho, line 71: Intentional?

suevalov: yes, as were mentioned in https://github.com/contentful/contentful-management.js/issues/584 if JSON.stringify fails we don't actually know what the real error was at first place

anho: Ok, then let's find something that still conveys the minimal amount of information about the error instead of completely silencing it.

suevalov: I updated `catch` branch

anho (approved):

PR #593, opened 2020-12-02
refactor: replace lodash with fast-copy

suevalov:
## Summary

* Replaces `lodash/cloneDeep` with `fast-copy`
* Replaces `lodash/isPlainObject` with `lodash.isplainobject` (300 bytes)
* Replaces all other lodash methods with native ES6


marcolink (approved):
LGTM

PR #598, opened 2020-12-08
feat: expose raw.bareRequest from plain client

suevalov:
## Summary

* Propagates generics to all `client.raw` methods
* Exposes a new `raw. bareRequest` method that will be used in 401 interceptor

When adding a new method:


giotiskl (approved):

marcolink (commented):
LGTM

PR #609, opened 2020-12-17
fix: allow required headers to be overwritten by custom ones

suevalov:
## Summary

To be able to override user agent when necessary (for example in the main Contentful web application)

When adding a new method:


marcolink (approved):
LGTM

anho (approved):

PR #610, opened 2020-12-18
feat: be able to directly delete an entry by its Id

fbouchezexagroup:
## Summary

Be able to directly delete an entry by its Id
Avoid using the plain client for that purpose. 

## Motivation and Context

I need to delete an entry which I already now its sys.id
With this new method, I can delete the entry without calling  getEntry(id)  and then entry.delete() which produce 2 API calls

### DIFF
--- a/lib/create-cma-http-client.ts
+++ b/lib/create-cma-http-client.ts
@@ -12,7 +12,7 @@ export type ClientParams = CreateHttpClientParams & {
   /**
    * Contentful CMA Access Token
    */
-  accessToken: string
+  accessToken: CreateHttpClientParams['accessToken']
   /**
    * API host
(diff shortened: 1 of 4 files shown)


marcolink:
@fbouchezexagroup Thanks for your contribution 🎉 

The build is currently failing on the liniting step. 
Would you mind running `npm run lint` and `npm run prettier` once to make sure this step passes?

Once the build is passing, I'm happy to merge in the changes!

phoebeschmidt:
@fbouchezexagroup if you merge @ruderngespra's [PR](https://github.com/exaprint/contentful-management.js/pull/1) then we would be happy to merge this PR asap :)

PR #661, opened 2021-01-28
fix: improve generics in raw plain client endpoints

suevalov:
## Summary

Makes it easier to type raw methods in plain client version.


anho (approved):
I like it 👍

PR #662, opened 2021-01-28
fix: dowgrade typedoc version

suevalov:
## Summary

Solves problem with publishing and releasing from the main branch.


anho:
You want to ignore the package in dependabot?

suevalov:
yes, can I do it in the configuration? @anho

anho:
https://dependabot.com/docs/config-file/#ignored_updates should do the trick. Maybe with a comment why this update is ignored.

anho (approved):

PR #669, opened 2021-02-03
feat: optional params for user.getCurrent()

Spring3:
## Summary

I have noticed that optional params that were added to uset.getCurrent() weren't exposed and typed.
In this PR we fix that

## Description

I was  not able to use wrap() wrapper because that would add unnecessary default params, which aren't supported by the getCurrent() function.
I was not able to use werapHttp() because it didn't support any params, and adjusting it wasn't possible due to genric types, so I ended up with a custom wrapping types

When adding a new method:

### DIFF
--- a/lib/plain/plain-client.ts
+++ b/lib/plain/plain-client.ts
@@ -65,7 +65,9 @@ export const createPlainClient = (params: ClientParams, defaults?: DefaultParams
     user: {
       getManyForSpace: wrap(wrapParams, endpoints.user.getManyForSpace),
       getForSpace: wrap(wrapParams, endpoints.user.getForSpace),
-      getCurrent: wrapHttp(http, endpoints.user.getCurrent),
+      getCurrent: (...args: RestParamsType<typeof endpoints.user.getCurrent>) => {
+        return endpoints.user.getCurrent(http, ...args)
+      },
       getForOrganization: wrap(wrapParams, endpoints.user.getForOrganization),
       getManyForOrganization: wrap(wrapParams, endpoints.user.getManyForOrganization),
     },


lib/plain/endpoints/user.ts
```diff
@@ -7,8 +7,8 @@ export const getForSpace = (http: AxiosInstance, params: GetSpaceParams & { user
   return raw.get<UserProps>(http, `/spaces/${params.spaceId}/users/${params.userId}`)
 }
 
-export const getCurrent = <T = UserProps>(http: AxiosInstance, params?: QueryParams) =>
-  raw.get<T>(http, `/users/me`, { params: params?.query })
+export const getCurrent = (http: AxiosInstance, params?: QueryParams) =>
+  raw.get<UserProps>(http, `/users/me`, { params: params?.query })
```
freakyfelt, line 11: Is there a reason we are reverting this? This allows for us to specify alternative shapes for the response so we can override it in consumers of the client

Spring3: sounds fair. reverting this

Silhoue (approved):

lib/plain/plain-client.ts
```diff
@@ -65,7 +65,9 @@ export const createPlainClient = (params: ClientParams, defaults?: DefaultParams
     user: {
       getManyForSpace: wrap(wrapParams, endpoints.user.getManyForSpace),
       getForSpace: wrap(wrapParams, endpoints.user.getForSpace),
-      getCurrent: wrapHttp(http, endpoints.user.getCurrent),
+      getCurrent: (...args: RestParamsType<typeof endpoints.user.getCurrent>) => {
```
freakyfelt, line 68: For my own education, is there a reason we have to diverge from the `getForSpace`, etc from above?

freakyfelt (approved):

lib/plain/plain-client.ts
Spring3: because wrap() applies default params that we don't want.
```
{
   spaceId,
   environmentId,
   organizationId
}
```

PR #726, opened 2021-02-24
feat: add base structure for adapters

andipaetzold:
## Summary

At the moment all calls of the library result in HTTP (REST) calls to the CMA. We want to enable the library to resolve method calls also over different communication channels/protocols (e.g. using GraphQL).

To achieve this goal, the HTTP/REST requests need to be abstracted in order to be replaced. For that reason, we introduce `Adapter`s.

This feature will be introduced without a breaking change.

## Description

An `Adapter` is currently (WIP) defined as 

```typescript
interface Adapter {
  makeRequest<R = unknown>(options: MakeRequestOptions): Promise<R>
}

export interface MakeRequestOptions {
  entityType: string
  action: string
  params?: Record<string, unknown>
  payload?: Record<string, unknown>
  headers?: Record<string, unknown>
}
```
All calls of the library are piped through `makeRequest`.

An adapter can be specified when creating the CMA client using `createClient`. The parameter `adapter` is unfortunately already used by Axios adapters (https://github.com/contentful/contentful-sdk-core/blob/402769b274f61a48a1b16e408846959aaa0f74bc/src/types.ts#L44-L45)

```typescript
const client = createClient({
  cmaAdapter: new RestAdapter({
    accessToken: 'token'
  })
})
```

If the client is initialized without specifying adapters, a `RestAdapter` is automatically created using the provided parameters.

In the current state of the PR, only space methods are implemented using the new adapters. The endpoints (`lib/plain/endpoints`) will move to `lib/adapters/REST/endpoints` and will therefore be an http specific implementation of the methods. In `lib/plain/plain-client.ts` we only map the existing methods to the corresponding `makeRequest` calls.

Unfortunately, the `PlainClientAPI` type cannot be properly inferred anymore and therefore needs to be explcitely typed. Type inference/type safety is a general problem atm as TypeScript cannot properly check whether the types returned from `lib/adapters/REST/endpoints/space`-methods equals what is defined in `PlainClientAPI`. But maybe I find an elegant way to solve that issue.

When adding a new method:


lib/create-contentful-api.ts
```diff
@@ -42,11 +45,13 @@ export default function createClientApi({ http }: { http: AxiosInstance }) {
     getSpaces: function getSpaces(
       query: QueryOptions = {}
     ): Promise<Collection<Space, SpaceProps>> {
-      return endpoints.space
-        .getMany(http, {
-          query: createRequestConfig({ query: query }).params,
+      return adapter
+        .makeRequest({
+          entityType: 'Space',
+          action: 'getMany',
```
shikaan, line 51: Maybe we can make this an `enum`?

lib/create-space-api.ts
```diff
@@ -27,7 +27,8 @@ export type ContentfulSpaceAPI = ReturnType<typeof createSpaceApi>
  * @prop {object} entities - Object with wrapper methods for each kind of entity
  * @return {ContentfulSpaceAPI}
  */
-export default function createSpaceApi({ http }: { http: AxiosInstance }) {
+export default function createSpaceApi({ adapter }: { adapter: Adapter }) {
+  const http = (adapter as RestAdapter).http
```
shikaan, line 31: Why is this here? Shouldn't we get rid of `http` and only use `makeRequest`? Is it maybe temporary?

shikaan (commented):
I am not sure about the name. Calling it `REST` makes me think that there will be, for instance, a `GraphQL` adapter, which doesn't really make sense here. Maybe this is more like an "AuthenticatedAdapter" or "AccessTokenAdapter".

At the end of the day what we are after here is abstracting the auth logic and I guess this should be captured by the name we give to this thing

lib/create-space-api.ts
andipaetzold: This is just to make sure the whole thing still compiles. This will be eventually removed.

lib/adapters/REST/rest-adapter.ts
```diff
@@ -0,0 +1,119 @@
...
+    return this.axiosInstances[clientType]!
+  }
+
+  public async makeRequest<R>({
+    entityType,
+    action,
+    params,
+    payload,
+    headers,
+    clientType,
```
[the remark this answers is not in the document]
andipaetzold: An Adapter can be used in multiple CMA clients. Therefore, we cannot store the client type in the adapter itself but pass it in for every request.

I am not happy with the solution myself as it feels strange to have an additional param just for a tracking header...

lib/create-contentful-api.ts
andipaetzold: In general I always try to stay away from enums as they are not vanilla JavaScript and typescript adds a ton of extra code.

Due to the overloadings of `makeRequest`, we have type safety here that would catch typos. Also, we have very specific actions here like: `processForLocale`, `createFromFiles` or `getManyForOrganization`. If it was only create/update/save/delete, an enum would probably be alright.

jelz:
> I am not sure about the name. Calling it REST makes me think that there will be, for instance, a GraphQL adapter, which doesn't really make sense here. Maybe this is more like an "AuthenticatedAdapter" or "AccessTokenAdapter".

Why it doesn't make sense to you @shikaan ? For me the reasoning was: "what do I talk to?". `RestAdapter` is for talking to Contentful REST Management API (CMA, api.contentful.com), `PostMessageChannelAdapter` is for talking to `postMessage` channel, `GraphQLAdapter` will be for talking to Contentful GraphQL management API (when/if we've got one).

lib/create-adapter.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export type ClientParams = RestAdapterParams & AdapterParams
+
+/**
+ * @private
+ */
+export function createAdapter(params: ClientParams): Adapter {
+  if (params.cmaAdapter) {
+    return params.cmaAdapter
```
jelz, line 20: Should be just `params.adapter`

lib/create-contentful-api.ts
jelz: I'd also be in favor of an enum. Also: why do we want to have this specific actions? Assets are always processed for a locale, I don't know why I would add `ForOrganization` when getting many (parent entity is known, as in AppDefinition belongs to and org vs Entry belonging to an env), I don't know what `createFromFiles` refers to.

lib/adapters/REST/rest-adapter.ts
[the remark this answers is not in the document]
jelz: maybe let's pass `userAgent` from the outside?

jelz (commented):
LGTM so far, no red flags in the approach

lib/create-adapter.ts
andipaetzold: The parameter `adapter` is unfortunately already used by Axios adapters
https://github.com/contentful/contentful-sdk-core/blob/402769b274f61a48a1b16e408846959aaa0f74bc/src/types.ts#L44-L45

shikaan:
> > I am not sure about the name. Calling it REST makes me think that there will be, for instance, a GraphQL adapter, which doesn't really make sense here. Maybe this is more like an "AuthenticatedAdapter" or "AccessTokenAdapter".
> 
> Why it doesn't make sense to you @shikaan ? For me the reasoning was: "what do I talk to?". `RestAdapter` is for talking to Contentful REST Management API (CMA, api.contentful.com), `PostMessageChannelAdapter` is for talking to `postMessage` channel, `GraphQLAdapter` will be for talking to Contentful GraphQL management API (when/if we've got one).

I think I have just misunderstood the scope of the change then. I thought it was to guarantee different type of integrations with the REST API (which is still possible anyway) as opposed to different type of integrations with CMA in general. 

In that case, it makes total sense to me. Sorry about the noise 😅

lib/create-contentful-api.ts
andipaetzold: I was looking through all existing methods of the plain client (https://github.com/contentful/contentful-management.js/blob/master/lib/plain/plain-client.ts) and assumed (without looking at the actual implementation) each method would eventually be one action.

I was referring to `asset.createFromFiles`, `asset.processForAllLocales` and `usage.getManyForOrganization`.

I would be happy to use enums if all actions are available for all entity types. But in this case, the actions are very specific and sometimes only apply to a single entity type. Also, I don't see the benefit of using an enum here as these actions are all library internal and we can achieve the same level of type safety by using union types.

andipaetzold:
@jelz @shikaan To keep PRs small, this won't be the main branch/PR of that feature anymore but only one of many. Therefore this PR will merge into `EXT-2525` which is the main branch. I also reverted some other commits and put them into their own PRs. Otherwise, things will become very confusing.
Individual PRs won't go green as fixing the unit tests will only add extra work as some basic structure might still change during the development. But the individual branches will build successfully.

Since your reviews, I changed that `userAgent` is passed into `makeRequest`.

lib/create-adapter.ts
jelz: Maybe `apiAdapter`?

jelz (approved):
LGTM

shikaan (approved):

PR #734, opened 2021-03-01
feat: use makeRequest in plain client

andipaetzold:
This PR finishes the migration of the plain client that was started in https://github.com/contentful/contentful-management.js/pull/726/files#diff-210590c0799f06cb42e3dbc0ba7baf7546cafee353501cecb2516fb75026e793
Now all methods of the plain client call `adapter.makeRequest` instead of using the endpoints.

Every endpoint is now an entity type and every method is an action. I am not 100% sure whether this is what we want, but it was the easiest way to migrate, for now.

Until now the plain client was only implicitly typed by inferring types from the endpoints. This PR contains an explicit type for the client.

This PR builds successfully but does not work yet. Also, it is expected that the tests don't pass.


jelz (approved):

jelz:
> Every endpoint is now an entity type and every method is an action. I am not 100% sure whether this is what we want, but it was the easiest way to migrate, for now.

Not really how I thought it will be (it was a blurry thought... :D) but it turned out super nice.

lib/common-types.ts
```diff
@@ -148,7 +148,7 @@ export interface Adapter {
     params: GetSpaceParams
     userAgent: string
   }): Promise<any>
-  // makeRequest<R = unknown>(options: MakeRequestOptions): Promise<R>
+  makeRequest(options: MakeRequestOptions): Promise<any>
```
shikaan, line 151: q: what was the problem with unknown here?

shikaan (approved):
🎉

lib/common-types.ts
andipaetzold: We get type issues in the plain client as `unknown` is not assignable to another type.
e.g. this fails:
```typescript
const varUnknown: unknown = 5;
const varString: string = varUnknown;
```

PR #739, opened 2021-03-03
feat: use makeRequest in nested client

andipaetzold:
Please review https://github.com/contentful/contentful-management.js/pull/734 first.

This PR removes all usages of the endpoints within the nested client and replaces them with `makeRequest` calls. Similar to the plain client PR, the endpoints and methods have been migrated to `entityType` and `action`.

All endpoint files are moved to `adapters/REST/endpoints`. I didn't touch/adjust the code besides fixing the type.

This PR only builds but doesn't work yet. Aligning the `makeRequest`-calls of the plain and nested client and adjusting the endpoint implementations, will happen in the next PR.

Sorry for the big PR, but it wasn't easily possible to split it up. Many changes are just fixing the imports.


jelz (approved):
:sparkles:

shikaan (approved):

PR #742, opened 2021-03-04
feat(tags): create with visibility

marcolink:
Allow creation of "public" tags with a default fallback to "private"

### DIFF
--- a/lib/create-environment-api.ts
+++ b/lib/create-environment-api.ts
@@ -1181,15 +1181,12 @@ export default function createEnvironmentApi({ http }: { http: AxiosInstance }) 
         .then((data) => wrapSnapshotCollection<ContentTypeProps>(http, data))
     },
 
-    createTag(id: string, name: string, visibility: TagVisibility = 'private') {
+    createTag(id: string, name: string, visibility?: TagVisibility) {
       const raw = this.toPlainObject() as EnvironmentProps
       const params = { spaceId: raw.sys.space.sys.id, environmentId: raw.sys.id, tagId: id }
 
       return endpoints.tag
-        .createWithId(http, params, {
-          name,
-          sys: { visibility },
-        })
+        .createWithId(http, params, { name }, visibility)
         .then((data) => wrapTag(http, data))
     },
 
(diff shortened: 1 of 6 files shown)


makinwab (approved):
LGTM. Any docs or examples we should be updating asides from the release note?

damienxy (approved):

lib/create-environment-api.ts
```diff
@@ -1181,15 +1181,12 @@ export default function createEnvironmentApi({ http }: { http: AxiosInstance })
...
+    createTag(id: string, name: string, visibility?: TagVisibility) {
       const raw = this.toPlainObject() as EnvironmentProps
       const params = { spaceId: raw.sys.space.sys.id, environmentId: raw.sys.id, tagId: id }
 
       return endpoints.tag
-        .createWithId(http, params, {
-          name,
-          sys: { visibility },
-        })
+        .createWithId(http, params, { name }, visibility)
```
makinwab, line 1189: Would it be better to have the last option as a headers object instead?

marcolink: @burakukula Not sure, it would change the API to (if you want to set the visibility via the `plainClient`):
```javascript
.createWithId(http, params, {name}, { 'X-Contentful-Tag-Visibility': visibility })
```

What do you think?

makinwab: Ah, what I had in mind was: `{ visibility }` and then the method knows which header key to use.

makinwab: In the `createWithId` signature, the last param can be named `headers`.

marcolink: Ah I see, so we would have an object with  `{visibility: 'public' | 'private'}` and would transform the key to `X-Contentful-Tag-Visibility`?

marcolink: @makinwab I think I personally prefer to just have the visibility value there, but not a strong opinion :)

makinwab: My comment is non-blocking, actually. Approved.

PR #752, opened 2021-03-11
fix: adjust unit tests to new adapter architecture

andipaetzold:
This PR fixes the existing unit tests and makes them work with the new adapter architecture.

This PR does not add new tests. There will be a follow-up PR for that.

It is expected that the integration tests fail. They will be fixed when merging in `master`.


lib/adapters/REST/endpoints/raw.ts
```diff
@@ -53,6 +53,7 @@ export function del<T = any>(http: AxiosInstance, url: string, config?: AxiosReq
     .then((response) => response.data, errorHandler)
 }
 
+// TODO: this is broken
```
shikaan, line 56: What is broken? Would it be part of a followup?

shikaan (approved):
💣

lib/adapters/REST/endpoints/raw.ts
andipaetzold: Oh - forgot about that comment.
Yes, I'll fix that in a follow up PR. The whole `http` method is broken if I remember correctly.

PR #757, opened 2021-03-15
fix(tests): Improve space membership integration test

damienxy:
## Summary

Improve space-membership integration test by creating new space from scratch instead of using default space

When adding a new method:


marcolink (approved):
🚀

PR #760, opened 2021-03-19
test: add tests for new adapter architecture

andipaetzold:
Add tests for `createClient` & `createAdapter`.
Also, this PR fixes an issue when using `client.rawRequest`.

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -344,12 +344,12 @@ export default function createClientApi(makeRequest: MakeRequest) {
      * .catch(console.error)
      * ```
      */
-    rawRequest: function rawRequest(opts: AxiosRequestConfig & { url: string }) {
+    rawRequest: function rawRequest({ url, ...config }: AxiosRequestConfig & { url: string }) {
       return makeRequest({
         entityType: 'Http',
         action: 'request',
-        params: { url: opts.url, config: opts },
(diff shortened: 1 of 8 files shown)


lib/create-contentful-api.ts
```diff
@@ -344,12 +343,12 @@ export default function createClientApi(makeRequest: MakeRequest) {
      * .catch(console.error)
      * ```
      */
-    rawRequest: function rawRequest(opts: AxiosRequestConfig & { url: string }) {
+    rawRequest: function rawRequest({ url, ...config }: AxiosRequestConfig & { url: string }) {
       return makeRequest({
         entityType: 'Http',
         action: 'request',
-        params: { url: opts.url, config: opts },
-      }).then((response) => response.data, errorHandler)
```
FLoppix, line 352: Why did we remove this error handler here?

FLoppix (approved):
Just one comment otherwise LGTM

lib/create-contentful-api.ts
andipaetzold: The error handling is http specific and therefore part of the endpoint code: https://github.com/contentful/contentful-management.js/blob/8b3d421da8b1cf56d92ec15e3b9616e5c668a5fb/lib/adapters/REST/endpoints/raw.ts#L61

PR #770, opened 2021-03-23
feat: add support for api adapters [EXT-2525]

andipaetzold:
The current implementation of this library talks to the CMA via REST over HTTP. This PR adds `Adapter`s. These adapters add an abstraction layer that enables us to replace the existing REST communication to the server with a different technology or protocol.

An adapter can be specified via the `apiAdapter` param when calling `createClient`. All existing network request logic was moved into `RestAdapter`. If no adapter is specified, `RestAdapter` is used. Currently, there is no alternative adapter implementation.

This PR is not a breaking change! The PR should have no impact on the observable behavior of the library.

The new version was already published as pre-release (`@next`): 7.12.3


lib/adapters/REST/endpoints/api-key.ts
```diff
@@ -1,20 +1,28 @@
...
+) => {
   return raw.get<ApiKeyProps>(http, `/spaces/${params.spaceId}/api_keys/${params.apiKeyId}`)
 }
 
-export const getMany = (http: AxiosInstance, params: GetSpaceParams & QueryParams) => {
+export const getMany: RestEndpoint<'ApiKey', 'getMany'> = (
+  http: AxiosInstance,
+  params: GetSpaceParams & QueryParams
+) => {
   return raw.get<CollectionProp<ApiKeyProps>>(http, `/spaces/${params.spaceId}/api_keys`, {
```
FLoppix, line 20: Any reason we hardcode these URLs here instead of using functions to build these programmatically? For the `AppDefinition` we use something like `getBaseUrl`. Should we be consistent across the entities?

andipaetzold: Yes, the URL could be created programmatically.

I won't do it in this PR as I just copied the file: https://github.com/contentful/contentful-management.js/blob/master/lib/plain/endpoints/api-key.ts

shikaan (approved):
:shipit:

PR #774, opened 2021-03-26
feat(AppUpload): support AppUpload entity [EXT-2641]

FLoppix:
## Summary

Introducing the `AppBundle` entity with the following methods:

- create
- get
- delete

## Description

Added the entity to the flat and nested client and exported all relevant types. Also adjusted the `Organziation` API to support the `AppUpload` entity for the nested client.

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/app-upload.ts
+++ b/lib/adapters/REST/endpoints/app-upload.ts
@@ -1,4 +1,5 @@
 import type { AxiosInstance } from 'contentful-sdk-core'
+import fs from 'fs'
 import * as raw from './raw'
 import { GetAppUploadParams, GetOrganizationParams } from '../../../common-types'
 import { RestEndpoint } from '../types'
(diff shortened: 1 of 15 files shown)


lib/adapters/REST/endpoints/app-upload.ts
```diff
@@ -0,0 +1,51 @@
...
+  params: GetOrganizationParams,
+  payload: { file: string | ArrayBuffer | Stream }
+) => {
+  const httpUpload = getUploadHttpClient(http)
+
+  const { file } = payload
+
+  if (!file) {
+    return Promise.reject(new Error('Unable to locate a file to upload.'))
+  }
```
andipaetzold, line 44: How could this happen? And do we need an extra check for that?

We also don't check whether the organizationId is an empty string or undefined.

lib/adapters/REST/endpoints/app-upload.ts
```diff
@@ -0,0 +1,51 @@
...
+export const del: RestEndpoint<'AppUpload', 'delete'> = (
+  http: AxiosInstance,
+  params: GetAppUploadParams
+) => {
+  const httpUpload = getUploadHttpClient(http)
+
+  return raw.del<void>(httpUpload, getAppUploadUrl(params))
+}
+
+export const create: RestEndpoint<'AppUpload', 'create'> = async (
```
andipaetzold, line 33:
```suggestion
export const create: RestEndpoint<'AppUpload', 'create'> = (
```

lib/adapters/REST/endpoints/app-upload.ts
FLoppix: Probably not. To be honest I stuck to the implementation we already use for the `Upload` entity.  In my opinion we can remove this and add a check for the `organizationId`.

andipaetzold: Okay let's leave it as it is for now.

andipaetzold (approved):

PR #785, opened 2021-03-30
feat(AppBundle): support AppBundle entity [EXT-2641]

FLoppix:
## Summary

Introducing the `AppBundle` entity with the following methods:
 - create
 - get
 - getMany
 - delete

## Description

Added the entity to the flat and nested client and exported all relevant types. Also adjusted the `AppDefinition` API to support the `AppBundle` entity for the nested client.

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/app-bundle.ts
+++ b/lib/adapters/REST/endpoints/app-bundle.ts
@@ -8,7 +8,7 @@ import {
   QueryParams,
 } from '../../../common-types'
 import { RestEndpoint } from '../types'
-import { AppBundleProps } from '../../../entities/app-bundle'
+import { AppBundleProps, CreateAppBundleProps } from '../../../entities/app-bundle'
 
 const getBaseUrl = (params: GetAppDefinitionParams) =>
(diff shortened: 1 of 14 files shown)


andipaetzold (changes requested):
There is a toc list in the package.json for the generated type docs. We might want to add AppBundle there.

Thanks for sorting things alphabetically :D

andipaetzold (approved):

lib/adapters/REST/endpoints/app-bundle.ts
```diff
@@ -41,17 +41,21 @@ export const del: RestEndpoint<'AppBundle', 'delete'> = (
 
 export const create = (
```
andipaetzold, line 42: We are missing `RestEndpoint<'AppBundle', 'create'>` here. That's why we didn't get a type error before.

andipaetzold (changes requested):
Looks good. Just a type safety change and then we're happy to 🚢

lib/adapters/REST/endpoints/app-bundle.ts
FLoppix: Good Spot :)

andipaetzold (approved):

PR #797, opened 2021-04-12
fix: correctly set payload & headers when using raw.post & raw.put

andipaetzold:
## Summary

v7.13 introduced a bug where the payload and the axios configuration for `raw.post` and `raw.put` was not set correctly.
The error was a wrong argument order.

This PR fixes the invalid order and adds unit tests to prevent this from happening again.


FLoppix (approved):
Change looks good to me. But I see the `AppBundle` tests are failing. Might be interesting if this is related to the change or we added some validation to the `extensibility-api` or the zipped bundle is not correct.

andipaetzold:
@FLoppix Seems to be an error of the API as dependabot PRs are also failing

PR #798, opened 2021-04-12
chore: provide correct zip for AppBundle integration tests

FLoppix:
## Summary

Provide a correctly zipped `build` output for the `AppBundle`

When adding a new method:


andipaetzold:
Can you make this a `chore:` or `test:` as it's not a fix to the library itself.

andipaetzold (approved):

PR #800, opened 2021-04-13
feat: [PEN-1508] add entry patch support

cribbles:
## Summary

Adds support for [patching an entry](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/entries/entry/patch-an-entry/console/curl).

When adding a new method:

### DIFF
--- a/lib/common-types.ts
+++ b/lib/common-types.ts
@@ -52,6 +52,7 @@ import {
 } from './entities/webhook'
 import { AssetKeyProps, CreateAssetKeyProps } from './entities/asset-key'
 import { AppUploadProps } from './entities/app-upload'
+import { OpPatch } from 'json-patch'
 
 export interface DefaultElements<TPlainObject extends object = object> {
   toPlainObject(): TPlainObject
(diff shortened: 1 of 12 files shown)


lib/plain/plain-client.ts
```diff
@@ -20,6 +20,13 @@ export const createPlainClient = (
           action: 'get',
           params: { url, config },
         }),
+      patch: (url, payload, config) =>
```
cribbles, line 23: not sure why codecov is complaining, this is implicitly tested in `test/unit/entities/entry-test.js`

lib/common-types.ts
```diff
@@ -1050,7 +1060,16 @@ export type MRReturn<
...
+interface MakeRequestOptionsPatch {
+  entityType: keyof MRActions
+  action: 'patch'
+  params?: Record<string, unknown>
+  payload: OpPatch[]
+  headers?: Record<string, unknown>
+  userAgent: string
+}
+
+interface MakeRequestOptionsNonPatch {
```
andipaetzold, line 1072: I think you can merge those. `MakeRequestOptions` is not very accurate anyway. E.g. `payload` is allowed for every action even though get- and delete-actions do not accept payloads.

andipaetzold (approved):

Issue #799, opened 2021-04-13
environment.createUiExtensionWithId not wrapped in extension


AndrewLeedham:
The `environment.createUIExtensionWithId` method does not seem to wrap the payload in `extension` which is causing Contentful to return a `422: Unprocessable Entity` stating as such.

```
ValidationFailed: {
  "status": 422,
  "statusText": "Unprocessable Entity",
  "message": "Validation error",
  "details": {
    "errors": [
      {
        "name": "required",
        "details": "The property \"extension\" is required here",
        "path": [
          "extension"
        ]
      },
...
```

`contentful-management` version: 7.14.1

PR #801, opened 2021-04-13
fix: wrap createUIExtension in extension

AndrewLeedham:
## Summary

Wraps the `createUIExtension` payload in `extension` key.

## Description

Currently running `environment.createUIExtension` fails because it is expecting to be wrapped in `extension` however the typings said the opposite. So this PR automatically wraps the payload in `extension` so the API works as the typings suggest.

## Motivation and Context

fixes: #799 

- [ ] Both unit and integration tests are passing (Couldn't get them to run locally, so will check CI)
- [ ] There are no breaking changes (Potentially breaking, since not TS projects may already wrap in `extension`)
- [x] Changes are reflected in the documentation (Was already how the typings suggested)


shikaan:
Hello @AndrewLeedham 

As far as I can see typings and behaviour currently match. The following snippet, for example, works as intended (see [this](https://github.com/contentful/contentful-management.js/blob/3951915f60cf375c5227fa9f8472bee77a8deaf4/lib/entities/extension.ts#L53-L58))

```typescript
import { createClient } from 'contentful-management';

const client = createClient({
  accessToken: process.env.ACCESS_TOKEN
})

client
  .getSpace(process.env.SPACE_ID)
  .then(s => s.getEnvironment(process.env.ENVIRONMENT_ID))
  .then(e => e.createUiExtension({
    extension: {
      srcdoc: '<html><body>Wat</body></html>',
      name: 'Google' + Date.now(),
      fieldTypes: [{ type: 'Symbol' }],
      sidebar: false,
    }
  })).then(console.log)
```

If there is a specific use case you have in mind, feel free to reopen the issue.

Thanks for contributing!

shikaan:
Hello @AndrewLeedham 

As far as I can see typings and behaviour currently match. The following snippet, for example, works as intended (see [this](https://github.com/contentful/contentful-management.js/blob/3951915f60cf375c5227fa9f8472bee77a8deaf4/lib/entities/extension.ts#L53-L58))

```typescript
import { createClient } from 'contentful-management';

const client = createClient({
  accessToken: process.env.ACCESS_TOKEN
})

client
  .getSpace(process.env.SPACE_ID)
  .then(s => s.getEnvironment(process.env.ENVIRONMENT_ID))
  .then(e => e.createUiExtension({
    extension: {
      srcdoc: '<html><body>Wat</body></html>',
      name: 'Ext',
      fieldTypes: [{ type: 'Symbol' }],
      sidebar: false,
    }
  })).then(console.log)
```

If there is a specific use case you have in mind, feel free to reopen the issue.

Thanks for contributing!

PR #806, opened 2021-04-14
feat(client): add bulkAction API support

luizfonseca:
## Summary

- Add SDK support for the new Contentful [BulkAction API](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/bulk-actions). 

## Description

Added BulkAction API and the following supported functionalities:

- Publish
  - publish multiple items at once by using the `environment.createPublishBulkAction(payload)` function.
- Validate
  - validate multiple items before publishing using the `environment.createValidateBulkAction(payload)` function.
- Unpublish
  - unpublish multiple items at once by using the `environment.createUnpublishBulkAction(payload)` function.

The payload can be seen in the documentation of each function as well as the main Contentful docs.

## Motivation and Context

- As one of the new APIs, this PR adds complete support for all the endpoints in the BulkActions API.

When adding a new method:


lib/create-environment-api.ts
```diff
@@ -159,6 +161,216 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment-id>'))
+     * .then((environment) => environment.createPublishBulkAction(payload))
+     * .then((inProgressBulkAction) => getBulkAction(inProgressBulkAction.sys.id))
+     * .then((bulkAction) => console.log(bulkAction))
+     * .catch(console.error)
+     *
+     * // Using async/await
+     * try {
+     *  const space = await clientgetSpace('<space_id>')
```
andipaetzold, line 233:
```suggestion
     *  const space = await client.getSpace('<space_id>')
```

lib/create-environment-api.ts
```diff
@@ -159,6 +161,216 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment-id>'))
+     * .then((environment) => environment.createValidateBulkAction(payload))
+     * .then((inProgressBulkAction) => getBulkAction(inProgressBulkAction.sys.id))
+     * .then((bulkAction) => console.log(bulkAction))
+     * .catch(console.error)
+     *
+     * // Using async/await
+     * try {
+     *  const space = await clientgetSpace('<space_id>')
```
andipaetzold, line 292:
```suggestion
     *  const space = await client.getSpace('<space_id>')
```

test/integration/bulk-action-integration.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    test('Unpublish BulkAction', async () => {
+      const entry = await testEnvironment.getEntry(TestDefaults.entry.testEntryId)
+
+      const createdBulkAction = await testEnvironment.createUnpublishBulkAction(
+        bulkActionPayload([makeLink('Entry', entry.sys.id)])
+      )
+
+      // Wait for BulkAction completion
+      await delay(1000)
```
andipaetzold, line 110: Is it save to assume the bulk action is done after a second? I think this might become flaky.
Can't we poll the api until the bulk action is done?

andipaetzold (commented):
How complex would it be to type the bulk action payload?

lib/adapters/REST/endpoints/bulk-action.ts
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type { AxiosInstance } from 'contentful-sdk-core'
+import { GetBulkActionParams, GetSpaceEnvironmentParams } from '../../../common-types'
+import { BulkActionProps } from '../../../entities/bulk-action'
+import { RestEndpoint } from '../types'
+import * as raw from './raw'
+
+export const get: RestEndpoint<'BulkAction', 'get'> = (
```
freakyfelt, line 8: Should we define a return type shape for this?

lib/adapters/REST/endpoints/bulk-action.ts
```diff
@@ -0,0 +1,52 @@
...
+    http,
+    `/spaces/${params.spaceId}/environments/${params.environmentId}/bulk_actions/unpublish`,
+    payload
+  )
+}
+
+export const validate: RestEndpoint<'BulkAction', 'validate'> = (
+  http: AxiosInstance,
+  params: GetSpaceEnvironmentParams,
+  payload: any
```
freakyfelt, line 45: Do we really want to have an `any` payload?

lib/common-types.ts
```diff
@@ -566,6 +580,27 @@ export type MRActions = {
       return: AssetKeyProps
     }
   }
+  BulkAction: {
+    get: {
+      params: GetBulkActionParams
+      return: BulkActionProps
```
freakyfelt, line 586: `BulkActionProps` needs to be generic so that the payload shape can be specified

lib/common-types.ts
```diff
@@ -566,6 +580,27 @@ export type MRActions = {
...
+    }
+    unpublish: {
+      params: GetSpaceEnvironmentParams
+      payload: any
+      return: BulkActionProps
+    }
+    validate: {
+      params: GetSpaceEnvironmentParams
+      payload: any
+      return: BulkActionProps
```
freakyfelt, line 601: This should be `BulkActionProps<ValidateBulkActionPayload>`

lib/create-environment-api.ts
```diff
@@ -159,6 +161,216 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     *   accessToken: '<content_management_api_key>'
+     * })
+     *
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment-id>'))
+     * .then((environment) => environment.getBulkAction('<bulk_action_id>'))
+     * .then((bulkAction) => console.log(bulkAction))
+     * ```
+     */
+    getBulkAction(bulkActionId: string) {
```
freakyfelt, line 184: Should probably make this a generic as well so that the payload can be specified (with the default payload being `any` in case the consumer doesn't care)

lib/create-environment-api.ts
```diff
@@ -159,6 +161,216 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * const contentful = require('contentful-management')
+     *
+     * const client = contentful.createClient({
+     *   accessToken: '<content_management_api_key>'
+     * })
+     *
+     * const payload = {
+     *  entities: {
+     *    sys: { type: 'Array' }
+     *    action: 'publish',
```
freakyfelt, line 216: `action` isn't valid on the `publish` payload

lib/entities/bulk-action.ts
```diff
@@ -0,0 +1,72 @@
...
+
+interface BulkActionFailedError {
+  sys: { type: 'Error'; id: 'BulkActionFailed' }
+  details: any
+}
+
+export type BulkActionSysProps = {
+  id: string
+  type: 'BulkAction'
+  status: BulkActionStatus
```
freakyfelt, line 36: I would recommend this accepting the raw string values so that people aren't required to use the enum.

```
enum BulkActionStatus {
  ...
}

const STATUSES = Object.values(BulkActionStatus)
type BulkActionStatuses = typeof STATUSES[number]
...

status: BulkActionStatuses
```

lib/entities/bulk-action.ts
```diff
@@ -0,0 +1,72 @@
...
+  space: Link<'Space'>
+  environment: Link<'Environment'>
+  createdBy: Link<'User'>
+  createdAt: ISO8601Timestamp
+  updatedAt: ISO8601Timestamp
+}
+
+export type BulkActionProps = {
```
freakyfelt, line 44: Make this generic with a `TPayload = any`

lib/entities/bulk-action.ts
```diff
@@ -0,0 +1,72 @@
...
+  payload: any
+  /** error information, if present */
+  error?: BulkActionFailedError
+}
+
+export type BulkActionPayload = {
+  entities: {
+    sys?: { type: string }
+    items: Link<Entity>[] | VersionedLink<Entity>[]
+    [key: string]: any
```
freakyfelt, line 57: what's this for?

lib/entities/bulk-action.ts
```diff
@@ -0,0 +1,72 @@
...
+export type BulkActionProps = {
+  sys: BulkActionSysProps
+  action: BulkActionType
+  /** original payload when BulkAction was created */
+  payload: any
+  /** error information, if present */
+  error?: BulkActionFailedError
+}
+
+export type BulkActionPayload = {
```
freakyfelt, line 53: export separate payload shapes for Validate, Publish, and Unpublish

test/integration/bulk-action-integration.ts
luizfonseca: I was thinking about it, do we have a "poller" in the SDK for these kind of entities? @andipaetzold 

I was planning to add a `waitForCompletion` kind of function (in `entities/bulk-action.ts`) that handles it. Let me know if it's a good idea.

andipaetzold: I don't think we currently have such a thing in the sdk. Adding it would make sense as it's probably a common use case to send the request and just wait for it to complete.

lib/common-types.ts
luizfonseca: I like that 👌🏽

lib/adapters/REST/endpoints/bulk-action.ts
luizfonseca: True! I will define them

lib/entities/bulk-action.ts
luizfonseca: left over from my ts testing, it's removed 👍🏽

lib/adapters/REST/endpoints/bulk-action.ts
luizfonseca: They get the return from the `plain/common-types.ts` file; but I added it here too

lib/create-environment-api.ts
luizfonseca: good 👀

test/integration/bulk-action-integration.ts
luizfonseca: @andipaetzold I added a new method for the bulkAction itself (using the normal client) that waits for the processing to be done.

For the plain API, I didn't find anything that could be used except exposing a helper function

lib/common-types.ts
```diff
@@ -566,6 +586,27 @@ export type MRActions = {
...
     }
   }
+  BulkAction: {
+    get: {
+      params: GetBulkActionParams
+      return: BulkActionProps
+    }
+    publish: {
+      params: GetSpaceEnvironmentParams
+      payload: any
```
freakyfelt, line 596: Is the `any` intentional on these? Or is that tightened up somewhere else?

lib/create-environment-api.ts
```diff
@@ -159,6 +167,218 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * @example ```javascript
+     * const contentful = require('contentful-management')
+     *
+     * const client = contentful.createClient({
+     *   accessToken: '<content_management_api_key>'
+     * })
+     *
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment-id>'))
+     * .then((environment) => environment.getBulkAction('<bulk_action_id>'))
```
freakyfelt, line 186: ⏬  should we stick with `camel_case`?

lib/create-environment-api.ts
```diff
@@ -159,6 +167,218 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * .then((bulkAction) => console.log(bulkAction))
+     * ```
+     */
+    getBulkAction<T extends BulkActionPayload = any>(bulkActionId: string): Promise<BulkAction<T>> {
+      const raw = this.toPlainObject() // Environment object
```
freakyfelt, line 191: wonder if using `const raw: EnvironmentParams = this.toPlainObject()` would help

lib/create-environment-api.ts
```diff
@@ -159,6 +167,218 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     *
+     *  // You can wait for a recently created BulkAction to be processed by using `bulkAction.waitProcessing()`
+     *  const bulkActionCompleted = await bulkActionInProgress.waitProcessing()
+     *  console.log(bulkActionCompleted)
+     * } catch (error) {
+     *  console.log(error)
+     * }
+     * ```
+     */
+    createPublishBulkAction(payload: BulkActionPayload) {
```
freakyfelt, line 249: Does the payload need to be a `BulkActionPublishPayload` here?

lib/create-environment-api.ts
```diff
@@ -159,6 +167,218 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     *
+     *  // You can wait for a recently created BulkAction to be processed by using `bulkAction.waitProcessing()`
+     *  const bulkActionCompleted = await bulkActionInProgress.waitProcessing()
+     *  console.log(bulkActionCompleted)
+     * } catch (error) {
+     *  console.log(error)
+     * }
+     * ```
+     */
+    createValidateBulkAction(payload: BulkActionPayload) {
```
freakyfelt, line 309: Same with this payload

lib/entities/bulk-action.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+const STATUSES = Object.values(BulkActionStatus)
+type BulkActionStatuses = typeof STATUSES[number]
+
+interface BulkActionFailedError {
+  sys: { type: 'Error'; id: 'BulkActionFailed' }
+  details: {
+    errors: any
+    [key: string]: any
```
freakyfelt, line 38: Do we need the `[key: string] :any` here? We know the shape and I don't think we need to allow other parties to define additional fields. Also we know the shape of the errors

lib/entities/bulk-action.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export type BulkActionPayload =
+  | BulkActionPublishPayload
+  | BulkActionUnpublishPayload
+  | BulkActionValidatePayload
+
+export interface BulkActionValidatePayload {
+  action?: 'publish'
+  entities: {
+    sys?: { type: string }
```
freakyfelt, line 50: `type: 'Array'` since we know that part

lib/entities/bulk-action.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+export interface BulkActionUnpublishPayload {
+  entities: {
+    sys?: { type: string }
+    items: Link<Entity>[]
+  }
+}
+
+export interface BulkActionPublishPayload {
+  entities: {
```
freakyfelt, line 62: Might be good to create a `Collection<T>` type where T is `Link<'Entry' | 'Asset'>` or `VersionedLink<'Entry' | 'Asset'>`

lib/methods/bulk-action.ts
```diff
@@ -0,0 +1,102 @@
...
+  BulkActionProps,
+  BulkActionStatus,
+} from '../entities/bulk-action'
+import { PlainClientAPI } from '../plain/common-types'
+
+const DEFAULT_MAX_RETRIES = 30
+const DEFAULT_INITIAL_DELAY_MS = 1000
+const DEFAULT_RETRY_INTERVAL_MS = 2000
+
+export class BulkActionProcessingError extends Error {
```
freakyfelt, line 14: I'm assuming we don't have a base ContentfulError in this client package to extend from?

lib/methods/bulk-action.ts
```diff
@@ -0,0 +1,102 @@
...
+    super(message)
+    this.bulkAction = bulkAction
+    this.name = 'BulkActionProcessingError'
+  }
+}
+
+export type BulkActionProcessingOptions = {
+  /** The amount of times to retry. Defaults to 30 */
+  retryCount?: number
+  /** The interval between retries, in miliseconds (ms). Defaults to 2000 (2s) */
```
freakyfelt, line 27: ⏬  `milliseconds`

lib/methods/bulk-action.ts
```diff
@@ -0,0 +1,102 @@
...
+  retryIntervalMs?: number
+  /**
+   * Initial delay in miliseconds when performing the first check.
```
freakyfelt, line 30: ⏬  `milliseconds`

lib/methods/bulk-action.ts
```diff
@@ -0,0 +1,102 @@
...
+  initialDelayMs?: number
+}
+
+/**
+ * @private
+ * @description Waits for a BulkAction to be completed and to be in one of the final states (failed or succeeded)
+ * @param {Function} fn - GET function that will be called every interval to fetch a bulkAction status
+ */
+export async function pollBulkActionStatus(
+  getBulkActionFunction: Function,
```
freakyfelt, line 43: technically we know the shape of this should be `() => Promise<BulkAction>`

lib/methods/bulk-action.ts
```diff
@@ -0,0 +1,102 @@
...
+  options?: BulkActionProcessingOptions
+): Promise<BulkActionProps> {
+  let retryCount = 0
+  let done = false
+  let action: BulkAction | undefined
+
+  const maxRetries = options?.retryCount || DEFAULT_MAX_RETRIES
+  const retryIntervalMs = options?.retryIntervalMs || DEFAULT_RETRY_INTERVAL_MS
+  const initialDelayMs = options?.initialDelayMs || DEFAULT_INITIAL_DELAY_MS
```
freakyfelt, line 52: I think the preferred approach is to use `??` nowadays instead of a coalesce `||` due to false-y value interpretation

lib/methods/bulk-action.ts
```diff
@@ -0,0 +1,102 @@
...
+      done = true
+      return action
+    }
+
+    await delay(retryIntervalMs)
+    retryCount += 1
+  }
+
+  throw new BulkActionProcessingError(
+    'BulkAction did not complete in the expected duration (60s).',
```
freakyfelt, line 74: Drop the 60s part since it may be configured

test/unit/mocks/entities.js
```diff
@@ -99,6 +100,22 @@ const appUploadMock = {
   }),
 }
 
+const bulkActionMock = {
```
freakyfelt, line 103: I would either export one mock for each bulk action type or use the non-versioned link for this since 2 of the 3 bulk action payloads in use today don't use a versioned link

test/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { Link, VersionedLink } from '../lib/common-types'
+
+export function makeLink(type, id): Link<'Entry' | 'Asset'> {
```
freakyfelt, line 3: Given these functions can be used for more than just Entry and Asset should we instead have this be a generic?

```
export function makeLink<T extends string>(type: T, id): Link<T> {
```

lib/common-types.ts
luizfonseca: 🤦🏽  forgot to add them there, fixed 👍🏽

lib/create-environment-api.ts
luizfonseca: oh, you are correct. I fixed the return but didn't update this one. Fixed 👍🏽

test/utils.ts
luizfonseca: (agreed)

lib/create-environment-api.ts
luizfonseca: (it does)

test/unit/mocks/entities.js
luizfonseca: I added one for publish (separate from the others).

lib/common-types.ts
```diff
@@ -1060,18 +1101,22 @@ export type MRReturn<
   Action extends keyof MRActions[ET]
 > = 'return' extends keyof MRActions[ET][Action] ? Promise<MRActions[ET][Action]['return']> : never
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface MakeRequestPayload {}
```
freakyfelt, line 1105: Can you at least add a JSDoc of what this is supposed to represent?

lib/methods/bulk-action.ts
```diff
@@ -0,0 +1,121 @@
...
+   * Initial delay in milliseconds when performing the first check.
+   * This is used to prevent short running bulkActions of waiting too long for a result.
+   * @default 1000 (1s)
+   * */
+  initialDelayMs?: number
+  /**
+   * Throws an error if the BulkAction does not complete with a successful (succeeded) status.
+   * @default false
+   */
+  throwOnFailedExecution?: boolean
```
freakyfelt, line 45: Are we sure we want that to default to false?

lib/plain/common-types.ts
```diff
@@ -168,6 +176,21 @@ export type PlainClientAPI = {
     ): Promise<EnvironmentAliasProps>
     delete(params: OptionalDefaults<GetSpaceEnvAliasParams>): Promise<any>
   }
+  bulkAction: {
+    get<T extends BulkActionPayload = any>(params: GetBulkActionParams): Promise<BulkActionProps<T>>
+    publish(
+      params: GetSpaceEnvironmentParams,
+      payload: BulkActionPayload
```
freakyfelt, line 183: Do args need to use the specific types?

test/defaults.ts
```diff
@@ -0,0 +1,9 @@
+export const TestDefaults = {
+  spaceId: 'ezs1swce23xe',
+  spaceIdAlternate: '7dh3w86is8ls',
+  SpaceIdV2: 'w6xueg32zr68',
```
freakyfelt, line 4: is the S capitalised for a reason?

lib/entities/bulk-action.ts
freakyfelt: *boops*

lib/entities/bulk-action.ts
freakyfelt: *boops*

freakyfelt (approved):
Minor nits

test/defaults.ts
luizfonseca: yes, a typo 😅

lib/methods/bulk-action.ts
luizfonseca: It can be the opposite, might be actually better ^.

lib/plain/common-types.ts
luizfonseca: Most of them define like such 🤔

lib/methods/bulk-action.ts
luizfonseca: yep, actually this is the 2nd error in the whole package. Mostly, the adapter throws the `id` of the error response as an error. We might define one error base class once another entity gets added (releases for example)

lib/plain/common-types.ts
freakyfelt: I'm saying that shouldn't this be BulkActionPublishPayload

luizfonseca: 🤦🏽  yes! (fixed)

PR #808, opened 2021-04-15
feat: add optional headers param to entry/asset GET

IBratWurst:
## Summary

Add an optional `headers` parameter to `entry.get` and `asset.get` in the plain client

## Motivation and Context

In order for our shared `EntityRepo` to work with the plain client it needs the ability to set those headers

### DIFF
--- a/lib/adapters/REST/endpoints/asset.ts
+++ b/lib/adapters/REST/endpoints/asset.ts
@@ -17,7 +17,7 @@ import { normalizeSelect } from './utils'
 export const get: RestEndpoint<'Asset', 'get'> = (
   http: AxiosInstance,
   params: GetSpaceEnvironmentParams & { assetId: string } & QueryParams,
-  _rawData?: never,
+  rawData?: unknown,
   headers?: Record<string, unknown>
 ) => {
   return raw.get<AssetProps>(
--- a/lib/adapters/REST/endpoints/asset.ts
+++ b/lib/adapters/REST/endpoints/asset.ts
@@ -16,13 +16,16 @@ import { normalizeSelect } from './utils'
 
 export const get: RestEndpoint<'Asset', 'get'> = (
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { assetId: string } & QueryParams
+  params: GetSpaceEnvironmentParams & { assetId: string } & QueryParams,
+  _rawData?: never,
+  headers?: Record<string, unknown>
 ) => {
   return raw.get<AssetProps>(
     http,
     `/spaces/${params.spaceId}/environments/${params.environmentId}/assets/${params.assetId}`,
     {
       params: normalizeSelect(params.query),
+      headers: { ...headers },
     }
   )
 }
--- a/lib/adapters/REST/endpoints/entry.ts
+++ b/lib/adapters/REST/endpoints/entry.ts
@@ -14,13 +14,16 @@ import { normalizeSelect } from './utils'
 
 export const get: RestEndpoint<'Entry', 'get'> = <T extends KeyValueMap = KeyValueMap>(
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { entryId: string } & QueryParams
+  params: GetSpaceEnvironmentParams & { entryId: string } & QueryParams,
+  _rawData?: never,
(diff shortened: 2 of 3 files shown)


lib/adapters/REST/endpoints/asset.ts
```diff
@@ -16,13 +16,16 @@ import { normalizeSelect } from './utils'
 
 export const get: RestEndpoint<'Asset', 'get'> = (
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { assetId: string } & QueryParams
+  params: GetSpaceEnvironmentParams & { assetId: string } & QueryParams,
+  _rawData?: never,
```
IBratWurst, line 20: Is it right to pad for the additional `rawData` parameter?

suevalov: I guess we can just skip rawData parameter and add headers as a third one

IBratWurst: Would it receive the correct parameter from [here](https://github.com/contentful/contentful-management.js/pull/808)?

suevalov: from where?

test/integration/entry-integration.js
```diff
@@ -58,8 +58,8 @@ describe('Entry Api', () => {
 
     test('Gets entries with content type query param', async () => {
       return environment.getEntries({ content_type: 'cat' }).then((response) => {
-        expect(response.total).eql(3)
-        expect(response.items.map((item) => item.sys.contentType.sys.id)).eql(['cat', 'cat', 'cat'])
+        expect(response.total).eql(4)
+        expect(response.items.every((item) => item.sys.contentType.sys.id === 'cat'))
```
djagya, line 62: That's the place I didn't get in my PR. Why the hell does it becomes 4 calls?

IBratWurst: This was from yesterday: a new cat was added in the test space. I was going out of my mind 🤯

lib/adapters/REST/endpoints/asset.ts
IBratWurst: @suevalov sorry 🤦  here https://github.com/contentful/contentful-management.js/blob/master/lib/adapters/REST/rest-adapter.ts#L84

test/integration/entry-integration.js
djagya: Same! That was the reason I couldn't finish my previous PR...

lib/adapters/REST/endpoints/asset.ts
suevalov: hmmm... it's very clunky :/ @andipaetzold is there anything we can with adapters API to avoid so confusing any of skipping one param?

IBratWurst: I come from verifying this. If we call `entry.get(params, headers)` then `headers`, the 2nd argument, ends up as the `payload` argument

suevalov: @IBratWurst ping @andipaetzold on Slack and ask his opinion here

IBratWurst: @suevalov he's on vacation, till April 25 according to bambooh. Good for him but not for our desired feedback 😕

suevalov: okay, let's go with `_rawData?: unknown` for now and merge the PR

suevalov, line 20:
```suggestion
  _rawData?: unknown,
```

DanweDE: Shouldn't it become `rawData?: unknown` (without `_`) if we're not using `never`?

DanweDE (approved):
LGTM with the proposed change.

lib/adapters/REST/endpoints/asset.ts
IBratWurst: I updated it with `rawData?: unknown`, linter might complain at some point for the unused var

IBratWurst:
Merged so it can unblock some dependent pr. I'm happy to go back at it if we want a nicer solution

PR #811, opened 2021-04-19
fix(tests): adjust expected entry count

marcolink:
Due to changed set of data on the test space, we have to adjust our integration test to match count


ruderngespra (approved):
I like it as much as our bot! :)

PR #829, opened 2021-04-30
feat: add entry references to entry and environment methods []

mshaaban0:
## Summary

Add `references` to entry api
Add `getReferences` to environment api

## Todo
- [x] Write `jsdoc` for the new methods
- [x] Write more tests

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/entry.ts
+++ b/lib/adapters/REST/endpoints/entry.ts
@@ -178,3 +178,15 @@ export const createWithId: RestEndpoint<'Entry', 'createWithId'> = <
     }
   )
 }
+
+export const references: RestEndpoint<'Entry', 'references'> = (
+  http: AxiosInstance,
+  params: GetSpaceEnvironmentParams & { entryId: string; maxDepth: number }
+): Promise<CollectionProp<EntryProps>> => {
+  const { spaceId, environmentId, entryId, maxDepth = 2 } = params
+
+  return raw.get<CollectionProp<EntryProps>>(
+    http,
+    `/spaces/${spaceId}/environments/${environmentId}/entries/${entryId}/references?include=${maxDepth}`
+  )
+}
--- a/lib/adapters/REST/endpoints/entry.ts
+++ b/lib/adapters/REST/endpoints/entry.ts
@@ -182,10 +182,10 @@ export const createWithId: RestEndpoint<'Entry', 'createWithId'> = <
 export const references: RestEndpoint<'Entry', 'references'> = (
   http: AxiosInstance,
   params: GetSpaceEnvironmentParams & { entryId: string; maxDepth?: number }
-): Promise<EntryReferenceProps<EntryProps>> => {
+): Promise<EntryReferenceProps> => {
   const { spaceId, environmentId, entryId, maxDepth = 2 } = params
 
-  return raw.get<EntryReferenceProps<EntryProps>>(
+  return raw.get<EntryReferenceProps>(
     http,
     `/spaces/${spaceId}/environments/${environmentId}/entries/${entryId}/references?include=${maxDepth}`
   )
--- a/lib/adapters/REST/endpoints/entry.ts
+++ b/lib/adapters/REST/endpoints/entry.ts
@@ -7,7 +7,7 @@ import {
   KeyValueMap,
   QueryParams,
 } from '../../../common-types'
-import { CreateEntryProps, EntryProps } from '../../../entities/entry'
+import { CreateEntryProps, EntryProps, EntryReferenceProps } from '../../../entities/entry'
 import { RestEndpoint } from '../types'
 import * as raw from './raw'
 import { normalizeSelect } from './utils'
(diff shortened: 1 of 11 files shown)


lib/adapters/REST/endpoints/entry.ts
```diff
@@ -178,3 +178,15 @@ export const createWithId: RestEndpoint<'Entry', 'createWithId'> = <
     }
   )
 }
+
+export const references: RestEndpoint<'Entry', 'references'> = (
+  http: AxiosInstance,
+  params: GetSpaceEnvironmentParams & { entryId: string; maxDepth: number }
```
anho, line 184: `maxDepth` should be declared optional if you provide a default value in line 186.

lib/common-types.ts
```diff
@@ -154,6 +154,9 @@ export interface CollectionProp<TObj> {
   skip: number
   limit: number
   items: TObj[]
+  includes?: {
```
anho, line 157: This feels worth a dedicated type instead of adding it to the default one. And in case of the references endpoint is it really optional?

mshaaban0: What I understand is all `Collection` type apis have optional `includes` field, no?

suevalov: I don't think all of them have `includes` parameter

lib/common-types.ts
```diff
@@ -740,6 +741,10 @@ export type MRActions = {
       payload: CreateEntryProps<any>
       return: EntryProps<any>
     }
+    references: {
+      params: GetSpaceEnvironmentParams & { entryId: string; maxDepth: number }
```
luizfonseca, line 745: you might have to make it a default here too:

```suggestion
      params: GetSpaceEnvironmentParams & { entryId: string; maxDepth?: number }
```

lib/plain/common-types.ts
```diff
@@ -269,6 +269,9 @@ export type PlainClientAPI = {
       >,
       rawData: CreateEntryProps<T>
     ): Promise<EntryProps<T>>
+    references(
+      params: OptionalDefaults<GetSpaceEnvironmentParams & { entryId: string; maxDepth: number }>
```
luizfonseca, line 273: same here
```suggestion
      params: OptionalDefaults<GetSpaceEnvironmentParams & { entryId: string; maxDepth?: number }>
```

test/unit/mocks/entities.js
```diff
@@ -509,6 +509,22 @@ export const scheduledActionCollectionMock = {
   items: [cloneDeep(scheduledActionMock)],
 }
 
+const entryWithReferencesMock = {
+  sys: {
+    type: 'Array',
+  },
+  items: [makeLink('Entry', 'entry-id')],
```
luizfonseca, line 516: might be misleading because the items are not just links btw (they are the entry fields, sys, metadata etc) 🤔 

maybe you could use the entryMock that is in this file?

lib/entities/entry.ts
```diff
@@ -353,6 +365,20 @@ function createEntryApi(makeRequest: MakeRequest): EntryApi {
       const raw = this.toPlainObject() as EntryProps
       return checks.isArchived(raw)
     },
+
+    references: function references(maxDepth: number) {
```
luizfonseca, line 369: you could probably document this as well, e.g.:

```
const testEntry = await environment.getEntry('<entry_id>')
const references = await testEntry.references(2)

// or

environment.getEntry('<entry_id>').then((entry) => entry.references(2) )

```

lib/entities/entry.ts
```diff
@@ -23,6 +24,12 @@ export type EntryProps<T = KeyValueMap> = {
 
 export type CreateEntryProps<TFields = KeyValueMap> = Omit<EntryProps<TFields>, 'sys'>
 
+export interface EntryReferenceProps<TObj> extends CollectionProp<TObj> {
+  includes?: {
+    [key: string]: TObj[]
```
luizfonseca, line 29: Not sure if we should just make it generic... but I think we _could_ try to type this a bit more:
from the references endpoint we know that it's going to be:

```
includes?: {
    Entry?: Collection<EntryProps>,
    Asset?: Collection<AssetProps>,
}
```

lib/create-environment-api.ts
```diff
@@ -770,6 +770,40 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * @example ```javascript
+     * const contentful = require('contentful-management');
+     *
+     * const client = contentful.createClient({
+     *  accessToken: '<contentful_management_api_key>
+     * })
+     *
+     * // Get entry references
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment_id>'))
```
luizfonseca, line 787: I think it'd be useful to have the `getEntry().references` example here too -- since we added to the `entities/entry` file

luizfonseca:
Don't forget to open a PR updating the examples https://github.com/contentful/doc-app/tree/master/content/snippets/cma/entries/entry-references

(also the changelog entry)

test/unit/mocks/entities.js
mshaaban0: updated

lib/entities/entry.ts
mshaaban0: done

test/unit/entities/entry-test.js
```diff
@@ -16,6 +16,7 @@ import {
   isUpdatedTest,
   isDraftTest,
   isArchivedTest,
+  entityReferenceCollectionTest,
```
freakyfelt, line 19: bonus points for sorting

freakyfelt (approved):

luizfonseca (approved):
Nice work! 🎉 

LGTM 🚀

PR #834, opened 2021-05-03
feat: add task API support

andipaetzold:
## Summary

Adds task endpoints 

### Plain client
* `task.getAll`
* `task.get`
* `task.create`
* `task.delete`
* `task.update`

### Nested client

* `entry.getTasks`
* `entry.getTask`
* `entry.createTask`
* `task.delete`
* `task.update`

I am not sure why the BulkAction tests are failing...

When adding a new method:


lib/entities/task.ts
```diff
@@ -0,0 +1,91 @@
...
+      return makeRequest({
+        entityType: 'Task',
+        action: 'delete',
+        params: {
+          ...getParams(raw),
+          version: raw.sys.version,
+        },
+      }).then(() => {
+        // noop
+      })
```
FLoppix, line 80: I think we can remove that

andipaetzold: Just copied that for consistency 😆 

https://github.com/contentful/contentful-management.js/blob/ebc7af28d184db031fad8179b24cab2d4fbc5bb6/lib/create-environment-api.ts#L67-L76

PR #836, opened 2021-05-04
fix(dependencies): batch update

marcolink:


package.json
```diff
@@ -114,17 +115,17 @@
...
+    "nodemon": "^2.0.7",
     "nyc": "^15.1.0",
     "prettier": "^2.0.5",
     "require-all": "^3.0.0",
     "rimraf": "^3.0.2",
-    "semantic-release": "^17.0.4",
+    "semantic-release": "^17.4.2",
     "sinon": "^10.0.0",
-    "typedoc": "0.19.2",
+    "typedoc": "^0.20.36",
```
damienxy, line 125: do you know if there is a reason this was pinned at a specific version?

marcolink: > do you know if there is a reason this was pinned at a specific version?

good catch - will revert it:
https://github.com/contentful/contentful-management.js/blob/master/.dependabot/config.yml#L9

damienxy (approved):

PR #838, opened 2021-05-04
fix: failing big files upload

IBratWurst:
## Summary
The client creation parameters don't handle axios's `maxBodyLength` so it's always -1. This makes big file uploads failing.
As a temporary solution while the `maxBodyLength` is handled in `contentful-sdk-core` we can set the value equal to `maxContentLength` which already is

## Motivation and Context
Costumer reports failing upload with big files

## Notes
Once [this pr](https://github.com/contentful/contentful-sdk-core/pull/152) for `contentful-sdk-core` is merged we will be able to use a separate `maxBodyLength`


anho (approved):

IBratWurst:
Used `contentful-sdk-core` 6.8.0 which handles `maxBodyLength`

PR #843, opened 2021-05-06
refactor(entry-references): use named args instead of positional

mshaaban0:
## Summary

Refactor `getEntryReferences` and `entry.references` to use named args instead of positional

## Checklist 

### DIFF
--- a/lib/entities/entry.ts
+++ b/lib/entities/entry.ts
@@ -32,6 +32,10 @@ export interface EntryReferenceProps extends CollectionProp<EntryProps> {
   }
 }
 
+export type EntryReferenceOptionsProps = {
+  maxDepth?: number
+}
+
 type EntryApi = {
   /**
    * Sends an update to the server with any changes made to the object's properties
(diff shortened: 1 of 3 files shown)


test/integration/entry-references-integration.ts
```diff
@@ -20,9 +20,8 @@ describe('Entry References', async function () {
   describe.only('Environment Scoped', () => {
```
freakyfelt, line 20: `only`

mshaaban0: 🤦🏼

lib/entities/entry.ts
```diff
@@ -236,7 +236,7 @@ type EntryApi = {
   /**
    * Recursively collects references of an entry and their descendants
    */
-  references(maxDepth: number): Promise<EntryReferenceProps>
+  references({ maxDepth }: { maxDepth: number }): Promise<EntryReferenceProps>
```
freakyfelt, line 239: Probably okay for now, but if we add anymore query options we should export a `EntryReferenceOptions` interface

freakyfelt (approved):

lib/entities/entry.ts
freakyfelt: Also does this arg need to be optional?

luizfonseca: In this case I'd add a new type ^

luizfonseca (approved):
Nice! Thanks for addressing it 🙏🏽 

Just one minor comment

freakyfelt (approved):
Thank you for doing that <3

PR #845, opened 2021-05-06
feat(releases): add releases API support to the sdk

luizfonseca:
## Summary

This PR adds support for the[ Releases API](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/releases/releases) to the SDK:

- Get Release `environment.getRelease(<release-id>)` and plainClient: `plainClient.release.get(<release-id>)`
- Create Release `environment.createRelease(<payload>)` and plainClient: `plainClient.release.create(<payload>)`
- Delete Release `environment.deleteRelease(<release-id>)` and plainClient: `plainClient.release.delete(<release-id>)`

### Next Pull Request will address these: 

- Publish Release `environment.publishRelease({ releaseId: <release-id>, version: <version>)`
- Unpublish Release `environment.unpublishRelease({ releaseId: <release-id>, version: <version>)`
- Validate Release `environment.validateRelease({ releaseId: <release-id> })`

When adding a new method:


lib/adapters/REST/endpoints/release-action.ts
```diff
@@ -0,0 +1,25 @@
...
+) => {
+  return raw.get(
+    http,
+    `/spaces/${params.spaceId}/environments/${params.environmentId}/releases/${params.releaseId}/actions/${params.actionId}`
+  )
+}
+
+export const query: RestEndpoint<'ReleaseAction', 'query'> = (
+  http: AxiosInstance,
+  params: GetReleaseParams
```
freakyfelt, line 19: Any query params that we support for release actions here?

lib/common-types.ts
```diff
@@ -83,6 +85,14 @@ export interface VersionedLink<T extends string> {
...
   }
 }
+
+export interface EntityPayload<T> {
+  entities: {
+    sys: { type: 'Array' }
+    items: T[]
+  }
+}
+
```
freakyfelt, line 95: Ehhh.... can we instead have a `BaseCollection` interface that just has the `{ sys, items }` bit? The `entities` tacked on feels very specific to Release

freakyfelt: Sometimes it's better to write everything twice (WET), especially if the shapes may diverge in the future.

lib/entities/release.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { freezeSys, toPlainObject } from 'contentful-sdk-core'
+import copy from 'fast-copy'
+import {
+  BaseCollection,
```
freakyfelt, line 5: Bonus points for sorting (I'm going to have to start paying out these bonus points someday)

lib/entities/release.ts
```diff
@@ -0,0 +1,107 @@
...
+export interface ReleaseQueryOptions {
+  /** Find releases filtered by the Entity type (Asset, Entry) */
+  'entities.sys.linkType'?: string
+  /** Find releases by using comma-separated list of Entities Link Ids */
+  'entities.sys.id'?: string
+  /** Find releases by using comma-separated list of Ids */
+  'sys.id'?: string
+  /** If present, will return results based of a pagination cursor */
+  pageNext?: string
+  /** Limit how many records are returned in the result */
```
freakyfelt, line 29: Can you add that the `@default` value for `limit`?

lib/entities/release.ts
```diff
@@ -0,0 +1,107 @@
...
+import { wrapCollection } from '../common-utils'
+import enhanceWithMethods from '../enhance-with-methods'
+
+/** Entity types supported by the Release API */
+type Entity = 'Entry' | 'Asset'
+
+export interface ReleaseQueryOptions {
+  /** Find releases filtered by the Entity type (Asset, Entry) */
+  'entities.sys.linkType'?: string
+  /** Find releases by using comma-separated list of Entities Link Ids */
```
freakyfelt, line 23:
```suggestion
  /** Find releases containing the specified, comma-separated entities. Requires `entities.sys.linkType` */
```

lib/entities/release.ts
```diff
@@ -0,0 +1,107 @@
...
+  entities: BaseCollection<Link<Entity>>
+}
+
+export interface ReleasePayload extends MakeRequestPayload {
+  title: string
+  entities: BaseCollection<Link<Entity>>
+}
+
+export interface ReleaseValidatePayload {
+  action: 'publish' | 'unpublish'
```
freakyfelt, line 58: Do we support `unpublish` at the moment?

lib/methods/action.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * @private
+ * @description Waits for an Action to be completed and to be in one of the final states (failed or succeeded)
+ * @param {Function} fn - GET function that will be called every interval to fetch an Action status
+ * @throws {ActionFailedError} throws an error if `throwOnFailedExecution = true` with the Action that failed.
+ * @throws {ActionProcessingError} throws an error with a Action when processing takes too long.
+ */
+export async function pollActionStatus<T extends Action = any>(
```
freakyfelt, line 56: ⏬  can we call this `pollAsyncActionStatus`?

lib/entities/release.ts
luizfonseca: (going to leave just `publish` for now)

lib/entities/release.ts
luizfonseca: I accept bitcoins :trollface:

lib/methods/action.ts
```diff
@@ -0,0 +1,94 @@
...
+  public action?: Action
+
+  constructor(message: string, action?: Action) {
+    super(message)
+    this.action = action
+    this.name = this.constructor.name
+  }
+}
+
+export class ActionFailedError extends ActionProcessingError {}
```
luizfonseca, line 25: TLDR here is that since Release Actions and BulkActions share a bit of the logic around failed/succeded actions I refactored it to accept anything that has a. `sys.status` property we can check.

lib/entities/release.ts
freakyfelt: no sorting means no bitcoins (or other coins) for you!

lib/entities/release.ts
```diff
@@ -0,0 +1,110 @@
...
+  entities: BaseCollection<Link<Entity>>
+}
+
+export interface ReleasePayload extends MakeRequestPayload {
+  title: string
+  entities: BaseCollection<Link<Entity>>
+}
+
+export interface ReleaseValidatePayload {
+  action: 'publish'
```
freakyfelt, line 61: Is `publish` required? Or is an empty payload acceptable?

lib/entities/release.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export interface ReleaseApiMethods {
+  /** Deletes a Release and all ReleaseActions linked to it (non-reversible) */
+  delete(): Promise<null>
```
freakyfelt, line 66: Is this `Promise<null>` or `Promise<void>` (e.g. does it explicitly `return null` at the end?)

lib/methods/bulk-action.ts
```diff
@@ -107,9 +14,9 @@ type PlainOptions = {
  * Used by the Plain client */
 export async function waitForBulkActionProcessing<T extends BulkActionPayload = any>(
   { plainClient, spaceId, environmentId, bulkActionId }: PlainOptions,
-  options?: BulkActionProcessingOptions
+  options?: ActionProcessingOptions
```
freakyfelt, line 17: ⏬  rename to `AsyncActionProcessingOptions` for consistency

test/integration/bulk-action-integration.ts
```diff
@@ -75,7 +75,7 @@ describe('BulkActions Api', async function () {
     })
 
     test('Publish BulkAction with wrong payload', async () => {
-      const entry = await testEnvironment.getEntry(TestDefaults.entry.testEntryId)
+      const entry = await testEnvironment.getEntry(TestDefaults.entry.testEntryReferenceId)
```
freakyfelt, line 78: This feels like we're overloading the testEntryReferenceId for multiple test suites. Should we instead create a separate testEntry for this suite?

lib/methods/action.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      return action
+    }
+
+    await sleep(retryIntervalMs)
+    retryCount += 1
+  }
+
+  throw new ActionProcessingError(
+    "Action didn't finish processing within the expected timeframe.",
```
freakyfelt, line 91: We might be able to get clever and do `${sys.type} did not finish processing within the expected timeframe.`

lib/entities/release.ts
luizfonseca: empty is acceptable (I made it optional in the request)

lib/entities/release.ts
luizfonseca: it actually returns `''` -- so maybe `void` is better in this case?

test/integration/bulk-action-integration.ts
luizfonseca: good point ^ 👍🏽 -- will create a new one

lib/methods/action.ts
luizfonseca: (agreed)

lib/common-types.ts
```diff
@@ -838,6 +851,30 @@ export type MRActions = {
...
+      return: ReleaseProps
+    }
+    update: {
+      params: GetReleaseParams & { version: number }
+      payload: ReleasePayload
+      return: ReleaseProps
+    }
+    delete: {
+      params: GetReleaseParams
+      return: null
```
freakyfelt, line 875: Should this be null, void, or undefined?

lib/entities/release.ts
freakyfelt: Might be worthwhile to have it explicitly be `?:` here so it's clear that the param itself is optional, not just the entire payload

lib/entities/release.ts
```diff
@@ -0,0 +1,110 @@
...
+      environmentId: release.sys.environment.sys.id,
+      releaseId: release.sys.id,
+      version: release.sys.version,
+    }
+  }
+
+  return {
+    async delete() {
+      const params = getParams(this)
+      return makeRequest({
```
freakyfelt, line 84: given we don't need to return data we _could_ just `await` this

test/integration/bulk-action-integration.ts
freakyfelt: to be clear: in this PR? or in another PR?

lib/plain/common-types.ts
```diff
@@ -383,6 +385,21 @@ export type PlainClientAPI = {
...
+    ): Promise<CollectionProp<ReleaseProps>>
+    create(
+      params: OptionalDefaults<GetSpaceEnvironmentParams>,
+      data: ReleasePayload
+    ): Promise<ReleaseProps>
+    update(
+      params: OptionalDefaults<GetReleaseParams & { version: number }>,
+      data: ReleasePayload
+    ): Promise<ReleaseProps>
+    delete(params: OptionalDefaults<GetReleaseParams>): Promise<null>
```
freakyfelt, line 401: Would strongly recommend having the return being a `Promise<void>`

freakyfelt (approved):
no blockers, but some quality of life suggestions

lib/plain/common-types.ts
luizfonseca: Fixed 👍🏽

lib/common-types.ts
luizfonseca: (moved it to void)

test/integration/bulk-action-integration.ts
luizfonseca: (did it on this one)

lib/entities/release.ts
luizfonseca: :done:

PR #857, opened 2021-05-10
feat: add editor interface defaults

loweisz:
## Summary

Adds the collection of the default lists of widgets for the `EditorInterface`, having one single source of truth and be used in multiple places. 

Sidebar and Editors are exported default constants. For controls there is an exported function that will return the default control of a field. 

When adding a new method:


test/unit/constants/controls-defaults-test.js
```diff
@@ -0,0 +1,93 @@
...
+    const field = { type: 'Link', linkType: 'Asset' }
+    expect(getDefaultControlOfField(field, 'displayfieldid')).equals('assetLinkEditor')
+  })
+
+  it('if field is a list of Assets', () => {
+    const field = { type: 'Array', items: { type: 'Link', linkType: 'Asset' } }
+    expect(getDefaultControlOfField(field, 'displayfieldid')).equals('assetLinksEditor')
+  })
+
+  it('if field is a list of Entries', () => {
```
kdamball, line 75: Might need to double check if this is default behaviour: what happens if an entry has bulk editing enabled? I think we usually return 
```js
    {
      "fieldId": "test",
      "widgetId": "entryLinksEditor",
      "widgetNamespace": "builtin",
      "settings": {
        "bulkEditing": true,
        "showCreateEntityAction": true,
        "showLinkEntityAction": true
      }
    }
```

Lets at least keep these settings somewhere here

lib/constants/editor-interface-defaults/controls-defaults.ts
```diff
@@ -0,0 +1,194 @@
...
+      widgetId: 'dropdown',
+      fieldId: field.id,
+      widgetNameSpace: 'builtin',
+    }
+  }
+
+  const isTextField = fieldType === 'Text'
+  const isDisplayField = field.id === displayFieldId
+
+  if (isTextField && isDisplayField) {
```
kdamball, line 189: nit: does it make sense to combine these two together? Looks like we're using both only once so wont hurt to say `isDisplayTextField`

lib/constants/editor-interface-defaults/controls-defaults.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/*
+ * Gets the default widget ID for a field:
+ * - If a field allows predefined values then `dropdown` widget is used
+ *   in the presence of the `in` validation.
+ * - If a Text field is a title then the `singleLine` widget is used.
+ * - Otherwise a simple type-to-editor mapping is used.
+ */
+export default function getDefaultControlOfField(field: ContentFields, displayFieldId: string) {
```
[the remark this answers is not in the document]
loweisz: Good point I mainly wanted to keep as close to the functionality as it was implemented in the web app. But you are probably right for this use case I don't see a reason to keep it. 👍

kdamball (approved):
Good work 🎉

PR #859, opened 2021-05-11
fix: type of BulkAction.waitProcessing

andipaetzold:
## Summary

`waitProcessing` is async and the type and therefore returns a Promise.

Also, the lodash `merge` method does not overwrite `sys.status` properties which made the updated tests fail.


luizfonseca (approved):
Oh, thanks for fixing it 👍🏽 !

PR #863, opened 2021-05-18
chore: add timezone to schedule action props

ChidinmaOrajiaku:
## Summary

Add" timezone" to schedule action props i.e : 

scheduledFor: {
  datetime: datetime,
  timezone: timezone,
}

When adding a new method:

### DIFF
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -49,6 +49,12 @@ export type ScheduledActionProps = {
   environment?: { sys: MetaLinkProps }
   scheduledFor: {
     datetime: ISO8601Timestamp
+    /**
+     * A valid IANA timezone Olson identifier
+     *
+     * @see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+     * @example 'Asia/Kolkata'
+     */
     timezone?: string
   }
 }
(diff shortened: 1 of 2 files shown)


test/unit/mocks/entities.js
```diff
@@ -548,6 +548,7 @@ export const scheduledActionMock = {
   environment: { sys: cloneDeep(linkMock) },
   scheduledFor: {
     datetime: 'scheduledFor',
+    timezone: 'timezone',
```
freakyfelt, line 551: Should we make these somewhat plausible values? e.g.

```suggestion
    datetime: '2006-01-02T15:04:05-0700',
    timezone: 'Americas/Denver',
```

lib/entities/scheduled-action.ts
```diff
@@ -49,6 +49,7 @@ export type ScheduledActionProps = {
   environment?: { sys: MetaLinkProps }
   scheduledFor: {
     datetime: ISO8601Timestamp
+    timezone?: string
```
freakyfelt, line 52: Any chance we can add some /** JSDoc **/ here describing what we are expecting? See [contentful/types](https://github.com/contentful/content-stack/blob/master/packages/types/lib/types.ts#L6-L17) for inspiration

ChidinmaOrajiaku: Thank you!

freakyfelt (approved):

PR #864, opened 2021-05-20
fix: require env id as an argument to scheduled actions delete function

Spring3:
## Summary
Deleting a scheduled action is impossible as it throws an error "Missing environment id". This PR fixes it

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/scheduled-action.ts
+++ b/lib/adapters/REST/endpoints/scheduled-action.ts
@@ -1,5 +1,10 @@
 import type { AxiosInstance } from 'contentful-sdk-core'
-import { CollectionProp, GetSpaceParams, QueryParams } from '../../../common-types'
+import {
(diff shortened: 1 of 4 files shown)


Spring3:
@luizfonseca I have this pr here that I closed https://github.com/contentful/contentful-management.js/pull/861

But there I tried to migrate it there. Do you think it's worth re-opening and rebasing it?

Spring3:
@luizfonseca will do that. Thanks for suggesting

freakyfelt:
> I do feel that this API is misplaced. If it depends on the environment (it's required) it should live in the `create-environment-api.ts` - But probably... later :D

Nope, it is explicitly space scoped and currently all operations require an environment (though future ones may not)

suevalov (approved):

luizfonseca (approved):

PR #867, opened 2021-05-26
fix(deps): downgrade type-fest

marcolink:
We version conflicts with `type-fest` when used in other projects. With this changes, we downgrade the version of `type-fest` to unblock users. 

We will investigate further how and when we can update again.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,5 +15,8 @@ updates:
       - dependency-name: typedoc
         versions:
           - '>= 0'
+      - dependency-name: type-fest
+        versions:
+          - '>= 1'
     registries:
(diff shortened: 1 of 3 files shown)


luizfonseca (approved):
LGTM.


Would this fix the issue? I tried it locally as well, no changes though

johanneswuerbach:
If this is locked, it should probably also be locked in https://github.com/contentful/contentful-management.js/blob/master/.github/dependabot.yml#L14

marcolink:
> If this is locked, it should probably also be locked in https://github.com/contentful/contentful-management.js/blob/master/.github/dependabot.yml#L14

Good point, added it to the ignore list.

PR #872, opened 2021-06-14
chore: update scheduled action sys props []

ChidinmaOrajiaku:
## Summary

Add `version`, `updatedAt` and `updatedBy` to `ScheduledSysProps`

When adding a new method:

### DIFF
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -7,6 +7,7 @@ import {
   MetaLinkProps,
   Link,
   MakeRequest,
+  SysLink,
 } from '../common-types'
 import { wrapCollection } from '../common-utils'
 import enhanceWithMethods from '../enhance-with-methods'
@@ -33,17 +34,17 @@ type SchedulableActionType = 'publish' | 'unpublish'
 export type ScheduledActionSysProps = {
   id: string
   type: 'ScheduledAction'
-  version?: number
-  space: { sys: MetaLinkProps }
+  version: number
+  space: SysLink
   status: ScheduledActionStatus
   createdAt: ISO8601Timestamp
-  createdBy: { sys: MetaLinkProps }
+  createdBy: Link<'User'>
   /** an ISO8601 date string representing when an action was moved to canceled */
   canceledAt?: ISO8601Timestamp
-  canceledBy?: { sys: MetaLinkProps }
+  canceledBy?: Link<'User'>
   /** an ISO8601 date string representing when an action was updated */
   updatedAt: ISO8601Timestamp
-  updatedBy: { sys: MetaLinkProps }
+  updatedBy: Link<'User'>
 }
 
 export type ScheduledActionProps = {
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -38,10 +38,10 @@ export type ScheduledActionSysProps = {
   space: SysLink
   status: ScheduledActionStatus
   createdAt: ISO8601Timestamp
-  createdBy: Link<'User'>
+  createdBy: SysLink
   /** an ISO8601 date string representing when an action was moved to canceled */
   canceledAt?: ISO8601Timestamp
-  canceledBy?: Link<'User'>
+  canceledBy?: SysLink
   /** an ISO8601 date string representing when an action was updated */
   updatedAt: ISO8601Timestamp
   updatedBy: Link<'User'>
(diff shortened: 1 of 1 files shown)


lib/entities/scheduled-action.ts
```diff
@@ -33,13 +33,17 @@ type SchedulableActionType = 'publish' | 'unpublish'
...
+  version?: number
   space: { sys: MetaLinkProps }
   status: ScheduledActionStatus
   createdAt: ISO8601Timestamp
   createdBy: { sys: MetaLinkProps }
   /** an ISO8601 date string representing when an action was moved to canceled */
   canceledAt?: ISO8601Timestamp
   canceledBy?: { sys: MetaLinkProps }
+  /** an ISO8601 date string representing when an action was updated */
+  updatedAt: ISO8601Timestamp
```
Spring3, line 45: Could you please use [one of these](https://github.com/contentful/contentful-management.js/blob/master/lib/common-types.ts#L139) for consistency?

Spring3 (changes requested):

lib/entities/scheduled-action.ts
```diff
@@ -33,13 +33,17 @@ type SchedulableActionType = 'publish' | 'unpublish'
...
+  updatedBy: { sys: MetaLinkProps }
```
luizfonseca, line 46: I added a `Link` type here, maybe we could use  `updatedBy: Link<'User'>` instead to make it more clear

lib/entities/scheduled-action.ts
```diff
@@ -33,13 +33,17 @@ type SchedulableActionType = 'publish' | 'unpublish'
 export type ScheduledActionSysProps = {
   id: string
   type: 'ScheduledAction'
+  version?: number
   space: { sys: MetaLinkProps }
```
Spring3, line 37: would be great to replace `{ sys: MetaLinkProps }` with just `SysLink`

lib/entities/scheduled-action.ts
```diff
@@ -33,13 +33,17 @@ type SchedulableActionType = 'publish' | 'unpublish'
 export type ScheduledActionSysProps = {
   id: string
   type: 'ScheduledAction'
+  version?: number
```
Spring3, line 36: I am concerned about the optional version here. Isn't version a consistent property that starts with 1 and gets increased with updates?

luizfonseca: I think for the SDK it can be mandatory (without `?`) -- for some other projects it might break since the version has not been introduced fully yet.

Spring3 (approved):
Looks good. Thank you 💯

lib/entities/scheduled-action.ts
Spring3: It will require typing check, true. But if you leave it optional when in fact it's required, then you'd have to cast things to string or check against undefined each time, which can be even more annoying than updating mocks/fixtures by adding version to it

Spring3: I mean I understand the concern, but should we then mark each new property as optional because it didn't use to be there before? Might not be a great approach

Spring3: Changing
```
createdBy: { sys: MetaSysLink }
```
to
```
createdBy: Link<'User'>
```

might be as breaking ,because what if I set my mock there to be "Person" instead of "User"? :)

luizfonseca: https://github.com/contentful/contentful-management.js/pull/872#discussion_r651896135

I think you misunderstood, for the SDK I agree with it being mandatory.

luizfonseca: > Changing
> 
> ```
> createdBy: { sys: MetaSysLink }
> ```
> 
> to
> 
> ```
> createdBy: Link<'User'>
> ```
> 
> might be as breaking ,because what if I set my mock there to have linkType of "Person" instead of "User" and now I updated the sdk and it broke? :)

Why would you set it to `Person`? 🤔 

I don't mind either tbh, `Link<'User'>` is actually what it is but `SysLink` is also a good abstraction.

ChidinmaOrajiaku: 😀 Okay. I'll change it to `SysLink`

luizfonseca: 🤦🏽  -- forgot that we are _adding_  fields here  as well -- for those I'd recommend using the correct `Link<'Type'>` - existing ones can use `SysLink`

luizfonseca (approved):
🚀

PR #873, opened 2021-06-15
feat: introduce team.getManyForSpace [HOMER-75]

Chaoste:
## Summary

After adding the [space scoped teams endpoint to gatekeeper](https://github.com/contentful/gatekeeper/pull/5214), we introduce this new entity to the CMA client.

## Description

I added both the entity itself and the exported functions in the same manner as for the `SpaceMember` & `Team` entities.

## Motivation and Context

The overall goal is to assign tasks to teams. Since we need to know which teams are assignable for a task in an entry within a space, we need to get a space-scoped version of the team entity.

Following this PR, we will integrate the function in the UI extensions SDK adapter and then use it in the newly built tasks app.

Blocked PRs:
- https://github.com/contentful/ui-extensions-sdk/pull/789
- https://github.com/contentful/user_interface/pull/8376

- [x] Changes are reflected in the documentation -> https://github.com/contentful/doc-app/pull/2007
- [x] Gatekeeper PR is merged

When adding a new method:

- [x] The new method is added to the documentation -> https://github.com/contentful/doc-app/pull/2007


lib/adapters/REST/endpoints/space-team.ts
```diff
@@ -0,0 +1,18 @@
+import type { AxiosInstance } from 'contentful-sdk-core'
+import { CollectionProp, GetSpaceTeamParams, PaginationQueryParams } from '../../../common-types'
+import { SpaceTeamProps } from '../../../entities/space-team'
+import { RestEndpoint } from '../types'
+import * as raw from './raw'
+
+export const get: RestEndpoint<'SpaceTeam', 'get'> = (
+  http: AxiosInstance,
+  params: GetSpaceTeamParams & { teamId: string }
+) => raw.get<SpaceTeamProps>(http, `/spaces/${params.spaceId}/teams/${params.teamId}`)
```
giotiskl, line 10: @matthew-contentful do we actually have an endpoint for this? 🤔 I thought we only introduced `/spaces/${params.spaceId}/teams`

giotiskl:
PR looks good overall, my 2 concerns are:
* I am hesitant to expose the name `SpaceTeam` because it can be misleading and looks like this is a separate entity, while `Teams` are still org-level. We should consider exposing instead something like `cma.team.getForSpace` which makes it more look like this is just a filtered set of teams
* I am not sure if we actually have a `show` endpoint that reads only a single team based on team ID @matthew-contentful can shed light on this

lib/adapters/REST/endpoints/space-team.ts
matthew-contentful: We do, rails made it easier to leave that in than to remove it. It returns a single team object, not an array or any other payload metadata.

matthew-contentful: And oops, I need to add it to the customer facing docs :)

giotiskl: Ok this is good news, then we can keep it

Chaoste: As discussed with Matthew, we will hide this endpoint completely for now. Gonna remove it in cma and docs

Chaoste:
> * I am hesitant to expose the name `SpaceTeam` because it can be misleading and looks like this is a separate entity, while `Teams` are still org-level. We should consider exposing instead something like `cma.team.getForSpace` which makes it more look like this is just a filtered set of teams

@giotiskl I agree with you that we should expose something like a new entity for this. I updated the code to use `team.getManyForSpace` instead (similar notation as in `user.getManyForSpace`)

lib/create-contentful-api.ts
```diff
@@ -324,6 +325,39 @@ export default function createClientApi(makeRequest: MakeRequest) {
       }).then((data) => wrapUsageCollection(makeRequest, data))
     },
 
+    /**
+     * Get space-scoped teams
```
giotiskl, line 329:
```suggestion
     * Get a filtered set of teams that have access to a space
```

giotiskl (approved):
LGTM except for a small comment

PR #874, opened 2021-06-18
fix: use correct name for new method [HOMER-75]

Chaoste:
## Summary

It's `SpaceApi.getTeams`, not `SpaceApi.getSpaceTeams`.

## Description

[Related open App-SDK PR](https://github.com/contentful/ui-extensions-sdk/pull/797/files)
[Previous PR](https://github.com/contentful/contentful-management.js/pull/873)

When adding a new method:

### DIFF
--- a/lib/create-space-api.ts
+++ b/lib/create-space-api.ts
@@ -557,15 +557,14 @@ export default function createSpaceApi(makeRequest: MakeRequest) {
      * .catch(console.error)
      * ```
      */
-    getTeams() {
+    getTeams(query: QueryOptions = { limit: 100 }) {
       const raw = this.toPlainObject() as SpaceProps
       return makeRequest({
         entityType: 'Team',
         action: 'getManyForSpace',
         params: {
           spaceId: raw.sys.id,
-          // The largest organization includes 29 teams, so the limit is sufficient and no pagination is required
-          query: createRequestConfig({ query: { limit: 100 } }).params,
+          query: createRequestConfig({ query }).params,
         },
       }).then((data) => wrapTeamCollection(makeRequest, data))
     },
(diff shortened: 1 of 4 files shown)


lib/create-space-api.ts
```diff
@@ -552,19 +552,20 @@ export default function createSpaceApi(makeRequest: MakeRequest) {
...
      *
      * client.getSpace('<space_id>')
-     * .then((space) => space.getSpaceTeams({'limit': 100}))
-     * .then((spaceTeamCollection) => console.log(spaceTeamCollection))
+     * .then((space) => space.getTeams())
+     * .then((teamsCollection) => console.log(teamsCollection))
      * .catch(console.error)
      * ```
      */
-    getSpaceTeams(query: QueryOptions = {}) {
```
FLoppix, line 560: why did we remove the query params here?

Chaoste: I didn't define them in the app-SDK at all. Since I have this difference between the app-sdk and the actual implementation, I wanted to remove this malicious implementation ASAP. As [Kado mentioned](https://contentful.slack.com/archives/C0201KJQQD9/p1623930046041100), we should look out for those cases where we define the method without having the implementations ready.
I saw that `space.getUsers` also doesn't allow parameters, so I thought it's fine to stick to this solution for now until we enhance it.

FLoppix: But from a raw cma `sdk` perspective we should add support for query options. I can imagine using the `sdk` in a different code base where I want pagination support. We should not tie the implementation of the cma sdk to the `app-sdk`.

Chaoste: As discussed, I'm readding the query params :)

FLoppix (approved):

PR #880, opened 2021-07-07
Use integration-test-util repo and migrate integration tests to new org

damienxy:
## Summary

- use https://github.com/contentful/integration-test-utils for integration tests
- migrate tests to new org
- fix some tests

When adding a new method:


marcolink (approved):
🚀

PR #882, opened 2021-07-08
Feat: update scheduled actions [SPA-284]

Spring3:
## Summary

Adds support for 
```js
plainClient.scheduledActions.update();
spaceClient.updateScheduledAction();
scheduledAction.update();
```

When adding a new method:

### DIFF
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -101,6 +101,42 @@ export default function getInstanceMethods(makeRequest: MakeRequest): ScheduledA
   }
 
   return {
+    /**
+     * Delete current scheduled action.
+     *
+     * @example ```javascript
+     *  const contentful = require('contentful-management');
+     *
+     *  const client = contentful.createClient({
+     *    accessToken: '<content_management_api_key>'
+     *  })
+     *
+     *  client.getSpace('<space_id>')
+     *    .then((space) => {
+     *      return space.createScheduledAction({
+     *        entity: {
+     *          sys: {
+     *            type: 'Link',
+     *            linkType: 'Entry',
+     *            id: '<entry_id>'
+     *          }
+     *        },
+     *        environment: {
+     *          type: 'Link',
+     *          linkType: 'Environment',
+     *          id: '<environment_id>'
+     *        },
+     *        action: 'publish',
+     *        scheduledFor: {
+     *          dateTime: <ISO_date_string>,
+     *          timezone: 'Europe/Berlin'
+     *        }
+     *      })
+     *    .then((scheduledAction) => scheduledAction.delete())
+     *    .then((deletedScheduledAction) => console.log(deletedScheduledAction))
+     *    .catch(console.error);
+     * ```
+     */
     async delete(): Promise<ScheduledAction> {
       const params = getParams(this)
 
@@ -110,6 +146,46 @@ export default function getInstanceMethods(makeRequest: MakeRequest): ScheduledA
         params,
       }).then((data) => wrapScheduledAction(makeRequest, data))
     },
+    /**
+     * Update current scheduled actions. All changes made to current instance will be saved in the database,
+     * provided the request succeeds.
+     *
+     * @example ```javascript
+     *  const contentful = require('contentful-management');
+     *
+     *  const client = contentful.createClient({
+     *    accessToken: '<content_management_api_key>'
+     *  })
+     *
+     *  client.getSpace('<space_id>')
+     *    .then((space) => {
+     *      return space.createScheduledAction({
+     *        entity: {
+     *          sys: {
+     *            type: 'Link',
+     *            linkType: 'Entry',
+     *            id: '<entry_id>'
+     *          }
+     *        },
+     *        environment: {
+     *          type: 'Link',
+     *          linkType: 'Environment',
+     *          id: '<environment_id>'
+     *        },
+     *        action: 'publish',
+     *        scheduledFor: {
+     *          dateTime: <ISO_date_string>,
+     *          timezone: 'Europe/Berlin'
+     *        }
+     *      })
+     *    .then((scheduledAction) => {
+     *      scheduledAction.scheduledFor.timezone = 'Europe/Paris';
+     *      return scheduledAction.update();
+     *    })
+     *    .then((scheduledAction) => console.log(scheduledAction))
+     *    .catch(console.error);
+     * ```
+     */
     async update(): Promise<ScheduledAction> {
       const params = getParams(this)
 
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -79,7 +79,7 @@ export interface ScheduledActionQueryOptions extends BasicCursorPaginationOption
   [key: string]: any
 }
 
-type ScheduledActionApi = {
+export type ScheduledActionApi = {
   delete(): Promise<ScheduledAction>
   update(): Promise<ScheduledAction>
 }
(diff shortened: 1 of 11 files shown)


test/integration/scheduled-action-integration.ts
```diff
@@ -0,0 +1,267 @@
...
+    testSpace = await getDefaultSpace()
+  })
+
+  describe('Read', () => {
+    test('Get Scheduled action', async () => {
+      const createdAction = await testSpace.createScheduledAction({
+        entity: makeLink('Entry', TestDefaults.entry.testEntryId),
+        action: 'publish',
+        scheduledFor: {
+          datetime: new Date().toISOString(),
```
freakyfelt, line 22: Heads up that you might want to add a few minutes/hours to this or you'll see flakiness due to time issues

test/integration/scheduled-action-integration.ts
```diff
@@ -0,0 +1,267 @@
...
+      // Creates 2 scheduled actions
+      const [action1, action2] = await Promise.all([
+        testSpace.createScheduledAction({
+          entity: makeLink('Entry', TestDefaults.entry.testEntryId),
+          action: 'publish',
+          scheduledFor: {
+            datetime: new Date().toISOString(),
+          },
+        }),
+        testSpace.createScheduledAction({
```
freakyfelt, line 51: Same with these

test/integration/scheduled-action-integration.ts
```diff
@@ -0,0 +1,267 @@
...
+      })
+
+      expect(fetchedActions.total).to.eql(1)
+      const fetchedAction = fetchedActions.items[0]
+      expect(fetchedAction.sys.id).to.eql(createdAction.sys.id)
+      expect(fetchedAction.action).to.eql(createdAction.action)
+      expect(fetchedAction.entity).to.eql(createdAction.entity)
+
+      // cleanup
+      await fetchedAction.delete()
```
freakyfelt, line 38: Do we need to put clean up in a `finally {}` block?

test/integration/scheduled-action-integration.ts
```diff
@@ -0,0 +1,267 @@
...
+      const scheduledActions = await plainClient.scheduledActions.getMany({})
+      expect(scheduledActions.items.length).to.equal(1)
+      expect(scheduledActions.items[0].sys.id).to.equal(action.sys.id)
+
+      const { sys, ...payload } = action
+      const updatedAction = await plainClient.scheduledActions.update(
+        { scheduledActionId: sys.id, version: sys.version },
+        {
+          ...payload,
+          action: 'unpublish',
```
freakyfelt, line 254: Wait do we allow people to update the action? I thought we only allowed updating the time

Spring3: @ChidinmaOrajiaku I summon you. And most importantly, why does this test pass then 😬

Spring3: If not then having the `update()` function on an instance of a scheduled action might misbehave, when you do

Spring3: Yeah you're right

```json
{
    "sys": {
        "type": "Error",
        "id": "BadRequest"
    },
    "message": "Can only update scheduleFor.datetime and scheduleFor.timezone properties"
}
```

Spring3: I'll check the mocks there. Probably it just returns the given object

test/integration/scheduled-action-integration.ts
Spring3: I guess in case of an error, scheduled action wouldn't be created, so there will be nothing to cleanup, right?

freakyfelt: depends on where it failed (could also fail on fetch)

lib/entities/scheduled-action.ts
```diff
@@ -105,9 +106,28 @@ export function createDeleteScheduledAction(
...
+      },
+      payload,
+    }).then((data) => wrapScheduledAction(makeRequest, data))
+  }
+}
+
 export default function createScheduledActionApi(makeRequest: MakeRequest): ScheduledActionApi {
   return {
     delete: createDeleteScheduledAction(makeRequest),
+    update: createUpdateScheduledAction(makeRequest),
```
luizfonseca, line 130: ⏬ : I wish the implementation was consistent with the work in `entities/bulkAction.ts` and `entities/release.ts`

test/integration/scheduled-action-integration.ts
```diff
@@ -0,0 +1,346 @@
...
+        entity: makeLink('Asset', asset.sys.id),
+        environment: makeLink('Environment', environment.sys.id),
+        action: 'unpublish',
+        scheduledFor: {
+          datetime,
+        },
+      })
+
+      // Deletes a Release and all its Release Actions
+      await scheduledAction.delete()
```
luizfonseca, line 282:
```suggestion
      await scheduledAction.delete()
```

test/integration/scheduled-action-integration.ts
```diff
@@ -0,0 +1,346 @@
...
+
+      const response = await testSpace.getScheduledActions({
```
luizfonseca, line 284: wondering if wouldn't be better to try to find the specific scheduled action.

What I am thinking is that if for some reason another test is running this might become flakey since you querying for all scheduled actions.

luizfonseca: wait a minute, we actually don't have a normal get added to the SDK? 🤦🏽

test/integration/scheduled-action-integration.ts
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    test('lifecycle', async () => {
+      const plainClient = getPlainClient(defaultParams)
+      const entry = await plainClient.entry.get({
+        entryId: TestDefaults.entry.testEntryReferenceId,
+      })
+
+      const action = await plainClient.scheduledActions.create(
+        {},
```
luizfonseca, line 305: what's the `{}` for? 🤔

test/integration/scheduled-action-integration.ts
Spring3: yeah, I wanted to, but then I heard "please not in the same PR", so I cut the ticket instead

test/integration/scheduled-action-integration.ts
Spring3: for defaults like space id and env id. I guess the fallbacks wasn't provided so it always requires something (even an empty object 😬 ). I can change that but would be considered a breaking change? Maybe not?

luizfonseca (approved):
Looks good! 👍🏽 

Minor nitpicks: 

- Add JSDoc explaining how the use the `scheduledAction.update()` and that you have to update the object itself before calling it
- Add a test covering/using the `update` from `entities/scheduled-action.ts`

test/integration/scheduled-action-integration.ts
luizfonseca: But if this is breaking, we can leave for another PR

Spring3: I checked and it's expected to work this way. It's the same for releases and other entities as well. There people just populated it with default values while in my case default values are internally taken from the sdk client

lib/entities/scheduled-action.ts
```diff
@@ -79,35 +79,126 @@ export interface ScheduledActionQueryOptions extends BasicCursorPaginationOption
...
+     *    })
+     *    .then((scheduledAction) => console.log(scheduledAction))
+     *    .catch(console.error);
+     * ```
+     */
+    async update(): Promise<ScheduledAction> {
+      const params = getParams(this)
+
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { sys, ...payload } = this.toPlainObject()
```
luizfonseca, line 193: defining it as `_sys` wouldn't make the `eslint-disable` disappear?

Spring3: no

PR #884, opened 2021-07-14
feat: add create and createWithId to contentType in plain client

loweisz:
It adds the `create` and `createWithId` call in the plain client for `contentType`. 

## Motivation and Context
Was not implemented yet.

When adding a new method:

### DIFF
--- a/lib/plain/common-types.ts
+++ b/lib/plain/common-types.ts
@@ -232,7 +232,7 @@ export type PlainClientAPI = {
     ): Promise<ContentTypeProps>
     createWithId(
       params: OptionalDefaults<GetSpaceEnvironmentParams & { contentTypeId: string }>,
-      rawData: ContentTypeProps
+      rawData: CreateContentTypeProps
     ): Promise<ContentTypeProps>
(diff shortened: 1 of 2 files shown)


lib/plain/common-types.ts
```diff
@@ -226,6 +226,14 @@ export type PlainClientAPI = {
       rawData: ContentTypeProps
     ): Promise<ContentTypeProps>
     unpublish(params: OptionalDefaults<GetContentTypeParams>): Promise<ContentTypeProps>
+    create(
+      params: OptionalDefaults<GetSpaceEnvironmentParams>,
+      rawData: CreateContentTypeProps
+    ): Promise<ContentTypeProps>
+    createWithId(
+      params: OptionalDefaults<GetSpaceEnvironmentParams & { contentTypeId: string }>,
+      rawData: ContentTypeProps
```
andipaetzold, line 235: This now requires `sys.environment`, `description` and `displayField` to be set. Those should be optional.

andipaetzold (approved):
Approving as this behaviour matches the nested client

PR #887, opened 2021-07-15
feat(scheduledActions): add get and delete function [SPA-321]

luizfonseca:
## Summary

- Adds `getScheduledAction`, previously missing from the SDK support.

## Description

- Adds the `read` method for a scheduled Action and refactor some tests to use it.

## Motivation and Context

- It was missing from the overall SDK support

When adding a new method:

### DIFF
--- a/lib/create-space-api.ts
+++ b/lib/create-space-api.ts
@@ -1131,6 +1131,46 @@ export default function createSpaceApi(makeRequest: MakeRequest) {
       }).then((response) => wrapScheduledActionCollection(makeRequest, response))
     },
     /**
+     * Get a Scheduled Action in the current space by environment and ID
+     * @returns Promise with the Scheduled Action
+     *
+     * @example ```javascript
+     *  const contentful = require('contentful-management');
+     *
+     *  const client = contentful.createClient({
+     *    accessToken: '<content_management_api_key>'
+     *  })
+     *
+     *  client.getSpace('<space_id>')
+     *    .then((space) => space.getScheduledAction({
+     *      scheduledActionId: '<scheduled-action-id>',
+     *      environmentId: '<environmentId>'
+     *    }))
+     *    .then((scheduledAction) => console.log(scheduledAction))
+     *    .catch(console.error)
+     * ```
+     */
+    getScheduledAction({
+      scheduledActionId,
+      environmentId,
+    }: {
+      scheduledActionId: string
+      environmentId: string
+    }) {
+      const space = this.toPlainObject() as SpaceProps
+
+      return makeRequest({
+        entityType: 'ScheduledAction',
+        action: 'get',
+        params: {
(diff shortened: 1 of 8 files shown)


lib/create-space-api.ts
```diff
@@ -1130,6 +1130,46 @@ export default function createSpaceApi(makeRequest: MakeRequest) {
         params: { spaceId: raw.sys.id, query },
       }).then((response) => wrapScheduledActionCollection(makeRequest, response))
     },
+    /**
+     * Get a Scheduled Action in the current space by environment and ID
+     * @returns Promise with the Scheduled Action
```
freakyfelt, line 1135: might also say what it will do in case an action doesn't exist or they don't have permission

lib/entities/scheduled-action.ts
```diff
@@ -147,8 +147,7 @@ export default function getInstanceMethods(makeRequest: MakeRequest): ScheduledA
       }).then((data) => wrapScheduledAction(makeRequest, data))
     },
     /**
-     * Update current scheduled actions. All changes made to current instance will be saved in the database,
-     * provided the request succeeds.
+     * Update current scheduled actions. All changes made to the current object will be saved, if the request succeeds.
```
freakyfelt, line 150:
```suggestion
     * Update the current scheduled action. All changes made to the current object will be saved, if the request succeeds.
```

Perhaps also mention we currently only allow updating the `scheduledFor`

test/integration/scheduled-action-integration.ts
```diff
@@ -248,25 +232,20 @@ describe('Scheduled actions api', async function () {
...
         datetime: updatedSchedule,
         timezone: 'Europe/Kiev',
       })
-
-      // cleanup
-      await updatedAction.delete()
     })
 
-    test('fails to update unsupported property of scheduled actions', async () => {
+    test('fails to update with unsupported property changes in the scheduled action', async () => {
```
freakyfelt, line 248: This test feels like it's going to bite us in the butt someday and tests the API more than it does the client

luizfonseca: maybe I can just test a scheduled in the past (to test a failed update), as we don't need to rely in the specific fields 🤔

freakyfelt: I'd just drop it tbh. We test it in AIS

luizfonseca: 👍🏽  done

luizfonseca:
heads up: since the `delete` was already added to the plainClient, I took the liberty of just adding it to the space  client API.

lib/create-space-api.ts
luizfonseca: done 👍🏽

freakyfelt (approved):
🚢

PR #889, opened 2021-07-21
doc: add app framework documentation

shikaan:


README.md
```diff
@@ -216,6 +216,35 @@ The benefits of using the plain version of SDK are:
 - The ability to scope CMA client instance to a specific `spaceId`, `environmentId`, and `organizationId` when initializing the client.
   - You can pass a concrete values to `defaults` and omit specifying these params in actual CMA methods calls.
 
+### App Framework
+
+Starting [`@contentful/app-sdk@4`](https://github.com/contentful/app-sdk) you can use this client to issue requests
+from your [Contentful app](https://www.contentful.com/developers/docs/extensibility/app-framework/).
+
+A dedicated [Adapter](https://github.com/contentful/contentful-management.js/blob/2350b47053459694b21b19c71025632fe57815cc/lib/common-types.ts#L493-L495)
+grants your app access to the supported space-environment scoped entities without compromising on security as you won't
```
shikaan, line 225: We should link the actual entities once the docs for that part is out

marcolink (commented):
Will this "apiAdapter" solution respect the global web app rate limits, or would it create an independent request queue?

I think both cases are worth mentioning.

shikaan:
> Will this "apiAdapter" solution respect the global web app rate limits, or would it create an independent request queue?
> 
> I think both cases are worth mentioning.

Hey @marcolink thanks for the review :)

The `apiAdapter` part has been out for a few months now: when you don't specify an adapter you are getting the good ol' axios wrapper. To answer your question, it depends on the adapter implementation: in the case of the App Framework, we are effectively using the same adapter as the rest of the webapp, so it does count towards the same limits. 

In terms of behaviour it does not change much, since all the other method exposed by AppSDK (ultimately all the requests which can be issued by the custom apps and extensions) were already going through the same pipe.

Will add a paragraph to make this more explicit. Thanks again for the feedback!

README.md
```diff
@@ -290,6 +325,17 @@ Interceptor called on every request. Takes Axios request config as an arg. Defau
 
 Interceptor called on every response. Takes Axios response object as an arg. Default does nothing. Pass your own function to log any desired data.
 
+#### apiAdapter (default: `new RestAdapter(configuration)`)
+
+An [`Adapter`](https://github.com/contentful/contentful-management.js/blob/2350b47053459694b21b19c71025632fe57815cc/lib/common-types.ts#L493-L495) used to
+issue requests. It defaults to a [`RestAdapter`](https://github.com/contentful/contentful-management.js/blob/b50534c629a8ddc81637170a07bc63477d136cec/lib/adapters/REST/rest-adapter.ts)
```
andipaetzold, line 331: "used to _make_ requests"? At least the method is called `makeRequest`

README.md
```diff
@@ -290,6 +325,17 @@ Interceptor called on every request. Takes Axios request config as an arg. Defau
 
 Interceptor called on every response. Takes Axios response object as an arg. Default does nothing. Pass your own function to log any desired data.
 
+#### apiAdapter (default: `new RestAdapter(configuration)`)
+
+An [`Adapter`](https://github.com/contentful/contentful-management.js/blob/2350b47053459694b21b19c71025632fe57815cc/lib/common-types.ts#L493-L495) used to
```
andipaetzold, line 330:
```suggestion
An [`Adapter`](https://github.com/contentful/contentful-management.js/blob/2350b47053459694b21b19c71025632fe57815cc/lib/common-types.ts#L493-L495) is used to
```

README.md
```diff
@@ -216,6 +216,41 @@ The benefits of using the plain version of SDK are:
...
+  )
+
+  // ...rest of initialization code
+})
+```
+
+> **Please Note**
+>
+> Requests issued by means of App SDK's adapter will count towards the same rate limiting quota as the ones issued by
+> other APIs (e.g., Space API). Ultimately, they will all fall into the same bucket as the calls performed by the
```
andipaetzold, line 251: I am not sure what you mean with "other APIs". APIs in within the App SDK?

andipaetzold:
```suggestion
> other APIs (e.g. Space API). Ultimately, they will all fall into the same bucket as the calls performed by the
```

README.md
```diff
@@ -216,6 +216,41 @@ The benefits of using the plain version of SDK are:
...
+    }
+  )
+
+  // ...rest of initialization code
+})
+```
+
+> **Please Note**
+>
+> Requests issued by means of App SDK's adapter will count towards the same rate limiting quota as the ones issued by
```
andipaetzold, line 250: > Requests issued by means of App SDK's adapter

To me this is phrased a little complex, but maybe that's my personal impression. I'd keep the english simpler by just writing "Requests made by the App SDK adapter" ¯\_(ツ)_/¯

README.md
```diff
@@ -216,6 +216,41 @@ The benefits of using the plain version of SDK are:
 - The ability to scope CMA client instance to a specific `spaceId`, `environmentId`, and `organizationId` when initializing the client.
   - You can pass a concrete values to `defaults` and omit specifying these params in actual CMA methods calls.
 
+### App Framework
+
+Starting [`@contentful/app-sdk@4`](https://github.com/contentful/app-sdk) you can use this client to issue requests
```
andipaetzold, line 221: I think this is more explicit. Alternative: `@contentful/app-sdk` v4

```suggestion
Starting [`@contentful/app-sdk` Version 4](https://github.com/contentful/app-sdk) you can use this client to issue requests
```

README.md
```diff
@@ -216,6 +216,41 @@ The benefits of using the plain version of SDK are:
...
+from your [Contentful app](https://www.contentful.com/developers/docs/extensibility/app-framework/).
```
andipaetzold, line 222: Launch and Compose are Contentful Apps

```suggestion
from your [apps built for Contentful](https://www.contentful.com/developers/docs/extensibility/app-framework/).
```

andipaetzold (commented):
Just some nit picking

I think size and content is fine. Just some tweaks about the wording.

README.md
shikaan: > I am not sure what you mean with "other APIs". APIs in within the App SDK?

Yeah, rephrasing it to make it clearer

README.md
shikaan: Not changing this to stick with existing convention (see a bit below in the Readme)

README.md
andipaetzold: @shikaan This still needs to change because "An Adapter used to issue requests" has a wrong meaning.

andipaetzold (approved):

marcolink:
@shikaan should we put this in `draft` for now?

PR #890, opened 2021-07-22
feat(scheduledAction): Expose error on Scheduled Actions API [SPA-314]

anwaar931:
## Summary

Added error in ScheduledActionsProp

## Description

We have a [CEP](https://contentful.atlassian.net/wiki/spaces/ENG/pages/3282698277/CEP-0117+Add+error+to+scheduled+actions+payload) written with all the details and security considerations. 

## Motivation and Context

- Tracking error types for BI data warehousing (e.g. Segment)
- Exposing historical scheduled action errors in Contentful Launch
- Exposing historical scheduled action errors via customer API
- Sending errors via webhooks to external systems

When adding a new method:

### DIFF
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -31,6 +31,17 @@ enum ScheduledActionStatus {
 type SchedulableEntityType = 'Entry' | 'Asset' | 'Release'
 type SchedulableActionType = 'publish' | 'unpublish'
 
+type Collection<T> = Array<T>
+type ErrorDetail = { error: any }
+interface ScheduledActionFailedError {
+  sys: {
+    type: 'Error'
+    id: string
+  }
+  message?: string
+  details?: { errors: Collection<ErrorDetail> }
+}
+
 export type ScheduledActionSysProps = {
   id: string
   type: 'ScheduledAction'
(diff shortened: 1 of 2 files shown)


lib/entities/scheduled-action.ts
```diff
@@ -62,8 +73,26 @@ export type ScheduledActionProps = {
...
+   *   message: 'Entry is invalid',
+   *   details: {
+   *     errors: [...]
+   *   }
+   * }
+   */
+  error?: ScheduledActionFailedError
 }
 
+export type CreateUpdateScheduledActionProps = Omit<ScheduledActionProps, 'sys' | 'error'>
```
luizfonseca, line 94: maybe we can go the opposite route as well: instead of `Omit` we could `Pick` to explicitly say which props we want

lib/entities/scheduled-action.ts
```diff
@@ -31,6 +31,17 @@ enum ScheduledActionStatus {
...
 
+type Collection<T> = Array<T>
+type ErrorDetail = { error: any }
+interface ScheduledActionFailedError {
+  sys: {
+    type: 'Error'
+    id: string
+  }
+  message?: string
+  details?: { errors: Collection<ErrorDetail> }
```
luizfonseca, line 42: you can write this as `ErrorDetail[]`

lib/entities/scheduled-action.ts
```diff
@@ -31,6 +31,17 @@ enum ScheduledActionStatus {
 type SchedulableEntityType = 'Entry' | 'Asset' | 'Release'
 type SchedulableActionType = 'publish' | 'unpublish'
 
+type Collection<T> = Array<T>
```
luizfonseca, line 34: are you using the `type Collection` ?

anwaar931: Oops

luizfonseca (approved):
Looks good 👍🏽 

(maybe double check in a test somewhere that `error` is exposed in the type -- it should be fine though)

Issue #868, opened 2021-05-26
createUiExtensionWithId typing error


Xaz16:
Hello!

I'm working with version `7.22.3` of contentful-management (latest for now) and trying to follow interface for this methods, but it looks like it is not correct one
If we check CMA we will see that we should wrap extension to object with key `extension`. https://www.contentful.com/developers/docs/references/content-management-api/#/reference/ui-extensions/extension/create-update-an-extension/console/js

But typing is not allow to do it in that way, if we follow typing we will have error like

```
ValidationFailed: {
  "status": 422,
  "statusText": "Unprocessable Entity",
  "message": "Validation error",
  "details": {
    "errors": [
      {
        "name": "required",
        "details": "The property \"extension\" is required here",
        "path": [
          "extension"
        ]
      },
      {
        "name": "unexpected",
        "details": "The property \"name\" is not expected",
        "path": [
          "name"
        ]
      },
      {
        "name": "unexpected",
        "details": "The property \"fieldTypes\" is not expected",
        "path": [
          "fieldTypes"
        ]
      },
      {
        "name": "unexpected",
        "details": "The property \"sidebar\" is not expected",
        "path": [
          "sidebar"
        ]
      },
      {
        "name": "unexpected",
        "details": "The property \"srcdoc\" is not expected",
        "path": [
          "srcdoc"
        ]
      }
    ]
  },
```

Repo with reproduction: https://github.com/Xaz16/contentful-management-issues/tree/master/createUiExtensionWithId

loweisz:
Hey @Xaz16,

You are right, there was actually a mismatch between the types and the actual implementation. 
The types should be adjusted now and be correct after version `7.33.1`

PR #892, opened 2021-08-10
fix: adjust create ui extension types

loweisz:
## Summary

Closing https://github.com/contentful/contentful-management.js/issues/868

`createUiExtension` and `createUiExtensionWithId` always expects the second argument to be wrapped in extension, only the types for this weren't correct.

This adjusts the types to be what we expect in the function

When adding a new method:


andipaetzold (approved):
👍

suevalov (approved):

PR #896, opened 2021-08-22
feat: field groups props in editor layout [DANTE-173]

Silhoue:
## Summary

Add types for `groupControls` and `editorLayout` properties of the `EditorInterface`.

### DIFF
--- a/lib/entities/editor-interface.ts
+++ b/lib/entities/editor-interface.ts
@@ -5,11 +5,7 @@ import { MetaSysProps, MetaLinkProps, DefaultElements, MakeRequest } from '../co
 import { wrapCollection } from '../common-utils'
 import { DefinedParameters } from './widget-parameters'
 
-export interface Control {
-  /**
-   * ID of the customized field
-   */
-  fieldId: string
+interface WidgetConfig {
   /**
    * Type of the widget used
    */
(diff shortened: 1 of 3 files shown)


lib/entities/editor-interface.ts
```diff
@@ -72,6 +93,10 @@ export type EditorInterfaceProps = {
    * Array of fields and it's associated widgetId
    */
   controls?: Control[]
+  /**
+   * Array of groups and it's associated widgetId
```
IBratWurst, line 97: Probably cloned from above
```suggestion
   * Array of groups and their associated widgetId
```

lib/export-types.ts
```diff
@@ -41,7 +41,9 @@ export type {
   EditorInterface,
   EditorInterfaceProps,
   Control,
+  GroupControl,
   Editor,
+  EditorLayoutItem,
```
IBratWurst, line 46: Should `FieldItem` also be exported?

lib/entities/editor-interface.ts
Silhoue: fixed

lib/export-types.ts
Silhoue: added

veu (approved):

IBratWurst (approved):

PR #900, opened 2021-08-31
Fix some scheduled action comments

zdhickman:
## Summary

Updates example object structure in comments for `createScheduledAction` and `updateScheduledAction`.

## Description

Updates example object structure in comments for `createScheduledAction` and `updateScheduledAction`.

- `environment` object should be in `environment.sys`
- `dateTime` should be `datetime`

## Motivation and Context

Just saving some future dev a headache 😄 

When adding a new method:


Spring3 (approved):
Thanks for adjusting these!

Spring3:
Successor: #1105 . I had to open the other PR with the same changes as this one looks stale.

Thank you nonetheless for spotting it and opening the original PR addressing these.

PR #961, opened 2021-09-10
docs: exclude internal types/constants/functions

andipaetzold:
Due to a major refactoring in version 7.13.0 the size of the type docs was significantly increased.

This PR marks many types/constants/functions as `@private` to exclude them from the generated documentation.

This removed the size of the `globals.html` by 20k lines.


Jwhiles (approved):
ace!

PR #972, opened 2021-09-20
feat: [EXT-3053] implement app details methods

Jwhiles:

When adding a new method:


lib/entities/app-metadata.ts
```diff
@@ -0,0 +1,86 @@
...
+import { Except } from 'type-fest'
+import { BasicMetaSysProps, DefaultElements, MakeRequest, SysLink } from '../common-types'
+import enhanceWithMethods from '../enhance-with-methods'
+
+type AppMetadataSys = Except<BasicMetaSysProps, 'version' | 'id'> & {
+  appDefinition: SysLink
+  organization: SysLink
+}
+
+// TODO should these be exported?
```
Jwhiles, line 12: ?

tsconfig.json
```diff
@@ -34,7 +34,7 @@
       "AppInstalation",
       "AppBundle",
       "AppSignedRequest",
-      "AppSigningSecret",
+      "AppDetails",
```
andipaetzold, line 37: maybe also reorder things alphabetical

```suggestion
      "AppSigningSecret",
      "AppDetails",
```

andipaetzold (changes requested):

lib/create-organization-api.ts
```diff
@@ -684,5 +686,80 @@ export default function createOrganizationApi(makeRequest: MakeRequest) {
         /* noop*/
       })
     },
+    /**
+     * Creates or updates an app signing secret
```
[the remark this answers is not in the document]
Jwhiles: Thanks! 

I thought I got all of these.

lib/create-organization-api.ts
```diff
@@ -696,7 +696,9 @@ export default function createOrganizationApi(makeRequest: MakeRequest) {
      * })
      *
      * client.getOrganization('<org_id>')
-     * .then((org) => org.upsertAppDetails('app_definition_id', { value: 'tsren3s1....wn1e' }))
+     * .then((org) => org.upsertAppDetails('app_definition_id',
```
andipaetzold, line 699: It feels strange to me that this method is on org level and not part of the app definition api (https://github.com/contentful/contentful-management.js/blob/master/lib/create-app-definition-api.ts) but we already did the same for the signing secret 🤷

andipaetzold (approved):

loweisz (approved):
🚀

PR #973, opened 2021-09-20
feat(adatper): single axios instance

marcolink:
Create a single axios instance to be used in `makeRequest`.

When adding a new method:


andipaetzold (changes requested):
We now always overwrite the `Content-Type` and `X-Contentful-User-Agent`. So, you can never specify these headers when creating the CMA client or when an endpoint requires different values.

Especially the `Content-Type` needs to be overwritten for some endpoints. E.g. https://github.com/contentful/contentful-management.js/blob/16c89f34b349ab4d8a762369c2b4eb714101cbf1/lib/adapters/REST/endpoints/entry.ts#L57

marcolink:
@andipaetzold 💡 
If I understand you correctly, flipping the order does the trick, right?

andipaetzold:
> @andipaetzold 💡
> If I understand you correctly, flipping the order does the trick, right?

We want the following priority to keep things unchanged (lower number takes priority):
1. endpoint headers
1. createClient args
1. `requiredHeaders`


The current PR (after flipping) has

1. endpoint headers
1. `requiredHeaders`
1. createClient args

andipaetzold (approved):

PR #983, opened 2021-09-24
chore: update CI & .nvmrc to node 14

andipaetzold:
## Summary

* Updates CI to node 14
* Updates `.nvmrc` to node 14

I didn't update `engines` in `package.json` because this package is not node only but also runs in the browser

This is a `chore` because according to the README we only support node LTS anyway

## Motivation and Context

Node 12 is EOL since April and packages like `semantic-release` do not support node 12 anymore.

CI on `master` is currently broken because of `semantic-release`.


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   lint:
     docker:
-      - image: circleci/node:12
+      - image: circleci/node:14
```
damienxy, line 5: `circleci/node` is [legacy](https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/), I suggest to use `cimg/node:14.17`

andipaetzold: Will update - good catch

Any reason for using 14.17 over the latest version?

damienxy: Because you [need to specify](https://github.com/CircleCI-Public/cimg-node#tagging-scheme) the minor version, and 14.17 is currently the latest

andipaetzold: Okay, I'll update the PR :)

damienxy (approved):

PR #1010, opened 2021-10-07
fix(rest-adapter): headers

marcolink:
The change to single axios instance (#973) dropped all headers, except those defined in `params.config.headers` .
This change adds the calculated set of headers to that field.

When adding a new method:


marcolink:
closed in favor of #1984

PR #1037, opened 2021-10-21
fix: add missing accessToken property on preview API key response

kamsar:
## Summary

This property is returned by the API, but not exposed in the typings. Without this, you can fetch preview API keys, but not read their actual key value unless you ignore typings.

The property is part of the API docs here: https://www.contentful.com/developers/docs/references/content-management-api/#/reference/api-keys/preview-api-key/get-a-single-preview-api-key/console/curl

## Motivation and Context

I'd love to get rid of an `as any` in my codebase :)


Spring3 (approved):
Thank you @kamsar 

Sorry for the delay

Spring3:
@suevalov do you have the rights to merge? The button looks disable to me and the checks don't see to work correctly

PR #1072, opened 2021-11-10
fix: add sane defaults to mitigate rogue endpoints

anho:
## Summary

Some endpoint like editor_interfaces do not return skip and limit params, that needs to be mitigated to properly consume items.

### DIFF
--- a/lib/plain/as-iterator.ts
+++ b/lib/plain/as-iterator.ts
@@ -17,7 +17,7 @@ export const asIterator = <P extends QueryParams, T, F extends IterableFn<P, T>>
       return {
         current: 0,
         async next() {
-          const { total, items, skip, limit } = await currentResult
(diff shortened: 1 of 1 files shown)


lib/plain/as-iterator.ts
```diff
@@ -17,7 +17,7 @@ export const asIterator = <P extends QueryParams, T, F extends IterableFn<P, T>>
       return {
         current: 0,
         async next() {
-          const { total, items, skip, limit } = await currentResult
+          const { total = 0, items = [], skip = 0, limit = 100 } = await currentResult
```
marcolink, line 20: Is it necessary to also default `total` and `items`?

anho: Don't think so, just being super cautious here.

marcolink (approved):

PR #1096, opened 2021-11-30
fix: typing for reference endpoints [SPA-523]

### LINKED ISSUE #932: getEntryReferences is quite wrong
Looking at this API when used on the Contentful side the argument should be `include`.
image

This isn't included in the options.
https://github.com/contentful/contentful-management.js/blob/7d2a0e20e36081978e4a376ff9dc42480c2d6c01/lib/create-environment-api.ts#L800-L812

In addition the return typings are wrong. CollectionProp is used.
https://github.com/contentful/contentful-management.js/blob/7d2a0e20e36081978e4a376ff9dc42480c2d6c01/lib/entities/entry.ts#L30-L31

This puts the results inside an item prop with Entry and Assets respectively. 
https://github.com/contentful/contentful-management.js/blob/7d2a0e20e36081978e4a376ff9dc42480c2d6c01/lib/common-types.ts#L156

This does not match the API response. Entry and Asset should both directly be an array.
image

Inbal-Gordon:
## Summary
Fix references endpoint.

### Fixed the type on `EntryReferenceProps`:
Previously the `includes.Entry` and `includes.Asset` had type `CollectionProps<T>`, while in fact both were an array. 

- `includes.Entry` is now `EntryProps[]`
- `includes.Asset` is now `AssetProps[]`

### Deprecated `maxDepth` on `plainClient.entry.references` and `environmentApi.getEntryReferences` calls
- added a `include` param to match the actual query parameter that is sent to the servers.
- `maxDepth` param works the same way, but is marked as `@deprecated` and now acts more like an alias for `include` param. Please use `include` param instead of `maxDepth`

## Description
resolves #932 

When adding a new method:

### DIFF
--- a/lib/create-environment-api.ts
+++ b/lib/create-environment-api.ts
@@ -799,7 +799,8 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
     /**
      * Get entry references
      * @param entryId - Entry ID
-     * @param {Object} options.maxDepth - Level of the entry descendants from 1 up to 10 maximum
+     * @deprecated @param {Object} options.maxDepth - alias for `include`
+     * @param {Object} options.include - Level of the entry descendants from 1 up to 10 maximum
      * @returns Promise of Entry references
      * @example ```javascript
      * const contentful = require('contentful-management');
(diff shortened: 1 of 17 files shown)


lib/create-environment-api.ts
```diff
@@ -811,10 +812,10 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
      * // Get entry references
      * client.getSpace('<space_id>')
      * .then((space) => space.getEnvironment('<environment_id>'))
-     * .then((environment) => environment.getEntryReferences('<entry_id>', {maxDepth: number}))
+     * .then((environment) => environment.getEntryReferences('<entry_id>', {include: number}))
      * .then((entry) => console.log(entry.includes))
      * // or
-     * .then((environment) => environment.getEntry('<entry_id>')).then((entry) => entry.references({maxDepth: number}))
+     * .then((environment) => environment.getEntry('<entry_id>')).then((entry) => entry.references({include: number}))
```
luizfonseca, line 818: Probably will need to update the `docs-app` as well

luizfonseca:
My concern here is that this PR might be doing more than necessary (for this PR):

- **First**: bug fix on the types (which I agree, are incorrect and need fixing)
- **Second**: deprecating the param because the name is not the same as in the URL (which I think we did on purpose because `include` doesn't explain what I does and from a client library not all function args need to be named exactly the same)


In short, I'd just move the deprecation part to another PR (that gives us a clear version bump)

Spring3 (approved):

PR #1098, opened 2021-12-01
fix: contentful-sdk-core v6.10.4

johanneswuerbach:
Ensure contentful/contentful-sdk-core#239 is being used.


marcolink (approved):
:shipit:

PR #1099, opened 2021-12-01
chore: update typedoc options for latest update

andipaetzold:
In https://github.com/contentful/contentful-management.js/pull/1073 `typedoc` had a major version bump which which changed its behaviour and the available options. This PR removed the broken options and specifies the now required entry point.

As we are specifying a single file as entry point, this file must also export all available types, which is done via `export * from './export-types'` in `lib/contentful-management.ts`.

With this change all types can not only be imported via `contentful-management/types` but also from `contentful-management` directly. Is that an issue? I don't know why `types.d.ts` was introduced in the first place.


Spring3 (approved):
Thanks a lot for fixing this for now. 🥇

marcolink (approved):
:shipit:

PR #1114, opened 2021-12-09
feat: introduce alpha client, add WorkflowDefinition

Chaoste:
Starting work on our new feature "Workflows", we want to use the CMA adapter in our new [workflows-marketplace-adapter](https://github.com/contentful/workflows-marketplace-app) for the new endpoints with an alpha header.

## Summary
Since the API is protected by an alpha flag, we want to exclude the feature also from the default CMA client. Only by passing an alpha flag, the type of the client will contain the new `workflowDefinition` endpoint. Besides the types, the logic will check for the exact alpha flag value to include this feature.

## Usage
1. By passing the `alphaFeature` parameter, the returned client type supports `workflowDefinition`
Screenshot 2021-12-09 at 10 21 39
2. When the `alphaFeature` parameter is missing, the returned client type doesn't support `workflowDefinition`
Screenshot 2021-12-09 at 10 22 11

Code Example:
```
const client = createClient(
  {
    space: spaceId,
    accessToken,
  },
  { type: 'plain', alphaFeature: ['cf-hidden-workflow-alpha-flag'] }
)

client.workflowDefinition.create(
  { environmentId },
  {
    name: 'MyNewWorkflow',
    steps: [{}],
    appliesTo: [],
    permissions: [],
  }
)

client.workflowDefinition.get({
  environmentId,
  workflowDefinitionId: 'test',
})

client.workflowDefinition.update(
  {
    environmentId,
    workflowDefinitionId: 'test',
  },
  {
    name: 'MyRenamedWorkflow',
    steps: [{}],
    appliesTo: [],
    permissions: [],
    sys: {
      version: 1,
    },
  }
)

client.workflowDefinition.delete({
  environmentId,
  workflowDefinitionId: 'test',
  version: 2,
})

```

## Motivation and Context

In general, we will add new entities & methods in the future to our API. Since the app framework encourages to use the CMA adapter for all kinds of queries, we want to provide a new structure to include alpha flagged features in the future.

*We don't want to document alpha features*

When adding a new method:


lib/contentful-management.ts
```diff
@@ -61,13 +61,22 @@ function createClient(
     defaults?: DefaultParams
   }
 ): PlainClientAPI
+function createClient(
+  params: ClientOptions,
+  opts: {
+    type: 'plain'
+    alphaFeature: string[]
```
giotiskl, line 68: I saw in some places we renamed to `alphaFeatures` (plural) but here is still singular, is it intentional?

lib/entities/workflow-definition.ts
```diff
@@ -0,0 +1,113 @@
...
+  type: 'WorkflowDefinition'
+  space: SysLink
+  environment: SysLink
+}
+
+export type WorkflowDefinitionProps = {
+  sys: WorkflowDefinitionSysProps
+  name: string
+  // TODO: Add new type for 'ValidatedLink' with validations property
+  appliesTo: any[]
```
giotiskl, line 27: this property has be renamed to `applicableEntities` in the API

lib/entities/workflow-definition.ts
```diff
@@ -0,0 +1,113 @@
...
+export type WorkflowDefinitionSysProps = Pick<
+  BasicMetaSysProps,
+  'id' | 'version' | 'createdAt' | 'createdBy' | 'updatedAt' | 'updatedBy'
+> & {
+  type: 'WorkflowDefinition'
+  space: SysLink
+  environment: SysLink
+}
+
+export type WorkflowDefinitionProps = {
```
giotiskl, line 23: we're missing the `description` property (type string, optional)

lib/entities/workflow-definition.ts
```diff
@@ -0,0 +1,113 @@
...
+  sys: WorkflowDefinitionSysProps
+  name: string
+  // TODO: Add new type for 'ValidatedLink' with validations property
+  appliesTo: any[]
+  // TODO: Use WorkflowStep as soon as it was added
+  // Require length min. 1 with `steps: [any, ...any[]]`
+  steps: any[]
```
giotiskl, line 30: type for steps should be:
```
type WorkflowStep = {
  id: string;
  name: string;
  description?: string;
  annotations?: string[];
}

giotiskl (approved):

PR #1122, opened 2021-12-15
feat: improve PlainClientAPI type to listen to an explicit string

Chaoste:
## Summary

The alpha feature introduced in https://github.com/contentful/contentful-management.js/pull/1114 should be protected by a `alphaFeatures` parameter which needs to contain a certain string to add the workflows endpoint to the plain client. Now, typescript understands whether the array really includes this string. If not, the type of the returned client doesn't contain `workflowDefinition`.

## Description

I started with this [magic solution to a similar problem](https://stackoverflow.com/a/57958451/4816930). I did some testing [on this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggTnAhiAsgSwHYHkPQLxQDaAzsHJgOYA0UAdPaeRhYQLqsBQHAxgPYakoADwBcsBMnTZcUAoQBEwCKXmcOoSOKQgAYrzjcIAEwCSSgLYAeDlCgAlKBCFKMR4lEaUqNqABVHzhCu7gAUdhCIRvwANiDw2paezAB8UAA+UHARURixRHasAJRQAGQ+tgDeRADSUJhQANYQILwAZn6sYlWEAAp1GFDVnfZQAL5jhBgAruYARhBwnKkEvlwa0PHIegbGZhDmAEzWtv5OLm5QYdkxcRIgiWSUqRlZkTdEAOSziHAfRaXlKDdWr1Jotdq+YbdPr1IZiL4-D5jCbTOYLViyVKYVoLPwcZZ49TgDZ3baGUwWADMljsBOB-UazTaHS6RBhAzhI3Goy4XD4AmAUFaYk2un05L2h1kRHkAAteFNiBB5DR5N84CqoPJzAqlao+fxBCARaTxbsLJYEb8aAp1Zq5brlaqdYrlawCbafvb5a77S69Zx+YIAF4m7Rk837S3qj42h2+1V252O1Qe+N6xNe5O+tS8owQbjRH7QVpTDDcYBofhCs0mDCi44jM5BC5Jag+U6BYKXcJvXK3BJt56Za79-L-Mq2So1Blg5mQ1m9BmchzcyYzeaLfEhLIARymaCyRjEdhomDAU2AYl8hTEADdeGgjABuLgAejfNZ2ddFIQ+AE5YyIIpnygD9HAQfQOHA1pa3rO4-0Am0PgARiAj4Dj+QpQPAhY4CgmC4N-ACgKqAAGeFALGbCwM-PCoI4WDv3g7REKAwgSJA2ioF4BpkEYoiEJI5CkKgTiaPA3j+KYwwfyE0SONEjD0MpdCABYsJwz8pJAATmOIhTUPQpTVJoD4NK4yS+N0mSIDk1jhM+NCzMwsylIsiTtOsvTZJY5A2OQ5yxNcsTTLEwDLK86TBIcwyyPQoLlLMsLzPQgBWdCADZ0IAdnQgAOYzNO4nS1mJLRkAAdTQYBZQAIR+SxfAJTtzlCXscliBsh3SEc+zyDiYwnIEZ1BJkIShNll2GK0kW5KAAH4-CgMRcDvBZXygDhMCUOBWkQQwoAAUXMRA0GiIEfHMZRiEQCgIDENtXx5LadoWfbDoAEV4ChLtsdUGhCW8oAfJ9npsIlNBOs7ohQG67ogABhfgXGAdwCDh4hbvurBWksaHzuSTapxJ0mtvWKBvooTHsaRlGgjR6UaYR3HLCponATJqcys0PReCagI2pGlD4QAd30BpWmiXhReIOaWoNAUoEQMQ+csDjxbgSXpdlv4PQ+dDNe1mW5dYIA) but couldn't create a type that defines an array with a certain value. I can only link it to the function declaration.

If there will be more alpha features at the same time in this client, we will need to revisit the logic and try to better mimick the above mention SO post. Otherwise, we would need to define a type for every order of named alpha features.

## Motivation and Context

Better typings

When adding a new method:

### DIFF
--- a/lib/contentful-management.ts
+++ b/lib/contentful-management.ts
@@ -9,7 +9,11 @@ import type { RestAdapterParams } from './adapters/REST/rest-adapter'
 import type { MakeRequest } from './common-types'
 import { AdapterParams, createAdapter } from './create-adapter'
 import createContentfulApi, { ClientAPI } from './create-contentful-api'
-import type { AlphaPlainClientAPI, PlainClientAPI } from './plain/common-types'
+import type {
+  AlphaPlainClientAPI,
+  AlphaWorkflowExtension,
+  PlainClientAPI,
+} from './plain/common-types'
 import type { DefaultParams } from './plain/plain-client'
 import { createPlainClient } from './plain/plain-client'
 import * as editorInterfaceDefaults from './constants/editor-interface-defaults'
@@ -61,6 +65,17 @@ function createClient(
     defaults?: DefaultParams
   }
 ): PlainClientAPI
+function createClient<
+  T extends (ReadonlyArray<string> | readonly ['workflows']) &
+    { [K in keyof T]: { [P in K]: 'workflows' } }[number]
+>(
+  params: ClientOptions,
+  opts: {
+    type: 'plain'
+    alphaFeatures: T
+    defaults?: DefaultParams
+  }
+): PlainClientAPI & AlphaWorkflowExtension
 function createClient(
   params: ClientOptions,
   opts: {
@@ -68,7 +83,7 @@ function createClient(
     alphaFeatures: string[]
     defaults?: DefaultParams
   }
-): AlphaPlainClientAPI
+): PlainClientAPI
 function createClient(
   params: ClientOptions,
   opts: {
--- a/lib/contentful-management.ts
+++ b/lib/contentful-management.ts
@@ -9,11 +9,7 @@ import type { RestAdapterParams } from './adapters/REST/rest-adapter'
 import type { MakeRequest } from './common-types'
 import { AdapterParams, createAdapter } from './create-adapter'
 import createContentfulApi, { ClientAPI } from './create-contentful-api'
-import type {
-  AlphaPlainClientAPI,
-  AlphaWorkflowExtension,
-  PlainClientAPI,
-} from './plain/common-types'
+import type { AlphaPlainClientAPI, PlainClientAPI } from './plain/common-types'
 import type { DefaultParams } from './plain/plain-client'
 import { createPlainClient } from './plain/plain-client'
 import * as editorInterfaceDefaults from './constants/editor-interface-defaults'
@@ -66,6 +62,7 @@ function createClient(
   }
 ): PlainClientAPI
 function createClient<
+  // T describes an array that contains the string "workflows" at any point in the list.
   T extends (ReadonlyArray<string> | readonly ['workflows']) &
     { [K in keyof T]: { [P in K]: 'workflows' } }[number]
 >(
@@ -75,7 +72,7 @@ function createClient<
     alphaFeatures: T
     defaults?: DefaultParams
   }
-): PlainClientAPI & AlphaWorkflowExtension
+): AlphaPlainClientAPI
 function createClient(
   params: ClientOptions,
   opts: {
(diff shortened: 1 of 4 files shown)


lib/contentful-management.ts
```diff
@@ -61,14 +65,25 @@ function createClient(
     defaults?: DefaultParams
   }
 ): PlainClientAPI
+function createClient<
+  T extends (ReadonlyArray<string> | readonly ['workflows']) &
+    { [K in keyof T]: { [P in K]: 'workflows' } }[number]
+>(
```
Chaoste, line 71: This describes an array that contains "workflows" somewhere in the list.

Spring3: Sounds like this should be a comment in code :p

lib/contentful-management.ts
```diff
@@ -61,14 +65,25 @@ function createClient(
...
+  params: ClientOptions,
+  opts: {
+    type: 'plain'
+    alphaFeatures: T
+    defaults?: DefaultParams
+  }
+): PlainClientAPI & AlphaWorkflowExtension
```
Spring3, line 78: Does it mean that when someone tries to use it they'd have to manually cast it to `as PlainClientAPI & AlphaWorkflowExtension`? Perhaps it would be better to define a type for it?

lib/contentful-management.ts
Spring3: Generally, this is very unclear to me. Could you please define a private type for it and name it so it defines the result. If it's a map or a string or an array. Too many layered nesting in types 😓

lib/contentful-management.ts
```diff
@@ -61,14 +65,25 @@ function createClient(
...
 function createClient(
   params: ClientOptions,
   opts: {
     type: 'plain'
     alphaFeatures: string[]
     defaults?: DefaultParams
   }
-): AlphaPlainClientAPI
+): PlainClientAPI
```
Spring3, line 86: Are you sure about this change? Sounds like this could break a lot of typescript projects

lib/plain/common-types.ts
```diff
@@ -857,4 +857,6 @@ export type AlphaExtensions = {
   }
 }
 
+export type AlphaExtensions = AlphaWorkflowExtension
```
Spring3, line 860: what is the benefit of doing so vs leaving it the way it was? I don't see the difference, but now we have 2 types which mean the same thing but named differently

lib/contentful-management.ts
Chaoste: I just invented this alpha api on last friday (the one that wasn't released until now)

lib/plain/common-types.ts
Chaoste: I guess my goal was to start with a more general extensions type that would later be a union of all alpha features. i'll remove it for now

lib/contentful-management.ts
Chaoste: Screenshot 2021-12-14 at 15 17 00
Screenshot 2021-12-14 at 15 17 13
When using the alpha client, you don't need to define it manually. As you can see in the screenshot, it will infer the correct type.

lib/plain/common-types.ts
Chaoste: Screenshot 2021-12-15 at 15 57 18

I would like to keep it for this helper function. Makes more sense to have a general union type for all alpha features to come

Spring3 (approved):
Ok, makes sense. Thanks for the rationale around the changes made

PR #1126, opened 2021-12-20
fix: move permissions property fron wf def to wf step

Chaoste:
## Summary

Update the typings from workflow definition and steps to match the original API proposal.
- Permission is part of the step entity
- "applicableEntities" should be "appliesTo"

### DIFF
--- a/lib/entities/workflow-definition.ts
+++ b/lib/entities/workflow-definition.ts
@@ -62,7 +62,7 @@ export type WorkflowDefinitionProps = {
   sys: WorkflowDefinitionSysProps
   name: string
   description?: string
-  applicableEntities?: WorkflowDefinitionValidationLink[]
+  appliesTo?: WorkflowDefinitionValidationLink[]
   steps: WorkflowStepProps[]
 }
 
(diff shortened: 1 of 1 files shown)


lib/entities/workflow-definition.ts
```diff
@@ -13,6 +13,34 @@ import { wrapCollection } from '../common-utils'
 import enhanceWithMethods from '../enhance-with-methods'
 import { ActionType } from './role'
 
+/* Workflow Step */
+
+export type WorkflowStepAction = {
+  action: Link<'AppAction'>
+  body: string
+  headers: Record<string, unknown>
```
FLoppix, line 21: I think headers can be of type `Record<string, string>`

FLoppix (approved):
just one small comment otherwise LGTM

lib/entities/workflow-definition.ts
Chaoste: I'm not sure if this is the same purpose but we use this headers type for every `headers` property in CMA: https://github.com/contentful/contentful-management.js/blob/724ff87a48de1d55d21392af198ed5e58be008d9/lib/common-types.ts#L578

PR #1137, opened 2022-01-04
feat: add support for wf definition query []

ronaldronson:
## Summary

Add support for the workflow definition query feature.

## Description

Allow querying workflows in the environment, support skip and limit.

When adding a new method:


Chaoste (commented):
Changes look good to me 👍  Maybe, one suggestion: Should we rather use `getMany` instead of `getAll`. I checked `MRActions` type and saw that in 25 cases we use `getMany` and only in two `getAll` (tasks and comments). For consistency, I ould stick to `getMany` since it also supports query parameters, so it's not always "all".

Chaoste (approved):
LGTM!

PR #1139, opened 2022-01-10
feat: [EXT-3396] add active property to webhooks

Jwhiles:
## Summary
Add active property to webhooks. It will always be defined in the API response, but is optional when creating and updating webhooks. It not present, it defaults to true.

When adding a new method:

### DIFF
--- a/lib/entities/webhook.ts
+++ b/lib/entities/webhook.ts
@@ -182,6 +182,11 @@ export type WebhookProps = {
    * Transformation to apply
    */
   transformation?: WebhookTransformation
+
+  /**
+   * Whether the Webhook is active. If set to false, no calls will be made
+   */
+  active?: boolean
 }
 
 export interface WebHooks extends WebhookProps, DefaultElements<WebhookProps> {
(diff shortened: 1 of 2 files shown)


lib/entities/webhook.ts
```diff
@@ -182,6 +182,11 @@ export type WebhookProps = {
    * Transformation to apply
    */
   transformation?: WebhookTransformation
+
+  /**
+   * Whether the Webhook is active. If set to false, no calls will be made
+   */
+  active: boolean
```
kdamball, line 189: Make it optional?

kdamball (approved):

lib/entities/webhook.ts
Jwhiles: It's optional in the creation and update parameters, but always present in the response. 

See https://github.com/contentful/contentful-management.js/pull/1139/files#diff-03db97242908183de4e74d324bbb2877366c4cac40e405f9d65e6d247d404343R76-R80

PR #1140, opened 2022-01-14
feat: organizations query param support [SPA-50]

Spring3:
## Summary

Adds ability to pass pagination query parameters to
`plainClient.organization.getAll()` and `client.getOrganizations()`

## Description

Before it would only fetch up to 25 organizations. With the changes in this PR, there will be ability to fetch more organizations.

## Motivation and Context

Users who had more than 25 organizations were not able to see more than 25 in response

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/organization.ts
+++ b/lib/adapters/REST/endpoints/organization.ts
@@ -1,18 +1,23 @@
 import type { AxiosInstance } from 'contentful-sdk-core'
-import { CollectionProp, GetOrganizationParams } from '../../../common-types'
+import { CollectionProp, GetOrganizationParams, PaginationQueryParams } from '../../../common-types'
 import { OrganizationProp } from '../../../entities/organization'
 import { RestEndpoint } from '../types'
 import * as raw from './raw'
 
-export const getMany: RestEndpoint<'Organization', 'getMany'> = (http: AxiosInstance) => {
-  return raw.get<CollectionProp<OrganizationProp>>(http, `/organizations`)
+export const getMany: RestEndpoint<'Organization', 'getMany'> = (
+  http: AxiosInstance,
+  params?: PaginationQueryParams
+) => {
+  return raw.get<CollectionProp<OrganizationProp>>(http, `/organizations`, {
+    params: params?.query,
+  })
 }
 
 export const get: RestEndpoint<'Organization', 'get'> = (
   http: AxiosInstance,
   params: GetOrganizationParams
 ) => {
-  return getMany(http).then((data) => {
+  return getMany(http, { query: { limit: 100 } }).then((data) => {
     const org = data.items.find((org) => org.sys.id === params.organizationId)
     if (!org) {
       const error = new Error(
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -11,6 +11,11 @@ if (process.env.API_INTEGRATION_TESTS) {
 
 const env = process.env !== undefined ? process.env : window.__env__
 
+// Please do not replace it with testUtils.initClient(), becuase test-utils repo has contentful-management
+// as dependency and will make it impossible to test new changes to client's api.
+// Eg: getAll() fn in prod doesn't have any args. I change it in my PR to getAll({ query })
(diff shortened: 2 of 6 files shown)


lib/adapters/REST/endpoints/organization.ts
```diff
@@ -1,18 +1,23 @@
...
+    params: params?.query,
+  })
 }
 
 export const get: RestEndpoint<'Organization', 'get'> = (
   http: AxiosInstance,
   params: GetOrganizationParams
 ) => {
-  return getMany(http).then((data) => {
+  return getMany(http, { query: { limit: 100 } }).then((data) => {
```
freakyfelt, line 20: Hmm... do we really need to pass this query param here given it's a `get` operation that should only return 1 item?

Spring3: I am not very sure but I thought that without it, setting an increased limit for `getAll()` could backfire later when `get()` is called because it would still get 25 orgs

test/helpers.js
```diff
@@ -11,7 +11,13 @@ if (process.env.API_INTEGRATION_TESTS) {
...
 const env = process.env !== undefined ? process.env : window.__env__
 
-export const initClient = () => testUtils.initClient()
+export const initClient = () => {
+  const accessToken = env.CONTENTFUL_INTEGRATION_TEST_CMA_TOKEN
+  return createClient({
+    accessToken,
+    ...params,
+  })
+}
```
freakyfelt, line 20: Is this a leftover change or what's the reason we swapped away from testUtils.initClient()

Spring3: `testUtils.initClient` does the same internally. The issue was that I couldn't test my changes because `testUtils` has this repo as dependency and the version they have doesn't yet have the changes i made, so when I tried to call `getAll` it had no args in the types

lib/adapters/REST/endpoints/organization.ts
freakyfelt: right but if you continue reading the code it's expecting to find a single object in `data.items` that matches the `params.organizationId`, meaning this is using the getMany to getOne or None

test/helpers.js
freakyfelt: Can you add a TODO here? Or should we just strip that out entirely from the repo? My fear is that we're going to drift and there will be some bad expectations

lib/adapters/REST/endpoints/organization.ts
Spring3: Yeah, I understand, but my point is that without the change that I made, if someone has let's say 26 orgs and then they use get() in attempt to search for the org that is last (26th)? They will get an error, although the org would be there. With my custom limit it will fetch 100 orgs and there's a higher chance that none of the customers will have an account that would have more than 100 orgs.

Ideally we should have smth like fetchEntireDataset in there, but it comes from a different package, which I can't import cause it would be a cyclic ref.

test/helpers.js
Spring3: Added a comment. I think it should do for now, but I share your concern and there is a high chance of it happening, so in my opinion, we should delete it from test-utils as well.

Spring3: But for that we need to research where else it is used to not break something on accident. If the same function from test-utils is used elsewhere, then I am afraid that this comment is all we can do

lib/adapters/REST/endpoints/organization.ts
freakyfelt: OH! Duh I see what you're saying now. We aren't passing in a filter by sys.id in this query

freakyfelt (approved):
🚀

lib/adapters/REST/endpoints/organization.ts
freakyfelt: (to which I would argue Someday® we should make sure the API supports it and then pass in a `sys.id[in]` filter here so we only have to fetch the one)

Spring3: if you don't mind, I would like to delegate it to the team owning the endpoint as it's a bit of a larger effort than what I am trying to achieve in this PR

PR #1150, opened 2022-01-19
chore: update release query options [SPA-507]

ChidinmaOrajiaku:
## Summary

Add `order` to release query options.

When adding a new method:

### DIFF
--- a/lib/entities/release.ts
+++ b/lib/entities/release.ts
@@ -40,6 +40,10 @@ export interface ReleaseQueryOptions {
    * @default -sys.updatedAt
    * */
   order?: string
+  /**
+   * Filters by creator of release
+   */
+  'sys.createdBy.sys.id'?: string
 }
 
 export type ReleaseSysProps = {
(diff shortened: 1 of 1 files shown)


luizfonseca:
can we also add the `sys.createdBy.sys.id` filter?

ChidinmaOrajiaku:
> can we also add the `sys.createdBy.sys.id` filter?

Done

Spring3 (approved):

lib/entities/release.ts
```diff
@@ -35,6 +35,15 @@ export interface ReleaseQueryOptions {
...
   limit?: number
+  /**
+   * Order releases by
+   * @default -sys.updatedAt
+   * */
+  order?: string
+  /**
+   * Filters by creator of release
+   */
+  'sys.createdBy.sys.id'?: string
```
luizfonseca, line 46: I think it's 
```suggestion
  'sys.createdBy.sys.id[in]'?: string
```

luizfonseca (approved):
Minor fix but 🚀

PR #1161, opened 2022-01-20
feat: add workflow entity

ronaldronson:
## Summary

Adding a new entity called Workflows under the alpha feature.

## Description

Adding typings and methods:
- create
- update
- query
- delete
- complete

No documentation update for alpha features.

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/workflow.ts
+++ b/lib/adapters/REST/endpoints/workflow.ts
@@ -1,0 +1,81 @@
+import { AxiosInstance } from 'axios'
+import copy from 'fast-copy'
+import { SetOptional } from 'type-fest'
+import { CollectionProp, GetSpaceEnvironmentParams, GetWorkflowParams } from '../../../common-types'
+import {
+  CompleteWorkflowParams,
+  CreateWorkflowParams,
+  CreateWorkflowProps,
+  DeleteWorkflowParams,
+  UpdateWorkflowProps,
+  WorkflowProps,
+  WorkflowQueryOptions,
+} from '../../../entities/workflow'
+import { RestEndpoint } from '../types'
+import * as raw from './raw'
+
+const getBaseUrl = (params: GetSpaceEnvironmentParams) =>
+  `/spaces/${params.spaceId}/environments/${params.environmentId}/workflow`
+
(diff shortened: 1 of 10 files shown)


lib/adapters/REST/endpoints/workflow.ts
```diff
@@ -0,0 +1,81 @@
...
+) => {
+  const data = copy(rawData)
+  return raw.post<WorkflowProps>(http, getBaseUrl(params), data, {
+    headers,
+  })
+}
+
+export const update: RestEndpoint<'Workflow', 'update'> = (
+  http: AxiosInstance,
+  params: GetWorkflowParams,
```
Chaoste, line 48: Even though it's the same, that makes more sense :)

```suggestion
  params: UpdateWorkflowParams,
```

lib/entities/workflow.ts
```diff
@@ -0,0 +1,132 @@
...
+} from '../common-types'
+import { wrapCollection } from '../common-utils'
+import enhanceWithMethods from '../enhance-with-methods'
+
+export type WorkflowSysProps = Pick<
+  BasicMetaSysProps,
+  'id' | 'version' | 'createdAt' | 'createdBy' | 'updatedAt' | 'updatedBy'
+> & {
+  type: 'Workflow'
+  space: SysLink
```
Chaoste, line 21: I'm wondering when do we use `SysLink` and when do we use an explicit type with `Link<T>`? I see that we do this in other places as well. Is there some pattern to it?

Chaoste (approved):
LGTM!
When bumping the CMA version in UI, also remember to pass this entity to the app sdk. [Here's how](https://github.com/contentful/experience-packages/pull/1417/files#diff-fb63c16adab012c7dc56b1b18581dd55e6ce72ddc5075b06edac76f2a257de81) we did it for the WorkflowDefinition.

lib/entities/workflow.ts
ronaldronson: Sys link doesn't enforce the type, as the Link does, in cases where it's less obvious by the key (like space) I think it makes more sense to be more restrictive and allow only a specific type.

PR #1188, opened 2022-01-31
feat: option to accept marketplace terms [EXT-3424]

### LINKED ISSUE #1143: Install app via SDK (accept EULA through passing headers)
Hi folks,

I came across the fact that the SDK doesn't support passing headers to the method `createAppInstallation()`.
Passing headers is necessary to accept the Marketplace EULA:
`'X-Contentful-Marketplace': 'i-accept-end-user-license-agreement,i-accept-marketplace-terms-of-service,i-accept-privacy-policy'`

After consulting the Contentful-Community Slack, @luizfonseca suggested calling this [endpoint](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/app-installations/app-installation/install-or-update-an-app/console/curl) directly and passing the above-mentioned header. This, I ended up doing and it works perfectly.

Nevertheless, it would be awesome if the method supported either passing headers, or accepting the EULA directly.

andipaetzold:
## Summary

Adds `acceptAllTerms` option to `createAppInstallation`

## Description

It is currently not possible to install marketplace apps using `createAppinstallation` because we need to pass the `X-Contentful-Marketplace` header.

## Open questions

* Should there be one flag to approve all terms?
* Do we want to pass that param as an object or do we assume this will stay the only option?

## Motivation and Context

closes #1143


kdamball:
> Should there be one flag to approve all terms?

I'd say the number of flags should equal the number of clicks/checkboxes required on marketplace.

kdamball:
> Do we want to pass that param as an object or do we assume this will stay the only option?

These have barely changed for 2yrs, so it's ok imo

kdamball (commented):
Do we have a meaningful error back in case they dont pass the headers?

andipaetzold:
> Do we have a meaningful error back in case they dont pass the headers?

In CMA.js we do not specify any error messages. We just pass along what is returned from the CMA

kdamball (approved):

fabe (approved):

PR #1189, opened 2022-01-31
feat: [EXT-3440] add app actions

Jwhiles:

When adding a new method:


shikaan (approved):
Looks good to me. Is the call part a followup?

Jwhiles:
> Looks good to me. Is the call part a followup?

sure is :)

PR #1206, opened 2022-02-08
feat: introduce WorkflowsChangelog & WorkflowsChangelogEntry [HOMER-567]

Chaoste:
## Summary

Bring our new WorkflowsChangelog to CMA to use it in our frontend app.

## Description

We're right now implementing the API `/.../:environmentId/workflows_changelog/` which is a pure read endpoint with `getMany` being the only method. It allows paginating, ordering, and filtering by `workflow`/ `workflowDefinition`/ `entity`.

When adding a new method:


lib/entities/workflows-changelog-entry.ts
```diff
@@ -0,0 +1,63 @@
...
+  DefaultElements,
+  Link,
+  MakeRequest,
+  PaginationQueryOptions,
+  SysLink,
+} from '../common-types'
+import { wrapCollection } from '../common-utils'
+import enhanceWithMethods from '../enhance-with-methods'
+
+export type WorkflowsChangelogQueryOptions = Omit<PaginationQueryOptions, 'order'> & {
```
giotiskl, line 13: This has unfortunately changed a bit after the latest discussion I had with @aromanko. For now we agreed on this final shape:

```
export type WorkflowsChangelogQueryOptions = Omit<PaginationQueryOptions, 'order'> & {
  'entity.sys.linkType': string
  'entity.sys.id': string
  'workflow.sys.id': string
}
```

i.e. entityId, entityType and workflowId all required params for this first iteration.

giotiskl (changes requested):
PR looks good overall just a change needed for the params shape

lib/entities/workflows-changelog-entry.ts
Chaoste: You mean @ronaldronson ? :P

giotiskl: > You mean @ronaldronson ? :P

damn yes 😆 always tricky to find him

Chaoste: https://media1.giphy.com/media/l0Iyl55kTeh71nTXy/giphy.gif?cid=ecf05e47k7d9a2d1sdktvhknh9gbwbbwm25ymlh07zzwrxlf&rid=giphy.gif&ct=g

giotiskl (approved):

PR #1211, opened 2022-02-09
chore(depedencies): update axios

marcolink:
**Description:**

This update includes a version bump ([0.21.x - 025.0](https://github.com/axios/axios/releases)) for direct dependency axios which introduced a breaking change.


andipaetzold (commented):
This should be a `fix:` instead of a `chore:` for two reasons:
- We are changing the dependencies of the package.
- It is a potential breaking change for typescript users

This might even break for some non-typescript users as this is a breaking change in a direct dependency.

Code change looks fine

ruderngespra (commented):
As already discussed in slack, I'd agree as well with the comment, other than that lgtm! Can't reproduce the failing tests locally, but it may be worth double checking if we are doing this mayor bump.

PR #1215, opened 2022-02-10
feat: add support for passing headers in `.createScheduledAction`

luizfonseca:
## Summary

- Adds support to passing headers to createScheduledAction sdk function.

When adding a new method:


Spring3 (commented):
it would be pretty cool to have a test case covering it for both - space client and plain client

Spring3 (approved):

ChidinmaOrajiaku:
Do we still need the PR now that we've taken out the alpha headers?

luizfonseca:
@ChidinmaOrajiaku nope, all good now

PR #1249, opened 2022-03-16
fix: Allow raw patch requests

FaHeymann:
## Summary

At the moment using `cma.raw.patch` fails with `Unknown endpoint` thrown in https://github.com/contentful/contentful-management.js/blob/master/lib/adapters/REST/rest-adapter.ts#L69-L71 It looks to me like the `patch` method is just missing in the Http endpoint module.

## Description

Add a patch method to the Http endpoint module.

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/http.ts
+++ b/lib/adapters/REST/endpoints/http.ts
@@ -26,6 +26,14 @@ export const put: RestEndpoint<'Http', 'put'> = <T = any>(
   return raw.put<T>(http, url, payload, config)
 }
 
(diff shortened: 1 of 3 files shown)


thomasspiesser:
Do we know how we communicate with the entires patch EP currently?

FaHeymann:
Entry has a dedicated one, so there we don't have to use the `raw` Entity type https://github.com/contentful/contentful-management.js/blob/master/lib/adapters/REST/endpoints/entry.ts#L46-L64

andipaetzold (approved):

PR #1262, opened 2022-03-24
Add task.getManyForEntry

dmsolutionsmn:
## Summary

Add task.getManyForEntry

## Description

This adds the task.getManyForEntry method.  The `getAll` is a bit of a misnomer as it does not support paging and does not in fact get "ALL".

## Motivation and Context

When an entry has over 100 tasks (even resolved tasks) you can not get them all without using the `raw` endpoint.  The raw endpoint is not allowed from a plugin app, thus making it hard to get all the tasks for an entry.

- [X - integration couldn't be run locally] Both unit and integration tests are passing

When adding a new method:

- [- skipped as no tests were written for other task endpoints] Added a unit test for the new method
- [- skpped as no existing getAll etc] Added an integration test for the new method
- [X] The new method is added to the documentation


Chaoste (approved):
Changes look good to me :)

Chaoste:
Hey, I just read your Pr description again and saw your feedback on `task.getAll`. You're right that this is not formerly correct and we usually name it `getMany`. Should we rename it? wdyt?

Chaoste:
Some stuff was missing and the pipeline didn't work for some reason.

PR #1267, opened 2022-03-29
feat: rename getAll to getMany for tasks and comments

Chaoste:
## Summary

- Add task.getManyForEntry
- feat: rename to getManyForEntry for consistency and allow passing pagination parameters

BREAKING CHANGE: Rename `getAll` to `getManyForEntry` for tasks and comments in plain client.

## Description

Due to pipeline issues in https://github.com/contentful/contentful-management.js/pull/1262 and some missing fixes & improvements I cherry picked the commit by @dmsolutionsmn (for keeping authorship) and added the refactoring & fix.

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/comment.ts
+++ b/lib/adapters/REST/endpoints/comment.ts
@@ -27,7 +27,7 @@ export const get: RestEndpoint<'Comment', 'get'> = (
   params: GetCommentParams
 ) => raw.get<CommentProps>(http, getCommentUrl(params))
 
-export const getManyForEntry: RestEndpoint<'Comment', 'getManyForEntry'> = (
+export const getMany: RestEndpoint<'Comment', 'getMany'> = (
   http: AxiosInstance,
   params: GetEntryParams & QueryParams
 ) =>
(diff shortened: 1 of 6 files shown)


giotiskl (approved):

andipaetzold (changes requested):
Changing the name of the action from `getAll` to `getManyForEntry` is a breaking change and would break all existing apps using that method via the cma adapter

marcolink:
I suggest adding a deprecation warning for `getAll` and introducing `getManyForEntry` additionally for the reasons @andipaetzold mentioned.

marcolink:
Sorry, I might have not read things carefully enough.

Chaoste:
Mmmh, so do we roll out changes to methods in CMA only with new major versions? Totally makes sense to me but I wasn't aware of it so far 😅 
This thing is not urgent because it's only about the clean-up in CMA, it doesn't really add functionality. So how about this is on hold until the next major version? That being said, is there a release plan?

andipaetzold (approved):

PR #1274, opened 2022-04-06
feat: added new permission, actor, and action types to sdk [HOMER-684]

mdmjg:
## Summary

Added SDK types for WorkflowStepPermission, WorkflowStepPermissionActors, WorkflowStepPermissionTypes, and WorkflowStepPermissionEffect


lib/entities/workflow-definition.ts
```diff
@@ -14,6 +14,33 @@ import { wrapCollection } from '../common-utils'
 import enhanceWithMethods from '../enhance-with-methods'
 import { ActionType } from './role'
 
+/* Workflow Step Permission */
+
+export type WorkflowStepPermissionActors = 'all' | Link<'User'>[] | Link<'Team'>[]
```
Chaoste, line 19:
```suggestion
export type WorkflowStepPermissionActors = 'all' | Array<Link<'User'> | Link<'Team'>>
```
Currently, it would only allow a list of users or a list of teams. I assume we would also accept a mixed list, won't we?

Chaoste: It would be even more correct to say we require at least one entry in that list.

```suggestion
type NonEmptyArray<T> = [T, ...T[]];
export type WorkflowStepPermissionActors = 'all' | NonEmptyArray<Link<'User'> | Link<'Team'>>
```

Copied from [this SO post](https://stackoverflow.com/a/56006703/17269164).

Chaoste (changes requested):
looks good except for one comment

Chaoste (approved):
LGTM!

PR #1279, opened 2022-04-11
feat(release): update release query options type with new query params [SPA-538]

luizfonseca:
## Summary

- Adds more type hints for the possible query options when using `environment.getReleases` (Client) or `releases.getMany` (PlainClient)

## Description

- More type hints and explanations for the filters available when querying releases.

When adding a new method:


mshaaban0 (approved):
💅🏼 💅🏼

PR #1280, opened 2022-04-11
fix(tests): remove .only from release test []

luizfonseca:
## Summary

- A `.only` flag ended up being committed into `master`, this PR removes the flag.

When adding a new method:


mshaaban0 (approved):
💅

PR #1285, opened 2022-04-12
refactor: [] use new shape of app actions

shikaan:

When adding a new method:


andipaetzold (approved):
Keep in mind that `refactor:` won't trigger a new version

PR #1299, opened 2022-04-19
chore!: bump min node required to v14

Spring3:
Resolves #611 

## Summary

Will be a breaking change hence is marked as a major release

* discontinue builds of legacy bundles to move on from IE
* bump the min node version to 16 which is lts at the time of writing
* remove `@deprecated` `maxDepth` param for references api as it was substituted with `include` to match the actual query param sent to the server

## Motivation and Context

Even microsoft moved away from supporting IE, so I think it is time to sunset the legacy bundle. I don't think someone is testing that it works here still either.

## Checklist (check all before merging)

- [x] Both unit and integration tests are passing
- [ ] There are no breaking changes
- [x] Changes are reflected in the documentation

When adding a new method:

- [x] The new method is exported through the default and plain CMA client
- [x] All new public types are exported from `./lib/export-types.ts`
- [x] Added a unit test for the new method
- [x] Added an integration test for the new method
- [x] The new method is added to the documentation

PR #1300, opened 2022-04-19
feat: add support for release actions query endpoint [SPA-707]

luizfonseca:
## Summary

Adds support for the new `GET /release_actions` endpoint, that is replacing the existing `GET /releases/:id/actions`
This new endpoint supports querying for multiple releases, actions or specific properties.

## Description

- Change the usage of `/releases/:id/actions` to use the new `/release_actions` endpoint
- Add deprecation notice for passing `releaseId` instead of `query { 'sys.release.sys.id[in]`: ... } `
- Add specs covering new query params

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/release-action.ts
+++ b/lib/adapters/REST/endpoints/release-action.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { AxiosInstance } from 'contentful-sdk-core'
-import { GetReleaseParams } from '../../../common-types'
+import { GetReleaseParams, GetSpaceEnvironmentParams } from '../../../common-types'
 import { ReleaseActionQueryOptions } from '../../../entities/release-action'
 import { RestEndpoint } from '../types'
 import * as raw from './raw'
@@ -15,15 +15,32 @@ export const get: RestEndpoint<'ReleaseAction', 'get'> = (
   )
 }
 
+export const getMany: RestEndpoint<'ReleaseAction', 'getMany'> = (
+  http: AxiosInstance,
+  params: GetSpaceEnvironmentParams & { query?: ReleaseActionQueryOptions }
+) => {
+  return raw.get(
+    http,
+    `/spaces/${params.spaceId}/environments/${params.environmentId}/release_actions`,
+    {
+      params: params.query,
+    }
+  )
+}
+
+/** @deprecated use getMany with `sys.release.sys.id[in]` */
 export const queryForRelease: RestEndpoint<'ReleaseAction', 'queryForRelease'> = (
   http: AxiosInstance,
   params: GetReleaseParams & { query?: ReleaseActionQueryOptions }
 ) => {
   return raw.get(
     http,
-    `/spaces/${params.spaceId}/environments/${params.environmentId}/releases/${params.releaseId}/actions`,
+    `/spaces/${params.spaceId}/environments/${params.environmentId}/release_actions`,
     {
-      params: params.query,
+      params: {
+        'sys.release.sys.id[in]': params.releaseId,
+        ...params.query,
+      },
     }
   )
 }
--- a/lib/adapters/REST/endpoints/release-action.ts
+++ b/lib/adapters/REST/endpoints/release-action.ts
@@ -28,7 +28,6 @@ export const getMany: RestEndpoint<'ReleaseAction', 'getMany'> = (
   )
 }
 
-/** @deprecated use getMany with `sys.release.sys.id[in]` */
 export const queryForRelease: RestEndpoint<'ReleaseAction', 'queryForRelease'> = (
   http: AxiosInstance,
   params: GetReleaseParams & { query?: ReleaseActionQueryOptions }
--- a/lib/common-types.ts
+++ b/lib/common-types.ts
@@ -431,6 +431,7 @@ type MRInternal<UA extends boolean> = {
   (opts: MROpts<'Release', 'validate', UA>): MRReturn<'Release', 'validate'>
 
   (opts: MROpts<'ReleaseAction', 'get', UA>): MRReturn<'ReleaseAction', 'get'>
+  (opts: MROpts<'ReleaseAction', 'getMany', UA>): MRReturn<'ReleaseAction', 'getMany'>
   (opts: MROpts<'ReleaseAction', 'queryForRelease', UA>): MRReturn<
     'ReleaseAction',
     'queryForRelease'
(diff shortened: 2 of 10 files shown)


lib/common-types.ts
```diff
@@ -1124,6 +1125,10 @@ export type MRActions = {
       params: GetReleaseParams & { actionId: string }
       return: ReleaseAction
     }
+    getMany: {
+      params: GetSpaceEnvironmentParams & { query: ReleaseActionQueryOptions }
+      return: Collection<ReleaseAction, ReleaseActionProps>
```
Spring3, line 1130: I find it strange that here we use type `Collection` while everywhere else we use `CollectionProp`. Should we align? (relates to other functions of releases as well)

Spring3: Here's [an open PR](https://github.com/contentful/contentful-management.js/pull/1299) that will be the next breaking change in case you maybe want to change it for in the major version

lib/create-environment-api.ts
```diff
@@ -2099,20 +2099,28 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
       releaseId,
       query,
     }: {
-      releaseId: string
+      /** @deprecated use the query options object with `sys.release.sys.id[in]` */
+      releaseId?: string
```
Spring3, line 2103: you can also remove this in the next major version

lib/common-types.ts
luizfonseca: Interestingly, `CollectionProp` is the "pure" collection object, `Collection` is what it returns here since we wrap it with `wrapCollection` which runs the `toPlainObject()` function. The others might actually be incorrect :|

luizfonseca: I might actually just use `CollectionProp` since every other endpoint uses it like such 


(source of Collection: https://github.com/contentful/contentful-management.js/blob/79c4e518b8fb3acd22d783db7f98a3284e05271f/lib/common-utils.ts#L15-L16 - here we add a new prop `toPlainObject` that `CollectionProp` doesn't have)

Spring3: I don't see any line of code in this PR calling wrapCollection() method

Spring3: oh I see. That is for env api, not for plain client

luizfonseca: `plainClient` uses a different file though: https://github.com/contentful/contentful-management.js/blob/740dc15534a672369af4116d9407b4210d26b73e/lib/plain/common-types.ts#L558 (which uses `CollectionProp` correctly)

Spring3: worth discussing with the rest of the teams about it, but looks like you might be right

test/integration/release-action-integration.ts
```diff
@@ -1,46 +1,87 @@
...
   let testEnvironment: Environment
   let testRelease: Release
+  let testRelease2: Release
+
+  let testReleaseAction: ReleaseAction
+  let testReleaseAction2: ReleaseAction
 
   before(async () => {
-    testSpace = await getDefaultSpace()
+    testSpace = (await createTestSpace(initClient(), 'Release Actions')) as unknown as Space
```
Spring3, line 21: this will actually spin up the new space, are you aware of that?

test/integration/release-action-integration.ts
```diff
@@ -1,46 +1,87 @@
...
       title: 'Release (test)',
       entities: {
         sys: { type: 'Array' },
         items: [makeLink('Entry', TestDefaults.entry.testEntryReleasesId)],
       },
-    })
-  })
+    }
+
+    ;[testRelease, testRelease2] = await Promise.all([
```
Spring3, line 55: sorry, what is that `;`? I think I asked this question before, but this looks like some hole in javascript

Spring3: Why use it instead of const or let?

test/integration/release-action-integration.ts
luizfonseca: The team responsible for the project has setup an `autoClean` feature whenever we use `createTestSpace` -- there's also no space limits for the org these tests run on. They are cleaned up after a certain "age" that the space reaches.

It's a pretty good approach since now each test runs in isolation from entities created in other tests.

luizfonseca: I also refactored `releases-integration` to use the same approach in a previous PR.

test/integration/release-action-integration.ts
luizfonseca: Scope mostly, I want these two variables to be available in the test (entirely), so I initialised them before:

```ts
let testRelease: Release
let testRelease2: Release


// now I can assign/use these variables
```

Your other question: using `;` is to indicate an intention of "this is a new line of code". 
Using const/let: I would have to initialise them here, but they wouldn't be available throughout the test.


This line of code is doing:

```ts
testRelease = await ...
testRelease2 = await ...
```

([stack overflow](https://stackoverflow.com/questions/58069078/semicolon-before-square-bracket))

Spring3: Ok, but does it look like a good practice to follow this approach? I feel like that would be mentioned in the "Bad parts" section of the book about Javascript

luizfonseca: not a bad practice AFAIK

test/integration/release-action-integration.ts
```diff
@@ -1,46 +1,87 @@
...
+      await testEnvironment.createRelease(releasePayload),
+      await testEnvironment.createRelease(releasePayload),
```
Spring3, line 57: afaik there is no need for awaits inside of the array

luizfonseca: oh good catch 👍🏽

Spring3 (approved):
Thanks for the change. Could you add that point about the proper type (Collection vs CollectionProp) to the next FE agenda to have it discussed?

Spring3:
🚢

lib/adapters/REST/endpoints/release-action.ts
```diff
@@ -15,15 +15,32 @@ export const get: RestEndpoint<'ReleaseAction', 'get'> = (
...
+  return raw.get(
+    http,
+    `/spaces/${params.spaceId}/environments/${params.environmentId}/release_actions`,
+    {
+      params: params.query,
+    }
+  )
+}
+
+/** @deprecated use getMany with `sys.release.sys.id[in]` */
```
Spring3, line 31: should we also remove this?

PR #1311, opened 2022-04-22
fix(ActionCategories): fix app action category types [EXT-3722]

kdamball:
## Summary

Updating Custom app action types to match API

When adding a new method:

### DIFF
--- a/lib/entities/app-action.ts
+++ b/lib/entities/app-action.ts
@@ -19,56 +19,64 @@ export enum AppActionCategory {
   Custom = 'Custom',
 }
 
-export type CreateAppActionProps = {
+type BuiltInCategoriesProps = {
+  /**
+   * Category identifying the shape of the action.
+   */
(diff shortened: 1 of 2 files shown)


lib/entities/app-action.ts
```diff
@@ -19,56 +19,65 @@ export enum AppActionCategory {
...
+   */
+  category: AppActionCategory.Custom
+  parameters: AppActionParameterDefinition[]
+}
+
+type AppActionCategoryProps = BuiltInCategoriesProps | CustomAppActionProps
+
+export type CreateAppActionProps = AppActionCategoryProps & {
   url: string
   category: AppActionCategory
```
andipaetzold, line 41: That needs to be removed as it's part of `AppActionCategoryProps`

andipaetzold (changes requested):

andipaetzold (approved):

PR #1314, opened 2022-04-25
fix(tests): add wait for environment creation function to prevent flaky tests []

luizfonseca:
## Summary

- adds `waitForEnvironmentToBeReady` to most tests that are using `createEnvironment`

When adding a new method:


Spring3:
@luizfonseca could you try using [the next major test utils package](https://github.com/contentful/integration-test-utils/releases/tag/v2.0.0)? Maybe it will fix it?

Spring3:
@luizfonseca should we close it? looks like there's not time to follow up?

PR #1315, opened 2022-04-25
fix: add browser field to package.json

andipaetzold:
## Summary

Add `browser` field to `package.json`.

Any reason why this wasn't done before?

## Motivation and Context

When using contentful-management in browser context without esm support, the node bundle is used. This bundle requires native node libraries that are not available in the browser. For that reason you cannot use CMA.js in stackblitz: https://stackblitz.com/edit/js-u7mabv

By adding the `browser` field, the correct bundle (should) be selected and stackblitz should work correctly.

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#browser


marcolink (approved):
🚢

PR #1329, opened 2022-04-29
fix(ApiAppActionCategories):  export the actual type [EXT-3618]

kdamball:
## Summary

forgot to expose the new type in #1328  🤦🏾 

When adding a new method:

### DIFF
--- a/lib/entities/app-action.ts
+++ b/lib/entities/app-action.ts
@@ -13,15 +13,15 @@ type AppActionSys = Except<BasicMetaSysProps, 'version'> & {
 
 export type AppActionParameterDefinition = Omit<ParameterDefinition, 'labels'>
 
-export enum AppActionCategory {
+export enum AppActionCategoryType {
   EntryListV1Beta = 'EntryList.v1.0-beta',
   NotificationV1Beta = 'Notification.v1.0-beta',
   Custom = 'Custom',
 }
 
-export type ApiAppActionCategory = {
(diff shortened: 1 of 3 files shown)


shikaan (approved):

andipaetzold:
The naming `Api....` is not really consistent with all other exported entity types. Can't we call this `AppActionCategory`? Yes, that's the name of the enum, but I think we should try to stay consistent

andipaetzold:
🤦  It should be `AppActionCategoryProps` for consistency. I messed up myself in the comment.

kdamball:
@andipaetzold considering it contains `sys`, I wouldn't call this props but rather the entity name itself.

Spring3 (approved):

andipaetzold:
@kdamball In CMA.js the `...Props` interfaces always represent the entity as a plain object wthout any methods. The "non-props" interfaces always include methods like `toPlainObject` or similar and are used for the nested CMA client

E.g. AppDefinition:
* [AppDefinitionProps](https://github.com/contentful/contentful-management.js/blob/18b12c912b412ad89db1c77cb9bec0bf3116002c/lib/entities/app-definition.ts#L40-L70)
* [`AppDefinition` and method creating `AppDefinition`](https://github.com/contentful/contentful-management.js/blob/18b12c912b412ad89db1c77cb9bec0bf3116002c/lib/entities/app-definition.ts#L77-L97)

PR #1338, opened 2022-05-02
feat: added order parameter [HOMER-793]

mdmjg:
## Summary

- Added `order` query parameter to `workflows` entity and set the default to `-sys.createdAt`

When adding a new method:


lib/entities/workflow.ts
```diff
@@ -61,6 +61,11 @@ export type WorkflowQueryOptions = Omit<PaginationQueryOptions, 'order'> & {
   /** Find workflows containing the specified, comma-separated entities. Requires `sys.entity.sys.linkType` */
   'sys.entity.sys.id[in]'?: string
   'sys.workflowDefinition.sys.id'?: string
+  /**
+   * Order releases by
+   * @default -sys.createdAt
+   * */
+  order?: string
```
ronaldronson, line 68: let's add the enum here with all available options

ronaldronson (changes requested):

lib/entities/workflow.ts
marcolink: For public-facing APIs, where the value can be a string value that can be passed through from other places, I suggest using a string union over an enum. 

This suggestion might be due to my lack of TS knowledge though 😁 


My understanding is, that whenever you have a property that can be dynamically set, you don't want to enforce an enum transformation.

mdmjg: so you mean that I should have something like: 
`type OrderQueryParam = 'sys.updatedAt' | '-sys.updatedAt' | 'sys.createdAt' | '-sys.createdAt'
`order: OrderQueryParam` 
instead of an `enum`?

marcolink: Yes!

lib/entities/workflow.ts
```diff
@@ -61,6 +63,11 @@ export type WorkflowQueryOptions = Omit<PaginationQueryOptions, 'order'> & {
   /** Find workflows containing the specified, comma-separated entities. Requires `sys.entity.sys.linkType` */
   'sys.entity.sys.id[in]'?: string
   'sys.workflowDefinition.sys.id'?: string
+  /**
+   * Order releases by
```
ronaldronson, line 67: Now I know where it was copied from)

ronaldronson (commented):
One last thing

ronaldronson (approved):

PR #1342, opened 2022-05-05
feat: introduce org app installations

marcolink:
## Description

Moving the App Definition API to the top level of the client. Adding a method that allows users to get all installations for a given definition within a given Contentful organization.

When adding a new method:


lib/adapters/REST/endpoints/app-definition.ts
```diff
@@ -66,3 +86,28 @@ export const del: RestEndpoint<'AppDefinition', 'delete'> = (
...
 }
+
+export const getInstallationsForOrg: RestEndpoint<'AppDefinition', 'getInstallationsForOrg'> = (
+  http: AxiosInstance,
+  params: GetAppInstallationsForOrgParams & PaginationQueryParams
+) => {
+  if (!params.appDefinitionId) {
+    throw new ValidationError(
+      'appDefinitionId',
+      'Please provide argument of { appDefinitionId: string, organizationId: string } for the getInstallationsForOrg method'
```
shikaan, line 97: The two error messages seem inconsistent. I would use the same or group the error into one clause

```
if (!params.appDefinitionId || !params.organizationId) {
...
}
```

Also, while I think having explicit errors makes for a better experience in general, on the other methods we don't do validation. If we want to go down this path, I would do it at least on all the methods in app definitions.

With this change of approach however I would recommend involving @contentful/team-dev-workflows to see if they have opinions or objections

lib/create-contentful-api.ts
```diff
@@ -180,6 +183,18 @@ export default function createClientApi(makeRequest: MakeRequest) {
       }).then((data) => wrapUser<T>(makeRequest, data))
     },
 
+    //TODO add usage example
```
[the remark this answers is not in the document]
martin3walker: Needs to be addressed. Good catch 👍

test/integration/app-definition-integration.js
```diff
@@ -95,4 +100,74 @@ describe('AppDefinition api', function () {
...
   })
+
+  test('top level getAppDefinition', async () => {
+    const { orgId, appId } = await createAppDefinition()
+    const appDefinition = await getAppDefinition({ organizationId: orgId, appDefinitionId: appId })
+
+    expect(appDefinition.sys.organization.sys.id).equals(orgId)
+    expect(appDefinition.sys.id).equals(appId)
+
+    await appDefinition.delete()
```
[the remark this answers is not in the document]
martin3walker: Actually already an after function handling exactly this. Again good catch :)

test/unit/mocks/entities.js
```diff
@@ -534,6 +534,22 @@ const appInstallationMock = {
   }),
 }
 
+const appInstallationsForOrgMock = {
```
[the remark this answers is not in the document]
martin3walker: Good call, i'll add the `Environment` array

test/integration/app-definition-integration.js
```diff
@@ -95,4 +100,74 @@ describe('AppDefinition api', function () {
...
+    const { orgId, appId } = await createAppDefinition()
+    const appDefinition = await getAppDefinition({ organizationId: orgId, appDefinitionId: appId })
+
+    try {
+      await appDefinition.getInstallationsForOrg({
+        appDefinitionId: appId,
+      })
+    } catch (e) {
+      expect(e.toString()).to.equal(
+        'ValidationError: Invalid "organizationId" provided, Please provide an object with the shape { appDefinitionId: string, organizationId: string } as argument when calling the getInstallationsForOrg method'
```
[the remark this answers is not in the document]
martin3walker: Okey, I'll change this if we keep the errors 👍

lib/adapters/REST/endpoints/app-definition.ts
martin3walker: @marcolink any thoughts on this? Removing them causes an error to bubble up from the extensibility api, which is way less specific, and probably confusing for users. So, I'm in favor of keeping them, and happy to add error handling to the entire App Definition API

martin3walker: @shikaan Marco and I had a chat, and we decided it's more consistent to improve the error message in the Extensibility API. Apropros errors, do we have any sort of overall strategy for error handling?

martin3walker: If we make the orgId fall back to an empty string when it is undefined, it bubbles up a decent error from the extensibility API. It's a bad request, which is correct, and it mentions the organizationId. I think that's probably the best outcome we can get if we don't handle the errors directly in this repo.

test/unit/mocks/entities.js
```diff
@@ -534,6 +534,25 @@ const appInstallationMock = {
...
+  }),
+  items: [
+    cloneDeep(appInstallationMock),
+    {
+      space: {
+        sys: {
+          id: 'aspace',
+        },
+      },
+    },
```
shikaan, line 549: nit: this still seems in the wrong place here

shikaan (approved):

PR #1349, opened 2022-05-13
[EXT-3730] Fix app definition params

martin3walker:
## Description

Get parameters for `installationsForOrg` method from App Definition response, to avoid making the user pass them twice.

When adding a new method:

### DIFF
--- a/lib/entities/workflow-definition.ts
+++ b/lib/entities/workflow-definition.ts
@@ -101,6 +101,7 @@ export type WorkflowDefinitionSysProps = Pick<
   type: 'WorkflowDefinition'
   space: SysLink
   environment: SysLink
+  isLocked: boolean
 }
 
 export type WorkflowDefinitionValidationLink = {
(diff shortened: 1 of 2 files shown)


lib/create-app-definition-api.ts
```diff
@@ -177,19 +177,14 @@ export default function createAppDefinitionApi(makeRequest: MakeRequest) {
...
-    }) {
+    getInstallationsForOrg() {
+      const raw = this.toPlainObject() as AppDefinitionProps
       return makeRequest({
         entityType: 'AppDefinition',
         action: 'getInstallationsForOrg',
         params: {
-          appDefinitionId,
-          organizationId,
+          appDefinitionId: raw.sys.id,
```
marcolink, line 186: out of curiosity, where is the appDefinitionId coming from when using in conjunction with the app-sdk? 
Does the adapter set it automatically?

marcolink (approved):

PR #1361, opened 2022-05-17
feat: [] add app action list for env

loweisz:
Adds the method for listing all app actions of an environment. 

url: `/spaces/:sid/environments/:env_id/actions`

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/app-action.ts
+++ b/lib/adapters/REST/endpoints/app-action.ts
@@ -3,11 +3,9 @@ import * as raw from './raw'
 import { normalizeSelect } from './utils'
 import {
   CollectionProp,
-  GetAppActionCallParams,
   GetAppActionParams,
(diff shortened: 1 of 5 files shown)


lib/adapters/REST/endpoints/app-action.ts
```diff
@@ -16,6 +17,9 @@ const getBaseUrl = (params: GetAppDefinitionParams) =>
 const getAppActionUrl = (params: GetAppActionParams) =>
   `${getBaseUrl(params)}/${params.appActionId}`
 
+const getAppActionsEnvUrl = (params: GetSpaceEnvironmentParams) =>
+  `/spaces/${params.spaceId}/environments/${params.environmentId}/actions`
```
kdamball, line 21: FYI, in case they dont pass `envId`, it will still work (defaults to `master`). I would make it optional

PR #1415, opened 2022-06-22
feat: add UIConfig and UserUIConfig [EXT-3738]

andipaetzold:
## Summary

Adds new entities UIConfig and UserUIConfig

Integration tests can only be added once the endpoints are available

## TODO

- [x] Remove page field

When adding a new method:


lib/create-ui-config-api.ts
```diff
@@ -0,0 +1,62 @@
...
+     * @example ```javascript
+     * const contentful = require('contentful-management')
+     *
+     * const client = contentful.createClient({
+     *   accessToken: '<content_management_api_key>'
+     * })
+     *
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment_id>'))
+     * .then((environment) => environment.getUIConfig('<entry_id>'))
```
loweisz, line 41: I don't think we need an entry-id here right?

lib/entities/user-ui-config.ts
```diff
@@ -0,0 +1,53 @@
...
+  title: string
+  order?: {
+    fieldId: string
+    direction: 'ascending' | 'descending'
+  }
+  displayedFieldIds?: string[]
+  contentTypeId: string | null
+  searchText?: string
+  searchFilters?: [string, string, string][]
+}
```
loweisz, line 39: Can we not share someof the types with the ui-config?

loweisz, line 39: Can we not share someof the types with the ui-config?

lib/create-ui-config-api.ts
loweisz, line 41: also here I think we don't the entry id

andipaetzold: Copy paste :D

lib/entities/user-ui-config.ts
andipaetzold: I explicitly kept the types separate in case they diverge from another.
E.g. the `View` has an additional `roles` field

loweisz: True, but we could share some parts and extend them for the use case. 
But I'm also fine with two different types 👍

loweisz (approved):
one typo, then its good to go

lib/entities/user-ui-config.ts
```diff
@@ -0,0 +1,54 @@
...
+
+interface View {
+  id: string
+  title: string
+  order?: {
+    fieldId: string
+    direction: 'ascending' | 'descending'
+  }
+  page?: number
```
kdamball, line 35: These have to go, no? Or was it the page property in UIConfig itself? @colomolo has a better idea which `page` property was deleted

kdamball: Yep, found similar comment from @colomolo  here: https://github.com/contentful/doc-app/pull/2379

andipaetzold: I have that in the TODO list in the PR description 🤓

kdamball: will wait for approval then

lib/entities/user-ui-config.ts
```diff
@@ -0,0 +1,53 @@
...
+
+interface View {
+  id: string
+  title: string
+  order?: {
+    fieldId: string
+    direction: 'ascending' | 'descending'
+  }
+  displayedFieldIds?: string[]
+  contentTypeId: string | null
```
kdamball, line 36: @colomolo can you confirm this is needed?

colomolo: yes, confirming.

kdamball (approved):

colomolo (approved):
lgtm :shipit:

PR #1439, opened 2022-07-06
feat: add flow type [HOMER-1038]

ronaldronson:
## Summary

Add new flowType property to the WorkflowDefinition entity.

When adding a new method:


Lelith (approved):
Looks good to me!

Issue #1419, opened 2022-06-27
Get published version of an Entry using Content Management Api


JuliaFerrandiz:
Hi,

I am building a Contentful App for a given contentful type and I need to check all the existing entries of this type for a validation. For that, I need the published version of the entries, no the Changed one.

In the EntryEditor.tsx, I have available the sdk (EditorExtensionSDK) and the cma (PlainClientAPI). To get all teh published entries, I could use ` sdk.space.getPublishedEntries({'content_type': "myContentType"})`, but I get the warning that it will be deprecated:
`since version 4.0.0 Consider using cma.entry.getMany instead See [https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-library] for more details`

If I try to use the CMA, I couldn't find a way to get the published version of the entries. The best I was able to do is:
```
cma.entry.getMany({
        query: {
          content_type: "myContentType",
          'sys.publishedVersion[exists]': true
        },
      })
```

Is there a way to do it?

Spring3:
Hey @JuliaFerrandiz . Please find the new function in the recent release (you can check the version in the PR linked in this thread)

PR #1472, opened 2022-08-30
feat:  Add support for GET published entries endpoint [SPA-817]

Inbal-Gordon:
##  Description

The purpose of this PR is to bring back the visibility of all published versions of the entries for CMA by adding support for GET published entries endpoint.

## Motivation and Context

A user asked this on [github](https://github.com/contentful/contentful-management.js/issues/1419#issue-1286007785) as once this functionality was available 

- [x] Changes are reflected in the documentation (will be on a separate PR)

When adding a new method:


Spring3 (changes requested):
Good start. I pointed to some of the mistakes made which caused the type errors

Also please update the name of the PR to better reflect what is being done. `chore: try to fix types` is a good name for a commit, but the PR does more than fixing types.

Same for the type of the branch. `feat` would be better because we are adding a new feature here. `chore` doesn't work becuase when you merge it, it will not bump the version of the package

lib/plain/plain-client.ts
```diff
@@ -160,6 +160,7 @@ export const createPlainClient = (
       delete: wrap(wrapParams, 'Task', 'delete'),
     },
     entry: {
+      getPublishedEntries: wrap(wrapParams, 'Entry', 'getPublishedEntries'),
```
Spring3, line 163: same here. Please rename it back to `getPublished`

lib/create-environment-api.ts
```diff
@@ -719,6 +719,39 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * .then((environment) => environment.getPublishedEntries{'content_type': 'foo'})) // you can add more queries as 'key': 'value'
+     * .then((response) => console.log(response.items))
+     * .catch(console.error)
+     * ```
+     */
+    getPublishedEntries(query: QueryOptions = {}) {
+      const raw = this.toPlainObject() as EnvironmentProps
+      return makeRequest({
+        entityType: 'Entry',
+        action: 'getMany',
```
Spring3, line 746: Here it is calling the function called `getMany` while we need to call the function called `getPublished` which you have added before

lib/adapters/REST/endpoints/entry.ts
```diff
@@ -30,6 +30,22 @@ export const get: RestEndpoint<'Entry', 'get'> = <T extends KeyValueMap = KeyVal
   )
 }
 
+export const getPublishedEntries:RestEndpoint<'Entry', 'getPublishedEntries'> = <T extends KeyValueMap = KeyValueMap>(
```
Spring3, line 33: here we can keep the old name - `getPublished` because users will call it as `entry.getPublished` and it will already be clear that they are requesting entries

with the current name they would call it as `entry.getPublishedEntries` so the word `entry` appears twice and it is unnecessary

Spring3 (dismissed):

lib/common-types.ts
```diff
@@ -599,6 +600,7 @@ export interface Adapter {
  */
 export type MRActions = {
   Http: {
+    queryPublished: { params: { url: string; config?: AxiosRequestConfig }; return: any }
```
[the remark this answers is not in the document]
Spring3: [Here I updated the pointer to the correct section](https://github.com/contentful/contentful-management.js/blob/2264d1e39658442c6d1ea405f090d3f343514764/lib/common-types.ts#L946)

Spring3:
Also, don't forget to add the type for your new function [here](https://github.com/contentful/contentful-management.js/blob/2264d1e39658442c6d1ea405f090d3f343514764/lib/common-types.ts#L372) for `Entry`

test/helpers.js
```diff
@@ -73,8 +73,12 @@ export async function getTestUser() {
   return organization.getUser(userId)
 }
 
-export async function getDefaultSpace() {
+export async function getDefaultSpace(plainClient = false) {
   const { spaceId } = TestDefaults
+
+  if (plainClient) {
```
SofiaMargariti, line 79: Why was this change needed here?

SofiaMargariti:
I've noticed package-lock.json has a lot of changes in this PR. Was that on purpose? How come this PR resulted in so many changes if we didn't update any dependencies?

Inbal-Gordon:
@SofiaMargariti the reason for this is because I was mistakenly asked by Dan to remove my package-json file and use yarn however the original project on github uses package-json and npm not yarn

test/helpers.js
luizfonseca: to create a different client base on the default space, though @Inbal-Gordon we can probably improve it:

- We can undo the change on this `helpers.js`
- In `entry-integration` `line 413` we can use ` plainClient = initPlainClient({ spaceId: TestDefaults.spaceId })` instead. Make sure the imports are working.

test/integration/entry-integration.js
```diff
@@ -395,4 +401,20 @@ describe('Entry Api', () => {
...
     })
   })
+
+  describe('read plainClientApi', () => {
+    /**
+     * @type {import('../../lib/contentful-management').PlainClientAPI}
+     */
+    let plainClient
+    before(async () => {
+      plainClient = await getDefaultSpace(true)
```
luizfonseca, line 411:
```suggestion
       plainClient = initPlainClient({ spaceId: TestDefaults.spaceId })
```

so we don't need that change in `helpers.js`

luizfonseca:
@Inbal-Gordon  Code looks good, just some minor comments. Let us know once they are addressed.

SofiaMargariti:
> @SofiaMargariti the reason for this is because I was mistakenly asked by Dan to remove my package-json file and use yarn however the original project on github uses package-json and npm not yarn

I see, it looks like it's due to using a newer version of npm, but it looks like it is backwards compatible so it should be ok 👍

Inbal-Gordon:
@SofiaMargariti @luizfonseca I have updated the test and removed the change we did in the helpers.js file

luizfonseca (approved):

lib/adapters/REST/endpoints/entry.ts
```diff
@@ -30,6 +30,24 @@ export const get: RestEndpoint<'Entry', 'get'> = <T extends KeyValueMap = KeyVal
...
+  T extends KeyValueMap = KeyValueMap
+>(
+  http: AxiosInstance,
+  params: GetSpaceEnvironmentParams & QueryParams,
+  rawData?: unknown,
+  headers?: AxiosRequestHeaders
+) => {
+  return raw.get<CollectionProp<EntryProps<T>>>(
+    http,
+    `/spaces/${params.spaceId}/environments/${params.environmentId}/public/entries`,
```
Inbal-Gordon, line 43: @luizfonseca Is the path with `/public` correct here?

PR #1473, opened 2022-08-30
fix: move getForOrganization to AppInstallation entity [EXT-3877]

vikaskumr:
## Summary

Following the correct format and trying to move the endpoint for clarity.

## Description

Shift the getInstallationsForOrg function and change the action and entity if possible for the endpoint.
The endpoint should follow this pattern: cma.appInstallation.getForOrg()
Merge app is using this endpoint. Remove the old endpoint after changes in Merge app.
https://github.com/contentful/merge-app/blob/main/src/core/hooks/data/useFetchEnvironments.ts#L12

When adding a new method:

### DIFF
--- a/lib/plain/common-types.ts
+++ b/lib/plain/common-types.ts
@@ -663,15 +663,18 @@ export type PlainClientAPI = {
       headers?: AxiosRequestHeaders
     ): Promise<AppDefinitionProps>
     delete(params: OptionalDefaults<GetAppDefinitionParams>): Promise<any>
-    getInstallationsForOrg(
-      params: OptionalDefaults<GetAppDefinitionParams>
-    ): Promise<AppInstallationsForOrganizationProps>
+    // getInstallationsForOrg(
+    //   params: OptionalDefaults<GetAppDefinitionParams>
+    // ): Promise<AppInstallationsForOrganizationProps>
   }
   appInstallation: {
     get(params: OptionalDefaults<GetAppInstallationParams>): Promise<AppInstallationProps>
     getMany(
       params: OptionalDefaults<GetSpaceEnvironmentParams & PaginationQueryParams>
     ): Promise<CollectionProp<AppInstallationProps>>
+    getForOrganization(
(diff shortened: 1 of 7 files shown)


andipaetzold (changes requested):
This is not a `refactor:` but a `fix:`. `refactor` does not trigger a new version

andipaetzold (commented):
Rest looks fine

lib/plain/common-types.ts
```diff
@@ -672,6 +675,9 @@ export type PlainClientAPI = {
     getMany(
       params: OptionalDefaults<GetSpaceEnvironmentParams & PaginationQueryParams>
     ): Promise<CollectionProp<AppInstallationProps>>
+    getForOrganization(
+      params: OptionalDefaults<GetAppDefinitionParams>
```
colomolo, line 679: Shouldn't we have a type with a different name here? Meaning it mentions App Definition, but lives in App Installation.

vikaskumr: We can make but I think it will be same type like `GetAppDefinitionParams` just with  different name so I just reused it. 
@andipaetzold wdyt?

andipaetzold: No strong opinion here.
I think it is fine to keep things like they are as I doubt many people read these types. Also in the existing code, we often reuse types. But feel free to create an alias type with an easier to remember name.

colomolo: Alright, I have no strong knowledge of this codebase, so was just asking.

lib/plain/common-types.ts
```diff
@@ -672,6 +675,9 @@ export type PlainClientAPI = {
     getMany(
       params: OptionalDefaults<GetSpaceEnvironmentParams & PaginationQueryParams>
     ): Promise<CollectionProp<AppInstallationProps>>
+    getForOrganization(
```
colomolo, line 678: But shouldn't it be `getForOrg` as in the ticket?

andipaetzold (approved):

PR #1474, opened 2022-08-31
fix: add defaultLocale to SpaceProps

mshaaban0:
## Summary

`createSpace` params types are missing `defaultLocale` and misleading in TS environment

As per [docs](https://www.contentful.com/developers/docs/references/content-management-api/\#/reference/spaces/spaces-collection/create-a-space/console/js) our CMA API supports it

## Motivation and Context

See https://github.com/contentful/contentful-cli/pull/1583#discussion_r957541315

## Checklist

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -17,6 +17,7 @@ import { UsageQuery } from './entities/usage'
 import { UserProps } from './entities/user'
 
 export type ClientAPI = ReturnType<typeof createClientApi>
+type CreateSpaceProps = Omit<SpaceProps, 'sys'> & { defaultLocale?: string }
 
 /**
  * @private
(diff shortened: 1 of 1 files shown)


lib/entities/space.ts
```diff
@@ -8,6 +8,7 @@ import enhanceWithMethods from '../enhance-with-methods'
 export type SpaceProps = {
   sys: BasicMetaSysProps & { organization: { sys: { id: string } } }
   name: string
+  defaultLocale?: string
```
marcolink, line 11: Should we have an extended type for `CreateSpaceProps` or are we sure wherever SpaceProps is used, we can set a `defaultLocale` param?

mshaaban0: Extension is correct. It's also used in `getSpace` and others that doesn't return `defaultLocale`

marcolink (approved):
Sweeeeet 🎉

PR #1477, opened 2022-09-08
fix: categories renamed in app action category type  [EXT-3783]

vikaskumr:

When adding a new method:


shikaan (approved):
LGTM, but we need to make the change a `fix: ...` or we won't get the version bumped

PR #1479, opened 2022-09-12
feat: add urn to entity meta sys props

alicankargin:
## Summary

This PR is for adding urn as an optional parameter to EntityMetaSysProps.

## Motivation and Context

The type is used in field-editors repository. For this [PR](https://github.com/contentful/field-editors/compare/DANTE-648-urn-from-api-for-entries?expand=1) to get merged we need this change.

When adding a new method:

### DIFF
--- a/lib/entities/entry.ts
+++ b/lib/entities/entry.ts
@@ -35,6 +35,12 @@ export type EntryReferenceOptionsProps = {
 
 export interface Entry extends EntryProps, DefaultElements<EntryProps>, ContentfulEntryApi {}
 
+export type WithResourceName<T extends { sys: unknown }> = T extends { sys: infer Sys }
+  ? Omit<T, 'sys'> & {
+      sys: Sys & { urn: string }
+    }
+  : never
+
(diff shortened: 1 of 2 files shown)


anho (changes requested):
Let's find a way to have the property mandatory whenever we need it.

lib/entities/entry.ts
```diff
@@ -35,6 +35,8 @@ export type EntryReferenceOptionsProps = {
 
 export interface Entry extends EntryProps, DefaultElements<EntryProps>, ContentfulEntryApi {}
 
+export type WithResourceName<T extends EntryProps> = T & { sys: T['sys'] & { urn: string } }
```
anho, line 38: This version will give us even better type support:

```
export type WithResourceName<T extends { sys: unknown }> = T extends { sys: infer Sys } ? Omit<T, 'sys'> & {
  sys: Sys & { urn: string }
} : never;
```

anho (approved):

PR #1480, opened 2022-09-12
chore: improve test space deletion 

damienxy:
Improves the deletion of test spaces created during integration tests. 

Things done: 
- set lower threshold for space cleanup (10 minutes instead of 1 hour)
- add missing space cleanup to tests that create test spaces
- add a global `before`, as the `after` doesn't always run (e.g. when you stop the test run midway)

### DIFF
--- a/test/mocha-global.js
+++ b/test/mocha-global.js
@@ -4,8 +4,12 @@ import { cleanupTestSpaces } from './helpers'
 const housekeeping = async () => {
   try {
     await cleanupTestSpaces()
-  } catch {
-    // ignore if the cleanup fails. Usually fails locally due to missing credentials
+  } catch (err) {
+    if (err.message === 'Missing credential') {
+      console.log('Skipped deletion of old test spaces due to missing credentials.')
+    } else {
+      console.log('Skipped deletion of old test spaces. Error:', err.message)
+    }
   }
 }
 
--- a/test/mocha-global.js
+++ b/test/mocha-global.js
@@ -1,10 +1,19 @@
-import { after } from 'mocha'
+import { before, after } from 'mocha'
 import { cleanupTestSpaces } from './helpers'
 
-after('clean up test spaces', async () => {
+const housekeeping = async () => {
   try {
     await cleanupTestSpaces()
   } catch {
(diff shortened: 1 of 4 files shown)


test/mocha-global.js
```diff
@@ -1,10 +1,19 @@
-import { after } from 'mocha'
+import { before, after } from 'mocha'
 import { cleanupTestSpaces } from './helpers'
 
-after('clean up test spaces', async () => {
+const housekeeping = async () => {
   try {
     await cleanupTestSpaces()
   } catch {
     // ignore if the cleanup fails. Usually fails locally due to missing credentials
```
marcolink, line 8: Should we log the error? This might help users to learn how to fix it

andreascful: While I agree with your points Meike I'd still be in favor of logging the error. An error that is not logged (especially if it also CAN be something else than a credential error) creates the risk of significant confusion/effort when needing to debug a problem in the future.

damienxy: Okay, added some logging :-)

test/mocha-global.js
```diff
@@ -1,10 +1,23 @@
...
   try {
     await cleanupTestSpaces()
-  } catch {
-    // ignore if the cleanup fails. Usually fails locally due to missing credentials
+  } catch (err) {
+    if (err.message === 'Missing credential') {
+      console.log('Skipped deletion of old test spaces due to missing credentials.')
+    } else {
+      console.log('Skipped deletion of old test spaces. Error:', err.message)
+    }
```
luizfonseca, line 12: nitpick: you can probably get away with just 1 console.log tbh, you can let other people check out the error etc

damienxy: This is intentional, as we expect the credential error to happen most of the time, and as it can basically be ignored, we don't want to log the word "Error" in that case (see related discussion [here](https://github.com/contentful/contentful-management.js/pull/1480#discussion_r968598636)).

luizfonseca: hmm gotcha 👍🏽

luizfonseca (approved):
LGTM 👍🏽 
Nothing blocking from my side.

andreascful (approved):

PR #1482, opened 2022-09-15
feat: support xspace fields [Dante-708]

Cyberxon:
## Summary

This PR was created to support the new Cross Space fields added in the creation of content types and entries.

## Description

the new xspace field type should be recognized and handled in the JS SDKs

also when editing content types and entries with xspace fields

we do not implement xspace links resolution in the SDK, just return the link object/CRN as is

### DIFF
--- a/lib/entities/field-type.ts
+++ b/lib/entities/field-type.ts
@@ -13,5 +13,5 @@ export type FieldType =
   | { type: 'ResourceLink'; linkType: 'Contentful:entry' }
   | { type: 'Array'; items: { type: 'Symbol' } }
   | { type: 'Array'; items: { type: 'Link'; linkType: 'Entry' } }
-  | { type: 'Array'; items: { type: 'Resourcelink'; linkType: 'Contentful:entry' } }
+  | { type: 'Array'; items: { type: 'Resourcelink'; linkType: 'Contentful:Entry' } }
   | { type: 'Array'; items: { type: 'Link'; linkType: 'Asset' } }
--- a/lib/entities/field-type.ts
+++ b/lib/entities/field-type.ts
@@ -10,7 +10,7 @@ export type FieldType =
   | { type: 'Location' }
   | { type: 'Link'; linkType: 'Asset' }
   | { type: 'Link'; linkType: 'Entry' }
-  | { type: 'ResourceLink'; linkType: 'Contentful:entry' }
+  | { type: 'ResourceLink'; linkType: 'Contentful:Entry' }
   | { type: 'Array'; items: { type: 'Symbol' } }
   | { type: 'Array'; items: { type: 'Link'; linkType: 'Entry' } }
   | { type: 'Array'; items: { type: 'Resourcelink'; linkType: 'Contentful:Entry' } }
--- a/lib/entities/field-type.ts
+++ b/lib/entities/field-type.ts
@@ -10,6 +10,8 @@ export type FieldType =
   | { type: 'Location' }
   | { type: 'Link'; linkType: 'Asset' }
   | { type: 'Link'; linkType: 'Entry' }
+  | { type: 'ResourceLink'; linkType: 'Contentful:entry' }
   | { type: 'Array'; items: { type: 'Symbol' } }
   | { type: 'Array'; items: { type: 'Link'; linkType: 'Entry' } }
+  | { type: 'Array'; items: { type: 'Resourcelink'; linkType: 'Contentful:entry' } }
   | { type: 'Array'; items: { type: 'Link'; linkType: 'Asset' } }
--- a/test/integration/entry-integration.js
+++ b/test/integration/entry-integration.js
@@ -420,13 +420,23 @@ describe('Entry Api', () => {
     })
   })
 
+  /**
+   * This test was created to make sure the SDK supports
+   * creating/updating/publishing/upublishing/deleting of
+   * entries with cross space links
+   *
+   * This test is slightly fragile as it can break if the
+   * entry '4zimQzVMxDsSX607PCfo2u' from the space '6mqcevu5a50r' is deleted
+   */
   describe('write with xspace references', () => {
     let xSpaceEnabledSpace
     let xSpaceEnabledEnvironment
     let xSpaceDisabledSpace
     let xSpaceDisabledEnvironment
-    let contentTypeDisabledSpace
-    let contentTypeEnabledSpace
+    let xSpaceDisabledContentType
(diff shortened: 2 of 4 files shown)


test/integration/entry-integration.js
```diff
@@ -301,7 +301,7 @@ describe('Entry Api', () => {
     })
```
alicankargin, line 301: A test in this file is failing. `Gets entries by creation order and id order`. I don't know if it's related to your changes though. Rerun the tests maybe?

lib/entities/field-type.ts
```diff
@@ -10,6 +10,8 @@ export type FieldType =
   | { type: 'Location' }
   | { type: 'Link'; linkType: 'Asset' }
   | { type: 'Link'; linkType: 'Entry' }
+  | { type: 'ResourceLink'; linkType: 'Contentful:entry' }
```
luizfonseca, line 13: Q: I've seen it around being `Contentful:Entry` or is it case-insensitive - aka doesn't matter etc? 🤔

test/integration/entry-integration.js
```diff
@@ -419,4 +419,124 @@ describe('Entry Api', () => {
...
+      )
+      await contentTypeEnabledSpace.publish()
+    })
+
+    after(async () => {
+      await contentTypeDisabledSpace.unpublish().then((unpublishedContentType) =>
+        unpublishedContentType.delete().then(() => {
+          xSpaceDisabledSpace.delete()
+        })
+      )
```
luizfonseca, line 501: nitpick: we should probably stick to `async/await` instead of `then().then()`

luizfonseca: that object is actually "refreshed" every time you perform an action in it, so you could write it as

```
await contentTypeDisabledSpace.unpublish()
await contentTypeDisabledSpace.delete()
await xSpaceDisabledSpace.delete()
```

Cyberxon: I wrote it that way before but for some reason I wanted to change it. I'll go back to this.

test/integration/entry-integration.js
Cyberxon: yeah apperently I had to delete the entries and Content types that were created and not deleted in my failing tests so this one works correvtly.

lib/entities/field-type.ts
alicankargin, line 13: I think you forgot to convert this to `Contentful:Entry`

test/integration/entry-integration.js
```diff
@@ -419,4 +419,131 @@ describe('Entry Api', () => {
...
+    before(async () => {
+      // default space has xspace enabled in it through the feature flags
+      xSpaceEnabledSpace = await getDefaultSpace()
+      xSpaceEnabledEnvironment = await xSpaceEnabledSpace.getEnvironment('master')
+      xSpaceDisabledSpace = await createTestSpace(initClient(), 'Entry')
+      xSpaceDisabledEnvironment = await createTestEnvironment(
+        xSpaceDisabledSpace,
+        'Testing Environment'
+      )
+      contentTypeDisabledSpace = await xSpaceDisabledEnvironment.createContentTypeWithId(
```
alicankargin, line 491: This reads like a space for me but it is actually a content type. Maybe `xSpaceDisabledContentType`. Though this comment is subjective.

test/integration/entry-integration.js
```diff
@@ -419,4 +419,131 @@ describe('Entry Api', () => {
...
+        generateRandomId('test-content-type'),
+        contentTypeData
+      )
+      await contentTypeDisabledSpace.publish()
+
+      contentTypeEnabledSpace = await xSpaceEnabledEnvironment.createContentTypeWithId(
```
alicankargin, line 497: This reads like a space for me but it is actually a content type. Maybe `xSpaceEnabledContentType`. Though this comment is subjective.

test/integration/entry-integration.js
```diff
@@ -419,4 +419,131 @@ describe('Entry Api', () => {
...
+          expect(entry.fields.title['en-US']).equals('this is the title', 'original title')
+          return entry.publish().catch((accessDeniedError) => {
+            let errorMessage = JSON.parse(accessDeniedError.message)
+            expect(accessDeniedError.name).equals('AccessDenied', 'Access Denied Error')
+            expect(errorMessage.status).equals(403, '403 forbidden status')
+            expect(errorMessage.details.reasons).equals(
+              'Cross space links feature is not enabled for this space',
+              'reason explained'
+            )
+            return xSpaceDisabledEnvironment.deleteEntry(entry.sys.id)
```
alicankargin, line 528: I think this clean up doesn't belong here. You're not asserting deletion. And you're already cleaning up in the after block.

alicankargin (approved):

test/integration/entry-integration.js
```diff
@@ -419,4 +419,131 @@ describe('Entry Api', () => {
...
+          })
+        })
+    })
```
luizfonseca, line 531: same comment here, if you use `await` and variables, it might be _slightly_ better  to read (and you can skip the `return` usage

```suggestion
     const testEnviroment = xSpaceDisabledEnvironment
     const testEntry = testEnvironment.createEntry(contentTypeDisabledSpace.sys.id, entryData)
     
     expect(testEntry.fields.title['en-US']).equals('this is the title', 'original title')
     expect(testEntry.isDraft()).to.eql(true)
    
     try { 
       await testEntry.publish()
     } catch (accessDeniedError) { 
       expect(accessDeniedError.name).equals('AccessDenied', 'Access Denied Error')
       expect(errorMessage.status).equals(403, '403 forbidden status')
       expect(errorMessage.details.reasons).equals(
           'Cross space links feature is not enabled for this space',
           'reason explained'
        )
     }
      
     // Cleanup
     await  xSpaceDisabledEnvironment.deleteEntry(entry.sys.id)
})
```

luizfonseca (approved):
LGTM, some non-blocking comments

test/integration/entry-integration.js
Cyberxon: yeah for the sake of making it more clear, I will change the naming

test/integration/entry-integration.js
Cyberxon: I was trying to keep the same test structures as the previous ones and deleting the entry here is easier because we already have the entry at hand

Cyberxon: but I guess I can do it in the after block so all the clean up would be in the same place

PR #1486, opened 2022-09-20
chore(organization-membership): type user as UserProps

denkristoffer:
## Summary

Type `user` in `OrganizationMembership` as `UserProps`.

## Description

I'm assuming there's a reason we don't do this already, does anyone know why? 😅 Are there situations where the full user object is not returned?

## Motivation and Context

This is to avoid casting the `user` object to `UserProps` manually in consuming apps.

### DIFF
--- a/lib/entities/organization-membership.ts
+++ b/lib/entities/organization-membership.ts
@@ -2,13 +2,14 @@ import copy from 'fast-copy'
 import { freezeSys, toPlainObject } from 'contentful-sdk-core'
 import enhanceWithMethods from '../enhance-with-methods'
 import { wrapCollection } from '../common-utils'
-import { MetaSysProps, DefaultElements, MetaLinkProps, MakeRequest } from '../common-types'
+import { MetaSysProps, DefaultElements, MakeRequest } from '../common-types'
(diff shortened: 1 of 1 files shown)


Spring3:
> ## Summary
> Type `user` in `OrganizationMembership` as `UserProps`.
> 
> ## Description
> I'm assuming there's a reason we don't do this already, does anyone know why? 😅 Are there situations where the full user object is not returned?
> 
> ## Motivation and Context
> This is to avoid casting the `user` object to `UserProps` manually in consuming apps.

please be mindful of the fact that this is a public repository, so perhaps some discussion need happen in our private chats?

ruderngespra (changes requested):
As far as I am aware, if you query the orgMembership endpoint it will only return a link object referencing a user, not the full resolved user Information (deducting from https://www.contentful.com/developers/docs/references/user-management-api/#/reference/organization-memberships). So I assume this shouldn't be changed, looking for example at https://github.com/contentful/contentful-management.js/blob/master/lib/adapters/REST/endpoints/organization-membership.ts#L25.

wdyt?

PR #1495, opened 2022-10-03
feat: update scheduled actions types [SPA-900]

SofiaMargariti:
## Summary

Update scheduled actions types as per https://contentful.atlassian.net/wiki/spaces/ENG/pages/3891691521/CEP-0144+Enable+jobs-api+to+process+entry+aggregates#Recommended-Solution

- [x] Changes are reflected in the documentation - IN PROGRESS

When adding a new method:

### DIFF
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -61,7 +61,7 @@ export const ScheduledActionReferenceFilters =
   'sys.contentType.metadata.annotations.ContentType[nin]'
 
 export type ScheduledActionPayloadProps = {
-  withReferences: Record<typeof ScheduledActionReferenceFilters, string[]>
+  withReferences?: Record<typeof ScheduledActionReferenceFilters, string[]>
 }
 
 export type ScheduledActionProps = {
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -57,10 +57,11 @@ export type ScheduledActionSysProps = {
   updatedBy: Link<'User'> | Link<'AppDefinition'>
 }
 
-type ScheduledActionReferenceFilters = 'sys.contentType.metadata.annotations.ContentType[nin]'
+export const ScheduledActionReferenceFilters =
+  'sys.contentType.metadata.annotations.ContentType[nin]'
 
 export type ScheduledActionPayloadProps = {
-  withReferences: Record<ScheduledActionReferenceFilters, string[]>
+  withReferences: Record<typeof ScheduledActionReferenceFilters, string[]>
 }
 
 export type ScheduledActionProps = {
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -8,6 +8,7 @@ import {
   Link,
   MakeRequest,
   SysLink,
+  ScheduledActionReferenceFilters,
 } from '../common-types'
 import { wrapCollection } from '../common-utils'
 import enhanceWithMethods from '../enhance-with-methods'
@@ -57,11 +58,8 @@ export type ScheduledActionSysProps = {
   updatedBy: Link<'User'> | Link<'AppDefinition'>
 }
 
-export const ScheduledActionReferenceFilters =
-  'sys.contentType.metadata.annotations.ContentType[nin]'
-
 export type ScheduledActionPayloadProps = {
-  withReferences?: Record<typeof ScheduledActionReferenceFilters, string[]>
+  withReferences?: Record<ScheduledActionReferenceFilters, string[]>
 }
 
 export type ScheduledActionProps = {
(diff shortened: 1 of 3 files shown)


Spring3:
@SofiaMargariti `chore:` will not bump the version of the package when merged. It's a common pitfall. Please change it to `feat:` when merging

lib/entities/scheduled-action.ts
```diff
@@ -57,6 +57,13 @@ export type ScheduledActionSysProps = {
   updatedBy: Link<'User'> | Link<'AppDefinition'>
 }
 
+export const ScheduledActionReferenceFilters =
+  'sys.contentType.metadata.annotations.ContentType[nin]'
+
+export type ScheduledActionPayloadProps = {
+  withReferences: Record<typeof ScheduledActionReferenceFilters, string[]>
```
Spring3, line 64: @luizfonseca WDYT should we mark it optional? I can't think of use case now, but later if there will be other variants of possible metadata, it might be possible that this particular key will not always be present

luizfonseca: yep, it should be _optional_ otherwise we are implying that whenever payload is present there should be a `withReferences` (not true for _patches_ for example)

Spring3 (approved):

lib/entities/scheduled-action.ts
```diff
@@ -57,6 +57,13 @@ export type ScheduledActionSysProps = {
   updatedBy: Link<'User'> | Link<'AppDefinition'>
 }
 
+export const ScheduledActionReferenceFilters =
```
luizfonseca, line 60: could we make this an enum? Just to make easier for users. 

You can also export it here https://github.com/contentful/contentful-management.js/blob/aab92cea7a2829cf2887879316a281f5b5e98c50/lib/export-types.ts#L0-L1

luizfonseca: We might introduce new filters in the future as well

test/integration/scheduled-action-integration.ts
```diff
@@ -325,5 +392,67 @@ describe('Scheduled Actions API', async function () {
...
+          ],
+          metadata: aggregateRootAnnotation,
+        })
+        await contentType.publish()
+      })
+
+      afterEach(async () => {
+        await entry.delete()
+        await contentType.unpublish()
+        await contentType.delete()
```
luizfonseca, line 425: nit: should probably cancel the `scheduledActions` or we will see a spike in 404s (because the entry doesn't exist anymore)

test/integration/scheduled-action-integration.ts
```diff
@@ -325,5 +392,67 @@ describe('Scheduled Actions API', async function () {
 
       expect(updatedAction.scheduledFor.timezone).to.equal('Europe/Berlin')
     })
+
+    describe('for AggregateRoot annotated entry', () => {
```
luizfonseca, line 396: nit: could we have a test covering an entry that is not page, but is providing the payload?

test/integration/scheduled-action-integration.ts
```diff
@@ -152,6 +163,62 @@ describe('Scheduled Actions API', async function () {
...
+        await contentType.unpublish()
+        await contentType.delete()
+      })
+
+      test('create scheduled action for an aggregate root entry', async () => {
+        entry = await environment.createEntry(contentType.sys.id, {
+          fields: { title: { 'en-US': 'this is the title' } },
+        })
+
+        const scheduledAction = await testSpace.createScheduledAction({
```
luizfonseca, line 203: `unpublish` is not supposed to work for pages, only `publish`.

test/integration/scheduled-action-integration.ts
SofiaMargariti: This is already happening in the cleanup method here: https://github.com/contentful/contentful-management.js/pull/1495/files#diff-905adda6bee733e47e8d05c1da61fab78e25a3b27e5bf38ccd1ef15eb5f6d8c2L13

test/integration/scheduled-action-integration.ts
Spring3: what would be the expected test case? Aren't we testing server in the sdk's repository?

SofiaMargariti: I agree, we already cover this scenario in the backend and this PR hasn't make any changes to the way errors are handled

SofiaMargariti:
> @SofiaMargariti `chore:` will not bump the version of the package when merged. It's a common pitfall. Please change it to `feat:` when merging

Oh that's good to know, thanks!

test/integration/scheduled-action-integration.ts
Spring3: We should describe this in documentation and provide an example + error response structure and this should do

luizfonseca (approved):
🚀

PR #1497, opened 2022-10-06
feat: add space filter for org level app installations

damienxy:
## Summary

This PR adds an optional space filter to the `appInstallation.getForOrganisation` method.

When adding a new method:


mayakarabula (approved):

lib/common-types.ts
```diff
@@ -714,7 +718,7 @@ export type MRActions = {
     }
     delete: { params: GetAppInstallationParams; return: any }
     getForOrganization: {
-      params: GetOrganizationParams & { appDefinitionId: string }
+      params: GetOrganizationParams & { appDefinitionId: string; spaceId?: string }
```
marcolink, line 721: I'm wondering if the space id belongs to a query object 🤔 ?

marcolink: It was just a question - especially thinking about extending it in the future with features like pagination.

marcolink: Wondering if we can document somehow that this is a "filter"?

damienxy: > Wondering if we can document somehow that this is a "filter"?

How & where would you like to document it?

damienxy: Update: There was an issue due to default spaces and I had to put it in a query object after all, also adjusted the JSDoc.

marcolink (approved):

PR #1503, opened 2022-10-13
feat: adds content model templates [MONET-639]

RichardM99:
## Summary

- This PR introduces Environment Templates to the CMA JavaScript client.

When adding a new method:


marcolink (approved):
🎉

ruderngespra (approved):
🚢

PR #1504, opened 2022-10-13
fix: reduce the generic typings to be compatible with TS v4.8

### LINKED ISSUE #1493: not compatible with typescript 4.8.4
when having typescript 4.8.4 getting following errors:
```
../../node_modules/contentful-management/dist/typings/contentful-management.d.ts(52,41): error TS2313: Type parameter 'T' has a circular constraint.
../../node_modules/contentful-management/dist/typings/contentful-management.d.ts(52,92): error TS2536: Type 'number' cannot be used to index type '{ [K in keyof T]: { [x: string]: "workflows"; }; }'.
../../node_modules/contentful-management/dist/typings/contentful-management.d.ts(54,15): error TS2313: Type parameter 'P' has a circular constraint.
```
however it is all fine with typescript 4.2.4

Chaoste:
## Description

resolve #1493

Previously we had a generic that allows passing any array as long as it contains the string `workflows` so we ship the alpha client with this feature activated. This "black magic" from the SO post I linked in the comment doesn't work anymore for TS v4.8.x so I decided to keep it as simple as possible until we actually need a more general solution. Right now, workflows is the only alpha feature so it's fine to reduce this to a minimum.

## Motivation and Context

Pointed out by users: https://github.com/contentful/contentful-management.js/issues/1493
PR that introduced this generic type: https://github.com/contentful/contentful-management.js/pull/1122

When adding a new method:

### DIFF
--- a/lib/contentful-management.ts
+++ b/lib/contentful-management.ts
@@ -62,9 +62,9 @@ function createClient(
   }
 ): PlainClientAPI
 function createClient<
-  // T describes an array that contains the string "workflows" at any point in the list.
-  T extends (ReadonlyArray<string> | readonly ['workflows']) &
-    { [K in keyof T]: { [P in K]: 'workflows' } }[number]
+  // Originally, we used a magic expression from here: https://stackoverflow.com/a/57958451/4816930
+  // Right now it's fixed to this specific array until there is a need to revisit this.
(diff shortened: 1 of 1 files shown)


Spring3 (approved):
Merci

PR #1507, opened 2022-10-19
Monet 872 content model template integration tests

mrlaessig:
## Summary

Adds integration tests for the environments template api

- [ ] ~~Changes are reflected in the documentation~~

### DIFF
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -152,5 +152,5 @@ export const cleanupTestEnvironmentTemplates = async (olderThan = 1000 * 60 * 60
     Date.parse(template.sys.updatedAt) + olderThan < Date.now()
 
   const cleanUpTemplates = templates.filter(filterTemplate).map((templates) => templates.delete())
-  await Promise.all(cleanUpTemplates)
+  await Promise.allSettled(cleanUpTemplates)
 }
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -61,8 +61,12 @@ export const initAlphaPlainClient = (alphaFeatures = [], defaults = {}) => {
   )
 }
 
+export function getTestOrganizationId() {
+  return env.CONTENTFUL_ORGANIZATION_ID
+}
+
 export async function getTestOrganization() {
-  const testOrgId = env.CONTENTFUL_ORGANIZATION_ID
+  const testOrgId = getTestOrganizationId()
   const organizations = await initClient().getOrganizations()
   return organizations.items.find(({ sys: { id } }) => id === testOrgId)
 }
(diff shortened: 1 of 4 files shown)


marcolink:
@mrlaessig 
Without knowing how templates exactly work, I wonder if the static used names actually do work.

- What happens if multiple instances of tests run in parallel?
- What happens a test run throws, and the cleanup function is not called? 
- Does it make sense to use a naming convention for templates, and add a similar logic as we have for the [space cleanup](https://github.com/contentful/integration-test-utils/blob/main/src/space/clean-up-test-spaces.ts)?

mrlaessig:
> * What happens if multiple instances of tests run in parallel?

Regarding creation there should be an issue. You can have `n` templates with the same name. Anyways to make them more distinguishable I added a random string for each test run to the name. 
But this was a valid point for the cleanup of environment templates! Each test environment template has now a unique description per test run which is used to filter templates for deletion after a test was executed. That way no parallel test runs should accidentally delete tests of the other test runs. 

> * What happens when a test run throws, and the cleanup function is not called?

Orphan environment test templates are now cleaned up before  the test suite is executed if they are older then 1h

> * Does it make sense to use a naming convention for templates, and add a similar logic as we have for the [space cleanup](https://github.com/contentful/integration-test-utils/blob/main/src/space/clean-up-test-spaces.ts)?

Yes, a [similar approach](https://github.com/contentful/contentful-management.js/pull/1507/commits/e999e065304af66324b9a3c5d06a493e2c601bd6#diff-3b89f96d1a840d06f933bf99913ac4906757a56d4acdef7b2b96c0b956600e26R151) is now in place

marcolink (approved):
@mrlaessig this looks really great :shipit:

test/helpers.js
```diff
@@ -137,3 +141,16 @@ export const generateRandomId = (prefix = 'randomId') => {
...
+export const cleanupTestEnvironmentTemplates = async (olderThan = 1000 * 60 * 60) => {
+  const client = initClient()
+  const { items: templates } = await client.getEnvironmentTemplates(getTestOrganizationId())
+
+  const filterTemplate = (template) =>
+    template.name.startsWith(baseEnvironmentTemplateDescription) &&
+    Date.parse(template.sys.updatedAt) + olderThan < Date.now()
+
+  const cleanUpTemplates = templates.filter(filterTemplate).map((templates) => templates.delete())
+  await Promise.all(cleanUpTemplates)
```
marcolink, line 155:
```suggestion
  await Promise.allSettled(cleanUpTemplates)
```
Could be an alternative to not make the tests fail when for some reason a template deletion fails.

PR #1508, opened 2022-10-19
fix: fix query params for org installation calls

damienxy:
This PR contains 2 bug fixes: 

1/ Fix for a bug in `normalizeSpaceId` - we didn't remove the `spaceId` property from the query, now we do
2/ Handle the org id query param like other query params instead of doing string concatenation.


mayakarabula (approved):
nice

marcolink (approved):
Would love to see a test (also for documentation)

marcolink (approved):
Love it! :shipit:

PR #1511, opened 2022-10-24
fix: cursor based pagination query param interface [MONET-883]

mrlaessig:
## Summary

Align the cursor base pagination url parameters to our spec.

## Description

Iterating through pages by adding `pageNext` or `pagePrev` as a url parameter. The current implementation was using `next` and `prev` which was not aligned with our spec.

## Motivation and Context

The current implementation did not fit the definition in our spec.

### DIFF
--- a/lib/entities/scheduled-action.ts
+++ b/lib/entities/scheduled-action.ts
@@ -3,12 +3,12 @@ import copy from 'fast-copy'
 import {
   DefaultElements,
   ISO8601Timestamp,
-  BasicCursorPaginationOptions,
   MetaLinkProps,
   Link,
   MakeRequest,
   SysLink,
   ScheduledActionReferenceFilters,
+  CursorPaginatedCollectionProp,
 } from '../common-types'
 import { wrapCollection } from '../common-utils'
 import enhanceWithMethods from '../enhance-with-methods'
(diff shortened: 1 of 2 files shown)


lib/entities/scheduled-action.ts
```diff
@@ -101,19 +101,15 @@ export type CreateUpdateScheduledActionProps = Pick<
...
 
-/* eslint-disable @typescript-eslint/no-explicit-any */
-export interface ScheduledActionQueryOptions extends BasicCursorPaginationOptions {
+export interface ScheduledActionQueryOptions {
   'environment.sys.id': string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   [key: string]: any
+  limit?: number
+  next?: string
+  prev?: string
```
RichardM99, line 112: Not sure if that's right.

The response from an api supporting cursor based pagination should be

```
pages: {
  next: foo,
  prev: bar,
}
```

and passed to a query it should be `pageNext` or `pagePrev`

Here the typing is called `ScheduledActionQueryOptions` - so it should be `pageNext` or `pagePrev` because it's a query right?

mrlaessig: For me it looked like scheduled actions is using `next` and `prev` as url parameters to iterate through pages. I didn't want to break their interface because it was implemented with `next` and `prev` before.

RichardM99: Understood 👍

RichardM99 (approved):

PR #1519, opened 2022-11-02
Support custom error message for Validators

AldinDuraki:
## Summary
Support custom error message for Validators

## Motivation and Context

Currently it's not possible to add a custom error message for the validators. With this PR it is now possible.

## Checklist (check all before merging)

- [ ] Both unit and integration tests are passing
- [X] There are no breaking changes
- [ ] Changes are reflected in the documentation

PR #1536, opened 2022-11-11
feat(validation): support custom validation error message

andreascful:
Based on https://github.com/contentful/contentful-management.js/pull/1519
PR is here to create correct commit message --> release handling


ruderngespra (approved):
🚢

PR #1559, opened 2022-11-24
fix(assets): fix asset upload timeout []

andreascful:
## Summary

We saw that occassionally asset uploads would fail for bigger files (on slower connections) since the default upload timeout would be 30 seconds. This PR changes the default timeout to 5 minutes and provides the option to set the upload timeout as an option in the `createAssetFromFiles` function.
Setting the timeout to 100 seconds could then look something like this:
`environment.createAssetFromFiles(fileData, { uploadTimeout: 100000 })`

I realize that there are many variants how to get those options into the library. I found this to be the most straightforward one without applying this option to functions you might not expect to be affected. I am very happy about any suggestions though.

### DIFF
--- a/test/integration/asset-integration.js
+++ b/test/integration/asset-integration.js
@@ -45,7 +45,7 @@ describe('Asset api', function () {
     let environment
 
     before(async () => {
-      space = await createTestSpace(initClient(), 'Assets')
+      space = await createTestSpace(initClient({ retryOnError: false }), 'Assets')
       environment = await space.getEnvironment('master')
       await environment.createLocale({
         name: 'German (Germany)',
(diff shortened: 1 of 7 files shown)


ruderngespra (commented):
Lgtm, nice fix!

A few comments:

- I like the idea of making the timeout value a parameter you pass to `createFromFiles`, just because we are doing the very same thing for the `processForAllLocales` method which also is exposed by the asset API (https://github.com/contentful/contentful-management.js/blob/882f9f4c724ed22294f896e61a5f5b80ed87ca82/lib/entities/asset.ts#L101)

- On that note, do I understand it correctly that this is the timeout for uploading a file whereas the process part is the timeout for how long we want to wait for the contentful api to take until process finishing?

- Could you maybe add a test or is this a lot of setup issues?

andreascful:
@ruderngespra 

`On that note, do I understand it correctly that this is the timeout for uploading a file whereas the process part is the timeout for how long we want to wait for the contentful api to take until process finishing?`

Yes, that’s correct. So the uploadTimeout solely controls the timeout of the upload, the processing timeout is not influenced by it.

Furthermore I added a test that shows that the timeout setting is being applied correctly.

test/integration/asset-integration.js
```diff
@@ -45,7 +45,7 @@ describe('Asset api', function () {
     let environment
 
     before(async () => {
-      space = await createTestSpace(initClient(), 'Assets')
+      space = await createTestSpace(initClient({ retryOnError: false }), 'Assets')
```
andreascful, line 48: Adding this option so that tests don't take unnecessarily long. Otherwise with five retries (with increasing backoff times) it would take about 25 seconds for the new test to fail

ruderngespra (approved):
nice :)

marcolink (approved):
love the clean solution :shipit:

PR #1571, opened 2022-11-30
feat: [EXT-3173] Narrow types for a plain client with adapter

colomolo:
## Summary
Not all cma client methods are available from inside of an app. Thus we're syncing types with cma adapter possibilities.
## Approach
We are introducing new return type for a plain client, which removes all methods that doesn't work inside of an app. At `lib/contentful-management.ts` we check whether adapter has been passed as a param and override return types in such cases.

### DIFF
--- a/lib/contentful-management.ts
+++ b/lib/contentful-management.ts
@@ -9,7 +9,12 @@ import type { RestAdapterParams } from './adapters/REST/rest-adapter'
 import type { MakeRequest } from './common-types'
 import { AdapterParams, createAdapter } from './create-adapter'
 import createContentfulApi, { ClientAPI } from './create-contentful-api'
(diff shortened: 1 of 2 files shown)


Chaoste (approved):

martin3walker (approved):
LGTM 👍

andipaetzold (changes requested):
The name is not quite right. With the adapter, you can still access any method. It is not restricted. At the moment, we only have a single use-case for the adapter, which is apps.

And apps are restricted. So, I'd instead call it `PlainClientAPIViaApp` or similar.

colomolo:
> And apps are restricted. So, I'd instead call it `PlainClientAPIViaApp` or similar.

Maybe `InAppPlainClientAPI`? Wdyt?

PR #1582, opened 2022-12-07
feat: add app action type [SDK-2388]

crissto:
## Summary
Add app action type to app actions as the api will return it in the near future.

### DIFF
--- a/lib/entities/app-action.ts
+++ b/lib/entities/app-action.ts
@@ -70,6 +70,7 @@ export type AppActionProps = AppActionCategory & {
    * Human readable description of the action
    */
   description?: string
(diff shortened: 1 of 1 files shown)


lib/entities/app-action.ts
```diff
@@ -70,6 +70,7 @@ export type AppActionProps = AppActionCategory & {
    * Human readable description of the action
    */
   description?: string
+  type?: 'endpoint' | 'function'
```
andreascful, line 73: So endpoint is basically when it’s a URL it’s sending a request to and function when it’s a CF hosted function?

andreascful (approved):

PR #1624, opened 2023-01-09
feat: added metadata as optional in ReleaseProps [SPA-874]

Inbal-Gordon:
## Summary

The metadata prop is optional, ReleaseProps definition  includes the metadata attribute


lib/common-types.ts
```diff
@@ -221,6 +221,19 @@ export interface MetadataProps {
...
+    withReferences: [
+      {
+        entity: { sys: { type: 'Link'; linkType: 'Entry'; id: 'my-page-1' } }
+        filter: {
+          'sys.contentType.metadata.annotations.ContentType[nin]': ['Contentful:AggregateRoot']
+        }
+      }
+    ]
+  }
+}
```
SofiaMargariti, line 235: This still includes hardcoded values from the example payload. We need to map this into types, something like:
```
export type MetadataAggeratedRootProps = {
  metadata: {
    withReferences: {
       entity: Link<Entry>,   // explanation: whatever is passed as entity matches the Link type for Entry
       filter: Record<ScheduledActionReferenceFilters, string[]>   // explanation: whatever is passed in the filter is an object with keys that are included in the `ScheduledActionReferenceFilters` enum and values an array of strings
    }[]   // the [] here means that withReferences is an Array
}
```

Also `ScheduledActionsReferenceFilters` doesn't have a very flexible name and we cannot change it now, maybe you can create an alias and use that one here instead, like this:
`export type ReleasesReferenceFilters = ScheduledActionsReferenceFilters`

lib/common-types.ts
```diff
@@ -221,6 +221,19 @@ export interface MetadataProps {
   tags: Link<'Tag'>[]
 }
 
+export type MetadataAggeratedRootProps = {
```
SofiaMargariti, line 224: This is probably better to go into lib/entities/release.ts. You can then import here on line 82

SofiaMargariti:
Also as I learned here: https://github.com/contentful/contentful-management.js/pull/1495#issuecomment-1268339761 you'll need to change the PR name to `feat:` otherwise it won't create a new release for the client.

SofiaMargariti (commented):
@Inbal-Gordon I left some comments that might be useful. We can also pair if you have any more questions

lib/common-types.ts
Inbal-Gordon: okie will do :)

lib/common-types.ts
Inbal-Gordon: Oh okay thanks! I'll update it too

lib/entities/release.ts
```diff
@@ -80,12 +81,24 @@ export type ReleaseSysProps = {
...
 }
-
+export type ReleasesReferenceFilters = ScheduledActionReferenceFilters
+
+export type MetadataAggeratedRootProps = {
+  metadata: {
+    withReferences: [
+      {
+        entity: Link<'Entry'>
+        filter: Record<ScheduledActionReferenceFilters, string[]>[]
```
SofiaMargariti, line 91: In order to not reference scheduled actions here, could you add an alias in `lib/common-types.ts` like this:

```
export type ReleasesReferenceFilters = ScheduledActionsReferenceFilters
```
and then import `ReleaseReferenceFilters` here?

SofiaMargariti:
@Inbal-Gordon could you remove the package-lock.json from this PR. It shouldn't be needed

SofiaMargariti:
You will also need to add some tests in https://github.com/contentful/contentful-management.js/blob/master/test/integration/release-integration.ts for creating/updating a release with metadata.

The failing tests do not seem related to releases so you might be able to add the extra tests already.

PR #1633, opened 2023-01-19
chore/bump node version on CircleCI [NONE]

marcolink:
We should at least build with node `16.x`, and we should only use `npm ci`


andipaetzold:
Can we use reusable executors, so we only need to specify the CI version once at the top of config.yml?

https://circleci.com/docs/reusing-config/#authoring-reusable-executors

mayakarabula (approved):

andipaetzold (approved):

PR #1640, opened 2023-01-20
chore: bulk update and fix dependencies [NONE]

marcolink:
Bulk update several dependencies and lock `semantic-release` to v19 as any higher major version requires node@18


crissto (approved):
Nice! 🚢

lib/constants/editor-interface-defaults/types.ts
```diff
@@ -11,4 +11,5 @@ export const DEFAULT_EDITOR_ID = 'default-editor'
 /**
  * @private
  */
-export const in_ = <K extends string, O>(key: K, object: O): key is K & keyof O => key in object
+export const in_ = <K extends string, O extends object>(key: K, object: O): key is K & keyof O =>
```
andreascful, line 14: What exactly does this change do?

crissto: It forces O to be a type of object, so you can't pass things like string or number where keyof wouldn't give you the wanted result

marcolink: What @crissto says. The change was required by more strict rules of the latest Typescript version

PR #1642, opened 2023-01-20
feat: add metadata as optional in ReleaseProps [SPA-874]

Inbal-Gordon:
## Summary

The metadata prop is optional, ReleaseProps definition includes the metadata attribute


lib/entities/release.ts
```diff
@@ -80,12 +81,24 @@ export type ReleaseSysProps = {
...
+      }
+    ]
+  }
+}
 /** The object returned by the Releases API */
 export interface ReleaseProps {
   title: string
   sys: ReleaseSysProps
   entities: BaseCollection<Link<Entity>>
+  metadata: MetadataAggeratedRootProps
```
anwaar931, line 101: if metadata is optional, it should be like  `metadata?: MetadataAggeratedRootProps`

lib/entities/release.ts
```diff
@@ -80,12 +81,24 @@ export type ReleaseSysProps = {
   updatedAt: ISO8601Timestamp
   lastAction?: Link<'ReleaseAction'>
 }
-
+export type ReleasesReferenceFilters = ScheduledActionReferenceFilters
```
anwaar931, line 84:
```suggestion
export type ReleaseReferenceFilters = ScheduledActionReferenceFilters
```

lib/entities/release.ts
```diff
@@ -80,12 +81,24 @@ export type ReleaseSysProps = {
...
+export type MetadataAggeratedRootProps = {
+  metadata: {
+    withReferences: [
+      {
+        entity: Link<'Entry'>
+        filter: Record<ScheduledActionReferenceFilters, string[]>[]
+      }
+    ]
+  }
+}
```
anwaar931, line 95:
```suggestion
export type MetadataAggeratedRootProps = {
    withReferences: [
      {
        entity: Link<'Entry'>
        filter: Record<ScheduledActionReferenceFilters, string[]>[]
      }
    ]
}
```

delete `metadata` object as its already there in `ReleaseProps`

anwaar931 (changes requested):

lib/entities/release.ts
```diff
@@ -80,12 +81,24 @@ export type ReleaseSysProps = {
   updatedAt: ISO8601Timestamp
   lastAction?: Link<'ReleaseAction'>
 }
-
+export type ReleasesReferenceFilters = ScheduledActionReferenceFilters
+
+export type MetadataAggeratedRootProps = {
```
anwaar931, line 86: I would prefer it as `ReleaseMetdata` and not link this to Aggregated Root.

lib/entities/release.ts
Inbal-Gordon: okay will do thanks!

lib/entities/release.ts
```diff
@@ -80,12 +81,24 @@ export type ReleaseSysProps = {
...
+  metadata: {
+    withReferences: [
+      {
+        entity: Link<'Entry'>
+        filter: Record<ScheduledActionReferenceFilters, string[]>[]
```
anwaar931, line 91: As we have created a new type of `ReleaseReferenceFilters` we should use that here.

anwaar931 (changes requested):

anwaar931:
Just some minor comments. I don't see any other issue. Maybe @SofiaMargariti can double check

lib/entities/release.ts
Inbal-Gordon: okie

lib/entities/release.ts
```diff
@@ -80,12 +81,22 @@ export type ReleaseSysProps = {
   updatedAt: ISO8601Timestamp
   lastAction?: Link<'ReleaseAction'>
 }
+export type ReleaseReferenceFilters = ScheduledActionReferenceFilters
 
+export type ReleaseMetdata = {
+  withReferences: [
+    {
+      entity: Link<'Entry'>
+      filter: Record<ScheduledActionReferenceFilters, string[]>[]
```
SofiaMargariti, line 90:
```suggestion
      filter: Record<ReleaseReferenceFilters, string[]>[]
```

lib/entities/release.ts
```diff
@@ -80,12 +81,22 @@ export type ReleaseSysProps = {
   updatedAt: ISO8601Timestamp
   lastAction?: Link<'ReleaseAction'>
 }
+export type ReleaseReferenceFilters = ScheduledActionReferenceFilters
 
+export type ReleaseMetdata = {
```
anwaar931, line 86: Sorry there is a type, its `ReleaseMetadata`

anwaar931 (changes requested):

lib/entities/release.ts
Inbal-Gordon: oh okay

anwaar931 (approved):

PR #1644, opened 2023-01-23
chore/improve dependency caching

marcolink:


crissto (approved):
Nice! 🚢

.circleci/config.yml
```diff
@@ -6,19 +6,36 @@ executors:
...
-  install:
+  install-dependencies:
     steps:
       - restore_cache:
           keys:
-            - npm-cache-{{ checksum ".nvmrc" }}-{{ checksum "package-lock.json" }}
+            - npm-cache-{{ arch }}-{{ checksum ".nvmrc" }}-{{ checksum "package-lock.json" }}
             - npm-cache
-      - run: npm ci --prefer-offline
+      - run: npm install
```
andipaetzold, line 18: How do we make sure that only the dependency version from the lockfile are installed?
`npm ci` cannot be used anymore because it removed `node_modules`, but `npm i` might just change the `package-lock.json`

marcolink: @andipaetzold good catch, that changed slipped in and has been reverted 🙏

andipaetzold: Do we still benefit from the workspace? Do we persist the npm cache and "hope" that npm ci uses that?

marcolink: Both work perfectly in conjunction. When installing the dependencies on the prepare step, we can still consume the dependencies from the locally cached repository instead of the NPM repository.

andreascful (approved):
Very cool!

PR #1649, opened 2023-01-24
fix: release metadata types []

SofiaMargariti:
## Description

The previous implementation in #1624 is not defining the `release.metadata.withReferences` type as an array properly which causes errors when trying to use the type. Here I am fixing this.

The new types were also not exported from `./lib/export-types.ts`, so I am doing that here.

When adding a new method:

### DIFF
--- a/lib/entities/release.ts
+++ b/lib/entities/release.ts
@@ -82,15 +82,15 @@ export type ReleaseSysProps = {
   lastAction?: Link<'ReleaseAction'>
 }
 export type ReleaseReferenceFilters = ScheduledActionReferenceFilters
+export const ReleaseReferenceFilters = ScheduledActionReferenceFilters
 
 export type ReleaseMetadata = {
-  withReferences: [
-    {
-      entity: Link<'Entry'>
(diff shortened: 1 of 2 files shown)


anwaar931 (approved):

dimitrycf:
Don't think it's important but, just noticed that `ReleaseReferenceFilters` is exported twice:  as `export const` and as `export type` .  Not sure how this combination works with TypeScript 🤔

image


And it's original type seems to be an enum 

image


https://github.com/contentful/contentful-management.js/blob/106991f06343474841b4ab37184ddbe07f19458e/lib/common-types.ts#L1737

SofiaMargariti:
@dimitrycf apparently when you declare an enum you define both a type and a const (I only found out when I tried to use `ReleaseReferenceFilters`). So if you want to  alias it to another variable you need to declare it twice both as a type and as a const. The way it was before you wouldn't be able to do this for example (TS complains that we're using a type as a value):
```js
let filters;
filters[ReleaseReferenceFilters.contentTypeAnnotationNotIn] = ['test']
```

PR #1658, opened 2023-01-30
fix: same space name for parallel test [MONET-1055]

mrlaessig:
## Summary

Parallel running test use the same spaces name therefor one test can delete the other test space.

## Description

Integration tests for environment template use the same space name this can cause deletion of spaces which are still used for parallel running test.

### DIFF
--- a/test/integration/environment-template-integration.ts
+++ b/test/integration/environment-template-integration.ts
@@ -126,10 +126,15 @@ describe('Environment template Api', () => {
     let installTemplate: InstallTemplate
 
     before(async () => {
-      space = (await createTestSpace(client, 'EnvironmentTemplate')) as Space
+      // Add random string to suite name to mitigate side effects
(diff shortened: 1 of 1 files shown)


test/integration/environment-template-integration.ts
```diff
@@ -126,10 +126,15 @@ describe('Environment template Api', () => {
     let installTemplate: InstallTemplate
 
     before(async () => {
-      space = (await createTestSpace(client, 'EnvironmentTemplate')) as Space
+      // Add random string to suite name to mitigate side effects
+      // with parallel running test.
+      // Slice to 22 characters to get not more then 30 chars in total.
+      const testSuitName = generateRandomId('EnvironmentTemplate').slice(0, 22)
```
pciazynski, line 132: Just thinking if wouldn't be better to rather shorten the prefix instead of slicing the random part?

PR #1665, opened 2023-02-01
Fix: skip x-space reference flaky test [NONE]

Vinz93:
Temporarily skip the x-space reference flaky test, to unblock the CI pipeline


marcolink (approved):

andipaetzold:
@Vinz93 Just canceled the build on `master` for this. This is a change of a test and does not require a new CMA.js version. It should therefore not be a `fix:` but rather `chore:` or `test:`.

PR #1694, opened 2023-02-20
feat: allow asset.getMany to accept extra headers

SofiaMargariti:
## Summary

Allow asset.getMany method to accept extra headers.

When adding a new method:


Spring3 (approved):
🌯

PR #1734, opened 2023-03-16
feat(comments): Add support for RichTextBody [HOMER-1758]

miguelcrespo:
## Summary

This pull request extends the comment type to allow receiving and sending the body to either a:
- string
- richTextBody

## General comments

- The changes in `package-lock.json` seem to be because someone else used a different version of `node` and `npm` from the one specified in `.nvmrc` which is v14 and generated a `lockVersion: 2`
- A new property called `bodyFormat` was introduced to control the output type, this approach was preferred over directly accepting a header because when using a different `Adapter` you would need to use a weird syntax to set propertly the headers, this change affects the following Comment functions:
   - get
   - getMany
- The output type in the Comment function is controlled directly by the type of the `body` itself

## Motivation and Context

These changes are made in the effort of supporting more rich features in the comment functionality like being able to mention users


lib/plain/plain-client.ts
```diff
@@ -138,11 +138,11 @@ export const createPlainClient = (
...
-      get: wrap(wrapParams, 'Comment', 'get'),
-      getMany: wrap(wrapParams, 'Comment', 'getMany'),
-      create: wrap(wrapParams, 'Comment', 'create'),
-      update: wrap(wrapParams, 'Comment', 'update'),
-      delete: wrap(wrapParams, 'Comment', 'delete'),
+      get: plainWrap(wrapParams, 'comment', 'get'),
+      getMany: plainWrap(wrapParams, 'comment', 'getMany'),
+      create: plainWrap(wrapParams, 'comment', 'create'),
+      update: plainWrap(wrapParams, 'comment', 'update'),
+      delete: plainWrap(wrapParams, 'comment', 'delete'),
```
miguelcrespo, line 145: I won't be able to use this as the `MRActions` is being used in a lot of other parts, specially in the `lib/adapters/REST/endpoints/comment.ts` so I need to update the normal non-plain client anyway, will move this PR to draft while I resolve this

lib/plain/common-types.ts
```diff
@@ -388,15 +392,33 @@ export type PlainClientAPI = {
...
+      params: OptionalDefaults<GetManyCommentsParams & QueryParams>,
+      headers: CommentBodyFormatHeader
+    ): Promise<CollectionProp<RichTextCommentProps>>
     getMany(
       params: OptionalDefaults<GetManyCommentsParams & QueryParams>
     ): Promise<CollectionProp<CommentProps>>
+    create(
+      params: OptionalDefaults<CreateCommentParams>,
+      rawData: RichTextBodyParam,
+      headers?: AxiosRequestHeaders
```
ronaldronson, line 410: Not sure how it works, I would imagine that `AxiosRequestHeaders` is a generic enough and does not enforce the `contentful-comment-body-format` to be `rich-text` so we don't guarantee the response type in this case.

lib/plain/wrappers/wrap.ts
```diff
@@ -82,3 +83,40 @@ export const wrap = <ET extends keyof MRActions, Action extends keyof MRActions[
       headers,
     })
 }
+
+/**
+ * @private
+ */
+export const plainWrap = <ET extends keyof PlainClientAPI, Action extends keyof PlainClientAPI[ET]>(
```
ronaldronson, line 90: Could you please help me understand this change and the necessity of bringing it here?

lib/entities/comment.ts
```diff
@@ -24,16 +26,47 @@ export type CommentSysProps = Pick<
...
 export type CommentProps = {
   sys: CommentSysProps
   body: string
 }
 
+export type RichTextBodyParam = { body: RichTextDocument }
+
+export type RichTextCommentProps = Omit<CommentProps, 'body'> & RichTextBodyParam
+
+export type CommentBodyFormatHeader = Record<'x-contentful-comment-body-format', 'rich-text'>
```
Chaoste, line 59: [change request]
I guess we want to allow passing both values explicitly even though `plain-text` is also the default case.
```suggestion
export type CommentBodyFormatHeader = Record<'x-contentful-comment-body-format', 'rich-text' | 'plain-text'>
```

Chaoste: I see that right now you need it this way for the overloaded function in `common-types`. I'd suggest adding another overload and using the types `CommentPlainTextBodyHeaders` and `CommentRichTextBodyHeaders`

package-lock.json
Chaoste, line 1: [Change request] Looks like a mistake. Maybe revert this and make sure to use the correct node and npm version.

lib/plain/plain-client.ts
Chaoste: Touching this lib at the very core might be something where @andipaetzold might wanna help out or share his opinion :)

lib/plain/wrappers/wrap.ts
miguelcrespo: As the typings are right now they are very coupled between the plain client and the common types, you have to duplicate your types in both these places to get it to work, the idea with this change is to only have to modify the `PlainClientAPI` to get it to work, I will consult this with @andipaetzold

lib/plain/common-types.ts
miguelcrespo: The response type is guaranteed by the `rawData`  that can be either a string or a `Rich Text Document`

andipaetzold:
Please revert the package-lock.json

lib/adapters/REST/endpoints/comment.ts
```diff
@@ -15,11 +15,16 @@ import {
...
   GetManyCommentsParams,
+  PlainTextBodyFormat,
+  RichTextBodyFormat,
+  RichCommentBodyPayload,
 } from '../../../entities/comment'
 import { RestEndpoint } from '../types'
 import * as raw from './raw'
 import { normalizeSelect } from './utils'
 
+const bodyFormatHeader = 'x-contentful-comment-body-format'
```
cniedrich, line 26: [suggestion] I think we make CAPITAL_SNAKECASE of constant values.

cniedrich (approved):
lgtm

lib/entities/comment.ts
```diff
@@ -34,6 +41,38 @@ export type UpdateCommentProps = Omit<CommentProps, 'sys'> & {
...
+  nodeType: 'text'
+  value: string
+  marks: Mark[]
+}
+
+export interface Mention {
+  nodeType: CommentNode.Mention
+  data: { target: Link<'User'> }
+  content: MentionNodeContent[]
+}
```
Chaoste, line 60: [change request]
Is there reason not to use the original type `Text` from the rich-text package? You can find it [here](https://github.com/contentful/rich-text/blob/47f96df347c73ac301cbcb39dea23dc9236d46c6/packages/rich-text-types/src/types.ts#L34).

Chaoste (changes requested):

Chaoste (approved):

PR #1759, opened 2023-03-29
fix: improve exported types for rich text comments [HOMER-1764]

Chaoste:
## Summary

- chore: add comments about future refactoring (we're currently having typing issues in web app since `Document` and `RichTextDocument` collide.
- refactor: rename RichComment to RichTextComment (use the same prefix in all types)
- feat: add parent to comment type (very old "bug" that just surfaced in web app when adding types)
- refactpr: rename RichTextDocument to RichTextCommentDocument (this type is only usable for comments, so better make it clear that this is not about normal rich text fields)


mdmjg (approved):
LGTM

PR #1761, opened 2023-03-30
fix: forbid combining REST and App adapter types in ClientOptions

vikaskumr:
## Summary

We are currently streamlining adapter types by removing unnecessary parameters. Although the code below was previously valid, passing { apiAdapter: sdk.cmaAdapter } negates the need for additional parameters to be passed. Therefore, if you pass apiAdapter, no further parameters are required. Now the below code will give error.
```
const client = contentful.createClient(
  { apiAdapter: sdk.cmaAdapter, accessToken: 'YOUR_ACCESS_TOKEN' },
  { type: "plain" }
);

```

## Description

Change the `ClientOptions` to accept only one of the either `RestAdapterParams` / `AdapterParams`

### DIFF
--- a/lib/common-types.ts
+++ b/lib/common-types.ts
@@ -649,6 +649,16 @@ export type MakeRequestWithUserAgent = MRInternal<true>
  */
 export type MakeRequest = MRInternal<false>
 
+/**
+ * @private
+ */
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never }
+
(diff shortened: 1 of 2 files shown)


andipaetzold (approved):
Maybe tweak the title and therefore the changelog, so people understand what's going on without looking at the actual change.

Could be as simple as "Forbid combining REST and App adapter types in `ClientOptions`"

PR #1770, opened 2023-04-06
feat: [EXT-4404] Add `actions` field to create bundle method payload

colomolo:
## Summary
This will let users to upload actions manifest to link actions to the backend code, located in a bundle.

### DIFF
--- a/lib/entities/app-bundle.ts
+++ b/lib/entities/app-bundle.ts
@@ -10,6 +10,15 @@ type AppBundleSys = Except<BasicMetaSysProps, 'version'> & {
   organization: SysLink
 }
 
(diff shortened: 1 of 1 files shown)


lib/entities/app-bundle.ts
```diff
@@ -10,6 +10,15 @@ type AppBundleSys = Except<BasicMetaSysProps, 'version'> & {
   organization: SysLink
 }
 
+interface ActionManifestProps {
+  id?: string
+  name: string
+  description: string
+  category: string
+  type: string
```
kdamball, line 18: We dont need type for now

kdamball (approved):

PR #1806, opened 2023-05-04
fix: polling logic for app action call

vikaskumr:
## Summary

Polling logic is not working as expected, trying to fix the polling logic for app action call on the basis of status code.

## Description

Included the polling logic for app action call on the basis of status code.

## Motivation and Context

We were struggling to get the response from the api call so we changed this polling logic on the basis of status code.


marcolink (approved):
Would love to see a little test for the re-poll behaviour.

vikaskumr:
> Would love to see a little test for the re-poll behaviour.

Will definitely write test in the following PR after this is working fine

PR #1807, opened 2023-05-04
feat: add catalog-info [PHX-2587]

mayakarabula:


Vinz93 (approved):

catalog-info.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: contentful-management.js
+  description: |
+    Contentful command line interface. Use Contentful features straight from the command line.
```
marcolink, line 6: I think that description is wrong. As a default, I suggest we go with the repo description.

catalog-info.yaml
```diff
@@ -0,0 +1,14 @@
...
+  annotations:
+    circleci.com/project-slug: github/contentful/contentful-management.js
+    github.com/project-slug: contentful/contentful-management.js
+    contentful.com/ci-alert-slack: prd-phoenix-ci-alerts
+spec:
+  type: cli
```
marcolink, line 12: Can we change it to `library` please?

marcolink (changes requested):

catalog-info.yaml
mayakarabula: true, I was copying it for all from contentful cli, must have missed to change it here

marcolink (approved):
🚢

Issue #1701, opened 2023-02-21
The definition source of createClient is the one that has been deprecated.


yanachuwan9sm:
Hi.
As the title suggests.
It seems to be oriented towards the deprecated source of the createClient definition, even though the second argument is not passed.👋

The following is a [following the README of contentful-typescript-codegen](https://github.com/intercom/contentful-typescript-codegen)
The code for getContentfulEnvironment.js created in the root of the project.

<img width="831" alt="image" src="https://user-images.githubusercontent.com/95360878/220226331-04fe82b5-977c-4512-8072-55a1000a6493.png">

As you can see below, is it possible that `@deprecated` is affecting the overload?
I felt that Is this a phenomenon that I am the only one reproducing? 

(If so, I felt it necessary to review the VSCode settings, etc.)

https://github.com/microsoft/tsdoc/issues/131#:~:text=%E5%95%8F%E9%A1%8C%E3%81%AB%E6%B0%97%E3%81%A5%E3%81%8D%E3%81%BE%E3%81%97%E3%81%9F%E3%80%82%40deprecated%E3%81%AF%E3%82%AA%E3%83%BC%E3%83%90%E3%83%BC%E3%83%AD%E3%83%BC%E3%83%89%E3%81%AB%E5%BD%B1%E9%9F%BF%E3%81%97%E3%81%BE%E3%81%99%E3%80%82

Also, by modifying `contentful-management.d.ts` as follows
The definition source is now oriented to the createClient, which can be expected.
(I don't think this is in line with [best practice](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#function-overloads), but... ,)

I am not very knowledgeable in this area, so please can someone lend me their knowledge🙏

```ts
/**
 * Create a client instance
 * @param params - Client initialization parameters
 *
 * ```javascript
 * const client = contentfulManagement.createClient({
 *  accessToken: 'myAccessToken'
 * })
 * ```
 */
declare function createClient(params: ClientOptions): ClientAPI;
declare function createClient(params: ClientOptions, opts: {
    type: 'plain';
    defaults?: DefaultParams;
}): PlainClientAPI;

/**
 * @deprecated
 */
export type ClientParams = RestAdapterParams & UserAgentParams;
type ClientOptions = (RestAdapterParams | AdapterParams) & UserAgentParams;
/**
 * @deprecated the alphaFeatures option is not longer supported.
 *
[...]

PR #1841, opened 2023-05-26
fix: change overload order to not show the deprecation per default [HOMER-1999]

Chaoste:
## Summary

In some IDEs, it will only pick up the JSDoc of the first overload for a function. By moving the deprecated overload to the end, we hope to mitigate the wrongly shown deprecation warning.

## Description

We need to change the order of function overloads for `createClient`. In some cases, the IDE will pick up only the JSDoc for the first overload and even mark the usage as `deprecated` as shown in [this issue](https://github.com/contentful/contentful-management.js/issues/1701). Weirdly, we can't reproduce this consistently with the same version of VSC & TS. In some cases, it works as expected. In a few other cases, the TS engine appears to pick up only the first overload and render any usage of the function as being deprecated.

This is a reported issue for VSC ([issue](https://github.com/microsoft/TypeScript/issues/407#issuecomment-650266584)) and WebStorm ([issue](https://youtrack.jetbrains.com/issue/WEB-59045)). It appears we can mitigate this scenario by changing the order of overloads.

We also tested this on the TS playground ([see here](https://www.typescriptlang.org/play?ts=5.0.4#code/PQKhCgAIUgBATApgBwE6IMYEMAuj6Q4AWikAlgM4DKArgEYU6pYY5kD2AdpO8m1+QqRO7HJAA2XAOaJUkCjWTJ2qPPAB0kAArjEWCqRoHIWeSizM8kALaJi7AgDMVhIrhOcC6a+wBupYkoePg5uR1R2a1dSRxpOVlDIbHFxdSgQYChY+P5uLHh4ADkaazpZCgAKLAAuYRKy1AAaSDpaznrZZt4cClqAbyhIIcpCxClcMn8Aflq6dnZdLE4AbkHh6npGZgSuWfnFlfAAXwBKNo7ULLidvILi0vKq84emlueGs7qXq5zE-KKLpUal8Gs1WiDOsEev01oJRuM2NM9gs9IdTu9ZD8biY7oCnhDXuD2i9mn04WMJv5mpRaAwmCxcpAjjChut4ZTEDMWvtUatWYJaVsGaEuXMUUtVkdIABeSB9dEEuWwjBcRjyZIWGXkhGTUhTSAAWgAjJBakbYWRHJAKjTNvSbiclfz+egcDRUNwqoaWo6YBQNZd+VLWa73Z7TABqH3QdVYcQWY7gcAqzhquhYABeWv+9walROSZTauc7GzuJelQATM0ACwnVZFlHqSRSColgvJ1VidNyWU5vEADmaADZSQK7dtcrUmDRSKcG6qmy2Kj2O+AgA)) but couldn't reproduce it there as well.

### DIFF
--- a/lib/contentful-management.ts
+++ b/lib/contentful-management.ts
@@ -44,17 +44,6 @@ export type ClientParams = RestAdapterParams & UserAgentParams
 type ClientOptions = UserAgentParams & XOR<RestAdapterParams, AdapterParams>
 
 /**
- * @deprecated the alphaFeatures option is not longer supported
- */
-function createClient(
-  params: ClientOptions,
-  opts: {
-    type?: 'plain'
-    alphaFeatures: string[]
-    defaults?: DefaultParams
-  }
-): ClientAPI | PlainClientAPI
-/**
  * Create a client instance
  * @param params - Client initialization parameters
  *
@@ -72,6 +61,17 @@ function createClient(
     defaults?: DefaultParams
   }
 ): PlainClientAPI
+/**
+ * @deprecated the alphaFeatures option is not longer supported
+ */
+function createClient(
+  params: ClientOptions,
+  opts: {
+    type?: 'plain'
+    alphaFeatures: string[]
+    defaults?: DefaultParams
+  }
+): ClientAPI | PlainClientAPI
 function createClient(
   params: ClientOptions,
   opts: {


Chaoste:
@denkristoffer do you know whether someone is looking the broken CI pipeline for this repo?

denkristoffer:
@Chaoste I don't know, no 😞

denkristoffer (approved):

Issue #1853, opened 2023-06-12
Typescript: ContentTypeFieldValidation interface does not include nodes field


luiscruzih:
Looks like validations[].nodes is used to define which entries can be selected when embedding an entry in a rich text field

<img width="278" alt="image" src="https://github.com/contentful/contentful-management.js/assets/116587381/55159281-4f8e-48ea-94cd-b2d6bdfb5d68">

However, while using contentful-management validation.nodes raises a TS error.

Could this be added to ContentTypeFieldValidation definition?

Something like
```
nodes?: {
      [key: string]: Pick<OriginalContentTypeFieldValidation, 'linkContentType' | 'message'>[];
};
```

or just

```
nodes?: {
      [key: string]: OriginalContentTypeFieldValidation[];
};
```

luiscruzih:
https://github.com/contentful/contentful-management.js/pull/1854

mgoudy91:
Looks like this was covered by https://github.com/contentful/contentful-management.js/pull/1858

PR #1854, opened 2023-06-12
add nodes to ContentTypeFieldValidation

luiscruzih:
add missing field to ContentTypeFieldValidation


IgnacioNMiranda:
I think I did the same here 🤣 #1853

PR #1856, opened 2023-06-20
feat: extend type CommentProps for new status marker

Lelith:
## Summary
The comments api got extended and delivers now a status field. This field is meant to represent the comment-thread status. Comments can now be marked as resolved, in order to help organize conversations in comments. The currently available states are 'active' and 'resolved'

### DIFF
--- a/lib/entities/comment.ts
+++ b/lib/entities/comment.ts
@@ -38,7 +38,10 @@ export type RichTextBodyProperty = 'rich-text'
 export type RichTextBodyFormat = { bodyFormat: RichTextBodyProperty }
 export type PlainTextBodyFormat = { bodyFormat?: PlainTextBodyProperty }
 
-export enum CommentStatus {'active', 'resolved'}
+export enum CommentStatus {
+  'active',
+  'resolved',
+}
 
 export type CommentProps = {
   sys: CommentSysProps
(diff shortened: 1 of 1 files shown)


cniedrich (approved):
LGTM

lib/entities/comment.ts
```diff
@@ -38,9 +38,15 @@ export type RichTextBodyProperty = 'rich-text'
...
 
+export enum CommentStatus {
+  'active',
+  'resolved',
+}
+
 export type CommentProps = {
   sys: CommentSysProps
   body: string
+  status?: CommentStatus
```
miguelcrespo, line 49: issue: I think `status` is always available and defaulting to `active`

miguelcrespo (changes requested):

lib/entities/comment.ts
cniedrich: yes! for get, `status` is always there, just or PUT/POST you do not need to provide it.

Lelith: But what about existing comments that have not yet the field set?

miguelcrespo: Then we would need to update these types for those cases:

https://github.com/contentful/contentful-management.js/blob/c80524ca262e18a7ac59c48be310b916691a9a17/lib/entities/comment.ts#L52-L55

miguelcrespo: @Lelith comments that were created before this property existed will return `active`

Lelith: ok, will change it to non optional :)

Lelith:
Do we have to mark this release then as breaking / major update?

Lelith:
closing for now, will update in a different story

PR #1858, opened 2023-06-21
feat: add types for nodes validations

BasKiers:
## Summary

Adds types for `nodes` validation object in `ContentFields.validations`

## Description

* Changes `enabledNodeTypes` validation to only accept valid string values
* Creates a new interface NodesValidation which includes the available types for Rich Text nodes
* Adds validation type for the new `embedded-resource-block` node
* Add nodes key to the ContentTypeFieldValidation interface

## Motivation and Context

The lack of a defined type makes it hard to work with RichText field validations while using Typescript and will not autocomplete the property while using javascript.

This PR supersedes #1829 and #1854 as this PR has somewhat stricter types defined and also includes the type for the `embedded-resource-block` node.

When adding a new method:

### DIFF
--- a/lib/entities/content-type-fields.ts
+++ b/lib/entities/content-type-fields.ts
@@ -1,4 +1,5 @@
 import { KeyValueMap } from '../common-types'
+import { INLINES, BLOCKS } from '@contentful/rich-text-types'
 
 interface NumRange {
   min?: number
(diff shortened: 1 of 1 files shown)


lib/entities/content-type-fields.ts
```diff
@@ -15,11 +16,23 @@ interface RegExp {
...
+    allowedResources: ContentTypeAllowedResources
+  }
+}
+
 export interface ContentTypeFieldValidation {
   linkContentType?: string[]
   in?: (string | number)[]
   linkMimetypeGroup?: string[]
-  enabledNodeTypes?: string[]
+  enabledNodeTypes?: `${BLOCKS}` | `${INLINES}`
```
andipaetzold, line 35: This should be an array, right?

BasKiers: You're right, overlooked it

lib/entities/content-type-fields.ts
```diff
@@ -15,11 +16,23 @@ interface RegExp {
...
+  [INLINES.EMBEDDED_ENTRY]?: Pick<ContentTypeFieldValidation, 'size' | 'linkContentType'>[]
+  [INLINES.ENTRY_HYPERLINK]?: Pick<ContentTypeFieldValidation, 'size' | 'linkContentType'>[]
+  [BLOCKS.EMBEDDED_ASSET]?: Pick<ContentTypeFieldValidation, 'size'>[]
+  [INLINES.ASSET_HYPERLINK]?: Pick<ContentTypeFieldValidation, 'size'>[]
+  [BLOCKS.EMBEDDED_RESOURCE]?: {
+    validations: Pick<ContentTypeFieldValidation, 'size'>[]
+    allowedResources: ContentTypeAllowedResources
+  }
+}
+
```
IgnacioNMiranda, line 30: I'd say we're missing the `message` property in some of these

BasKiers: Yes we are

lib/entities/content-type-fields.ts
```diff
@@ -15,11 +16,32 @@ interface RegExp {
...
+  >[]
+  [INLINES.ENTRY_HYPERLINK]?: Pick<
+    ContentTypeFieldValidation,
+    'size' | 'linkContentType' | 'message'
+  >[]
+  [BLOCKS.EMBEDDED_ASSET]?: Pick<ContentTypeFieldValidation, 'size' | 'message'>[]
+  [INLINES.ASSET_HYPERLINK]?: Pick<ContentTypeFieldValidation, 'size' | 'message'>[]
+  [BLOCKS.EMBEDDED_RESOURCE]?: {
+    validations: Pick<ContentTypeFieldValidation, 'size' | 'message'>[]
+    allowedResources: ContentTypeAllowedResources
```
andipaetzold, line 36: Me again with the same question: This should also be an array, right?

andipaetzold (changes requested):

lib/entities/content-type-fields.ts
BasKiers: Thank you for reviewing so thoroughly, you are right again!

IgnacioNMiranda (approved):

andipaetzold (approved):

PR #1859, opened 2023-06-22
feat: [EXT-3599] Expose plain client type

colomolo:
## Summary
Properly exposing `PlainClientAPI` type.

### DIFF
--- a/lib/export-types.ts
+++ b/lib/export-types.ts
@@ -1,4 +1,5 @@
 export * from './common-types'
+export * from './plain/common-types'
 export type { ApiKey, ApiKeyProps, CreateApiKeyProps } from './entities/api-key'
 export type {
   AppAction,


marcolink:
@colomolo out of curiosity, will this also have an effect on the generated docs?

colomolo:
> @colomolo out of curiosity, will this also have an effect on the generated docs?

I don't know, honestly. Which docs do you mean?

marcolink:
> > @colomolo out of curiosity, will this also have an effect on the generated docs?
> 
> I don't know, honestly. Which docs do you mean?

https://contentful.github.io/contentful-management.js/contentful-management/10.37.0/

colomolo:
> > > @colomolo out of curiosity, will this also have an effect on the generated docs?
> > 
> > 
> > I don't know, honestly. Which docs do you mean?
> 
> https://contentful.github.io/contentful-management.js/contentful-management/10.37.0/
> 
> ... you can test it by running [this](https://github.com/contentful/contentful-management.js/blob/master/package.json#L27) script locally

Yes, it's there.

lib/export-types.ts
```diff
@@ -1,4 +1,5 @@
 export * from './common-types'
+export * from './plain/common-types'
```
andipaetzold, line 2: export all or just that specific type?  Atm it's the same, but in the future we might export more types from `common-types`

andipaetzold (approved):

marcolink (approved):

lib/export-types.ts
colomolo: All of them. Hence the name 'common-types'

andipaetzold:
Before we merge: We are currently in the weird situation where we export types from `types` and from the root. `PlainClientAPI` is already exported from the root. Probably something we should eventually fix by exporting everything from the root.

-> It's already possible to import the type without this PR.

colomolo:
> Before we merge: We are currently in the weird situation where we export types from `types` and from the root. `PlainClientAPI` is already exported from the root. Probably something we should eventually fix by exporting everything from the root.
> 
> -> It's already possible to import the type without this PR.

Yeah, this is not only about `PlainClientAPI`, but relates to all exported types. So I think we should consider a broader solution for this root vs '/types' export thing.

PR #1880, opened 2023-07-10
feat: add new annotation type [MONET-1301]

cemreyuksel:
## Summary

This PR changes the typing for `ContentfulMetaData` in order to align with new Annotations.

## Description

`AnnotationAssignment` type is added to `ContentfulMetaData`.

When adding a new method:

### DIFF
--- a/lib/entities/content-type.ts
+++ b/lib/entities/content-type.ts
@@ -28,7 +28,9 @@ export type ContentTypeMetadata = {
   >
 }
 
-export type AnnotationAssignment = Link<'Annotation'> & { parameters?: Record<string, string | number | boolean> }
+export type AnnotationAssignment = Link<'Annotation'> & {
+  parameters?: Record<string, string | number | boolean>
+}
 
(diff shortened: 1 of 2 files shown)


kdamball (commented):
This is a breaking change. Do we need a full version bump?

lib/entities/content-type.ts
```diff
@@ -21,13 +21,17 @@ import { omitAndDeleteField } from '../methods/content-type'
 export type ContentTypeMetadata = {
   annotations?: RequireAtLeastOne<
     {
-      ContentType?: Link<'Annotation'>[]
+      ContentType?: AnnotationAssignment[]
       ContentTypeField?: Record<string, Link<'Annotation'>[]>
```
RichardM99, line 25:
```suggestion
      ContentTypeField?: Record<string, AnnotationAssignment[]>
```

cemreyuksel:
@kdamball yes, good point. I will add. 👍

PR #1892, opened 2023-07-19
chore: bump axios to 1.x [PHX-2741]

marcolink:
## Summary

Bump `contentful-sdk-core` to the latest major version which also pulls in `axios@1.4.0`

## Description

Bump `contentful-sdk-core` to the latest major version which also pulls in `axios@1.4.0`. 
This requires some adjustment in the codebase: 

### Headers shape
The old shape of `AxiosRequestHeaders` is now represented by `RawAxiosRequestHeaders`.

### Remove webpack plugin
After updating the dependencies, one of the build steps threw an error. This was caused by the babel plugin `babel-minify-webpack-plugin` which we used to minify the bundles. [We still minify through loader options](https://github.com/contentful/contentful-management.js/blob/master/webpack.config.js#L26).

## Motivation and Context

`axios@1.x` is now released for ~10 months and stable. To ensure we can receive updates for all our dependencies, we have to eventually update to the latest major. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -159,3 +159,5 @@ yarn.lock
 
 
 reports
+
+.tool-versions
--- a/lib/adapters/REST/endpoints/api-key.ts
+++ b/lib/adapters/REST/endpoints/api-key.ts
@@ -1,4 +1,4 @@
-import { AxiosRequestHeaders } from 'axios'
+import { RawAxiosRequestHeaders } from 'axios'
 import type { AxiosInstance } from 'contentful-sdk-core'
 import copy from 'fast-copy'
 import { SetOptional } from 'type-fest'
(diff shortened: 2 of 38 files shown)


andreascful (approved):

andipaetzold:
Should this be a breaking change of CMA.js?

marcolink:
@andipaetzold 
> Should this be a breaking change of CMA.js?

The main change (header type) is NOT changing the type of the publicly exposed API  
I also tested this PR against the Web App - no type violations detected. 

The [other props exposed by our underlaying library](https://github.com/contentful/contentful-sdk-core/blob/master/src/types.ts#L23-L118) don't seem to introduce a breaking change. 

Any other touch points we should check beforehand?

luizfonseca:
2 cents: this should've been a breaking change as @andipaetzold mentioned -- more so because of how this plays with peer deps and the fact that `contentful-sdk-core` has a different version now (might break other systems from building artifacts): https://github.com/contentful/contentful-sdk-core/blob/b014b841c0c09613a1847a6b34282224bc0e16dc/package.json#L70

PR #1940, opened 2023-08-31
feat: support optional expiresIn param in token creation [AHOY-2470]

budimirbudimir:
## Summary

- Support optional `expiresIn` parameter in `CreatePersonalAccessTokenProps`.
- Support optional `expiresAt` parameter in `PersonalAccessTokenProp.sys`.

## Description

- Update 2 out-of-sync types to support new CMA access tokens TTL features.
- Test `personalAccessToken.create` and `personalAccessToken.get` methods on the plain client
- Test `createPersonalAccessToken` and `getPersonalAccessToken` methods on the default client

## Motivation and Context

We are updating the client types to reflect the API changes.
In the follow up pull requests, additional changes will be introduced, moving away from Personal Access Tokens (PATs) and using new entity of unified Access Tokens (PAT, CLI, OAuth) will be introduced.
For now, though, we are just updating PAT types.

When adding a new method:

### DIFF
--- a/lib/entities/personal-access-token.ts
+++ b/lib/entities/personal-access-token.ts
@@ -5,7 +5,7 @@ import { wrapCollection } from '../common-utils'
 import { MetaSysProps, DefaultElements, MakeRequest } from '../common-types'
 
 export type PersonalAccessTokenProp = {
-  sys: MetaSysProps & { expiresAt?: number }
+  sys: MetaSysProps & { expiresAt?: string }
   name: string
   scopes: 'content_management_manage'[]
   revokedAt: null | string
--- a/lib/entities/personal-access-token.ts
+++ b/lib/entities/personal-access-token.ts
@@ -5,14 +5,16 @@ import { wrapCollection } from '../common-utils'
 import { MetaSysProps, DefaultElements, MakeRequest } from '../common-types'
 
 export type PersonalAccessTokenProp = {
-  sys: MetaSysProps
+  sys: MetaSysProps & { expiresAt?: number }
   name: string
   scopes: 'content_management_manage'[]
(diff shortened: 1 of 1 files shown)


piankovev (approved):
Looks good to me 👍

lib/entities/personal-access-token.ts
```diff
@@ -5,14 +5,16 @@ import { wrapCollection } from '../common-utils'
 import { MetaSysProps, DefaultElements, MakeRequest } from '../common-types'
 
 export type PersonalAccessTokenProp = {
-  sys: MetaSysProps
+  sys: MetaSysProps & { expiresAt?: number }
```
Annmary12, line 8:
```suggestion
  sys: MetaSysProps & { expiresAt?: string }
```

budimirbudimir: @Annmary12 I actually believe it should be a number here, right?
As we're sending the number to the API as well:

Screenshot 2023-09-01 at 15 22 51

budimirbudimir: Also, in `GenerateCMATokenDialog` on lines 148-150, we are also setting the following payload as number:

Screenshot 2023-09-01 at 15 25 53

Annmary12: I was not referring to `expiresIn`, it is `expiresAt` that I highlighted which is a date string

budimirbudimir: Sorry, you are correct, not sure why I went onto `expiresIn`. 🤦 
I'll see to fix this. 👍

Annmary12 (approved):
🎉

PR #1948, opened 2023-09-01
feat: allow retry and retryInterval to be specified in appActionCall.createWithResponse [INTEG-1208]

jsdalton:
## Summary

Allows the caller to specify a `retries` and/or `retryInterval` in calls to `appActionCall.createWithResponse()`

## Motivation and Context

The current `retry` and `retryInterval` values are hard coded to 10 and 2 seconds respectively. But some operations implemented with App Action can take longer than 20 seconds (for example, AI image processing).

## Checklist (check all before merging)

- [x] Both unit and integration tests are passing
- [x] There are no breaking changes
- [x] Changes are reflected in the documentation

When adding a new method:

- [ ] The new method is exported through the default and plain CMA client
- [ ] All new public types are exported from `./lib/export-types.ts`
- [ ] Added a unit test for the new method
- [ ] Added an integration test for the new method
- [ ] The new method is added to the documentation

PR #1953, opened 2023-09-05
fix: retry AppActionCall for up to 30s [INTEG-1208]

jsdalton:
## Summary

Extends the retry duration when using `AppActionCall.createWithResponse` to 30 seconds (15 retries with a 2 second retry interval.

## Description

Today, createWithResponse retries only 10 times with a 2s interval, which works out to about 20 seconds. We just extend the polling here to 30 seconds.

## Motivation and Context

Long running hosted app actions can take longer than 20 seconds. We are discovering that AI image manipulation frequently takes 20-25 seconds.

When adding a new method:


jsdalton:
This PR is complementary to #1948 (and perhaps less controversial).

ryunsong-contentful (approved):

marcolink (approved):
LGTM 👍 
(We assume that the polling stops if an app action timed out)

jjolton-contentful (approved):

PR #1966, opened 2023-09-14
fix(types): content type validation [NONE]

marcolink:
## Summary
We need to respect the legacy shape of content type validations where the field `linkContentType` can also just by a string.

### DIFF
--- a/lib/entities/content-type-fields.ts
+++ b/lib/entities/content-type-fields.ts
@@ -46,7 +46,7 @@ interface NodesValidation {
 }
 
 export interface ContentTypeFieldValidation {
-  linkContentType?: string[]
+  linkContentType?: string | string[]
   in?: (string | number)[]
   linkMimetypeGroup?: string[]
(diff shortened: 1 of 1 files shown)


veu (approved):

andipaetzold (approved):

iTechHarryTurnbull:
Hello there,

I'm afraid this change has broken a lot of our content model as code migrations, as they had previously only considered the `linkContentType` validation as `string[]`, and are now not narrowing `string | string[]`.

Should this be considered a breaking change?

I've been working with Contentful for 2 years now and I've not seen any documentation about the `string` form of `linkContentType`, could you point me to this please? Should we just avoid it if using the new `string[]` form?

Many thanks,
Harry

marcolink:
@iTechHarryTurnbull sorry to hear this is causing a problem for you. 
We addressed your findings with #1985

iTechHarryTurnbull:
@marcolink sorry for the delay, thanks for sorting this out! 🙏

PR #1977, opened 2023-09-20
fix(types): environment template adapter [MONET-1248]

RichardM99:
## Summary
- Extends environment template `get` to accept query parameters.
- Fixes environment template `getMany` and `versions` to correctly pass cursor based pagination options as part of query params (along with including `select` in the query params).

### DIFF
--- a/lib/adapters/REST/endpoints/environment-template.ts
+++ b/lib/adapters/REST/endpoints/environment-template.ts
@@ -12,7 +12,9 @@ export const get: RestEndpoint<'EnvironmentTemplate', 'get'> = (
   { organizationId, environmentTemplateId, version, query = {} }
 ) =>
   version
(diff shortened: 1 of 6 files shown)


ruderngespra (approved):
lgtm, thanks for keeping the sdk clean :)

So this is basically already existing api behavior which was missed when first introducing the client for environment templates?

Maybe we can add an additional test case for the querying somewhere [here](https://github.com/contentful/contentful-management.js/blob/master/test/integration/environment-template-integration.ts#L39) or extend the existing one with a query?

RichardM99:
Yes exactly - was missed first time around.

Now that I am looking at adding a test - realised there are some issues with the PR.

Will reach out again once I have resolved them and added a test 👍

PR #1981, opened 2023-09-21
docs: add inline docs for the plain client Webhook interface [EXT-4720]

t-col:
## Summary

adds inline documentation for the Webhook plain client interface

## Motivation and Context


mgoudy91 (approved):
💯

PR #1982, opened 2023-09-22
docs: add inline docs for the plain client App Signed Request and Signing Secret interface [EXT-4726]

t-col:
## Summary

adds inline documentation for the App Signing Secret and Signed Request plain client interfaces

## Motivation and Context

### DIFF
--- a/lib/create-space-api.ts
+++ b/lib/create-space-api.ts
@@ -109,7 +109,7 @@ export default function createSpaceApi(makeRequest: MakeRequest) {
      * })
      *
      * client.getSpace('<space_id>')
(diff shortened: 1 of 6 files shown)


lib/create-space-api.ts
```diff
@@ -150,7 +150,7 @@ export default function createSpaceApi(makeRequest: MakeRequest) {
     },
 
     /**
-     * Creates an Environement
+     * Creates an environment
```
mgoudy91, line 153: lol

mgoudy91 (approved):

PR #1984, opened 2023-09-25
feat: use a single axios instance [PHX-2863]

marcolink:
## Motivation
A single `axios` instance allows for [stateful adapters](https://github.com/contentful/contentful-sdk-core/pull/214). there is technically also no need to create an instance for every call. 

## Implementation
We create a `axiosInstance` calls member to the RestAdapter, and pass it in to the endpoint function.

### DIFF
--- a/lib/adapters/REST/rest-adapter.ts
+++ b/lib/adapters/REST/rest-adapter.ts
@@ -21,6 +21,8 @@ export type RestAdapterParams = CreateHttpClientParams & {
    * @default upload.contentful.com
    */
   hostUpload?: string
+
+  userAgent?: string | undefined
 }
 
 /**
@@ -35,7 +37,7 @@ export class RestAdapter implements Adapter {
   private readonly params: RestAdapterParams
   private readonly axiosInstance: AxiosInstance
 
-  public constructor(params: RestAdapterParams, userAgent?: string | undefined) {
+  public constructor(params: RestAdapterParams) {
     if (!params.accessToken) {
       throw new TypeError('Expected parameter accessToken')
     }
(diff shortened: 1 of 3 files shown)


lib/adapters/REST/rest-adapter.ts
```diff
@@ -70,20 +80,9 @@ export class RestAdapter implements Adapter {
       throw new Error('Unknown endpoint')
     }
 
-    const requiredHeaders = {
-      'Content-Type': 'application/vnd.contentful.management.v1+json',
+    return await endpoint(this.axiosInstance, params, payload, {
+      ...headers,
       'X-Contentful-User-Agent': userAgent,
```
andipaetzold, line 85: Doesn't this always overwrite the header set in line 52?

lib/adapters/REST/rest-adapter.ts
```diff
@@ -33,8 +33,9 @@ const defaultHostParameters = {
 
 export class RestAdapter implements Adapter {
   private readonly params: RestAdapterParams
+  private readonly axiosInstance: AxiosInstance
 
-  public constructor(params: RestAdapterParams) {
+  public constructor(params: RestAdapterParams, userAgent?: string | undefined) {
```
andipaetzold, line 38: Should we make `userAgent` part of `params`? This class & constructor is public and we might want to add more options in the future. Passing it as part of `params` might be cleaner.

andreascful (approved):
Nice change. I think having just one instance is the way to go here

lib/adapters/REST/rest-adapter.ts
marcolink: No it only overwrites if defined

lib/adapters/REST/rest-adapter.ts
marcolink: done!

lib/adapters/REST/rest-adapter.ts
```diff
@@ -21,6 +21,8 @@ export type RestAdapterParams = CreateHttpClientParams & {
    * @default upload.contentful.com
    */
   hostUpload?: string
+
+  userAgent?: string | undefined
```
andipaetzold, line 25: Maybe add jsdoc as for the other params

andipaetzold (approved):

mgoudy91 (approved):

t-col:
@marcolink any reason not to merge this? (after a rebase of course)

PR #1986, opened 2023-09-25
docs: add inline docs for the plain client UI Extensions [EXT-4719]

jjolton-contentful:
## Summary

adds inline documentation for the UI Extension plain client interfaces

## Motivation and Context

### DIFF
--- a/lib/plain/entities/extension.ts
+++ b/lib/plain/entities/extension.ts
@@ -25,8 +25,8 @@ export type ExtensionPlainClientAPI = {
    */
   get(params: OptionalDefaults<GetExtensionParams & QueryParams>): Promise<ExtensionProps>
   /**
-   * Fetches all Extensions of a Space
-   * @param params entity IDs to identify the Environment from which to fetch Extensions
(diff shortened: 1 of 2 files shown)


lib/plain/entities/extension.ts
```diff
@@ -0,0 +1,177 @@
...
+   * const uiExtension = await client.extension.get({
+   *   spaceId: '<space_id>',
+   *   environmentId: '<environment_id>',
+   *   extensionId: '<extension_id>',
+   * });
+   * ```
+   */
+  get(params: OptionalDefaults<GetExtensionParams & QueryParams>): Promise<ExtensionProps>
+  /**
+   * Fetches all Extensions of a Space
```
t-col, line 28: is this supposed to be `Space` or `Environment`? the next line refers to `Environment`

t-col (approved):
💯

lib/plain/entities/extension.ts
jjolton-contentful: good call, it is using `Space/Environment` elsewhere, switched to that

PR #2011, opened 2023-10-05
feat: add client methods for AppEventSubscriptions and AppKeys [EXT-4755] [EXT-4757]

jjolton-contentful:
## Summary

Adds the `upsert`, `get`, and `delete` client methods for AppEventSubscriptions, and the `create`, `get`, `getMany` and `delete` client methods for AppKeys

## Motivation and Context

CMA JS client <-> API feature parity

When adding a new method:

### DIFF
--- a/lib/entities/app-key.ts
+++ b/lib/entities/app-key.ts
@@ -10,19 +10,35 @@ type AppKeySys = Except<BasicMetaSysProps, 'version'> & {
   organization: SysLink
 }
 
+export interface JWK {
+  alg: 'RS256'
+  kty: 'RSA'
+  use: 'sig'
+  x5c: [string]
+  kid: string
+  x5t: string
+}
+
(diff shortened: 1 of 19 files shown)


lib/entities/app-key.ts
```diff
@@ -0,0 +1,84 @@
...
+   */
+  sys: AppKeySys
+  jwk: JsonWebKey
+}
+
+export type CreateAppKeyProps = {
+  /** Toggle for automatic private key generation */
+  generate?: boolean
+  /** JSON Web Key, required if generate is falsy */
+  jwk?: string
```
jjolton-contentful, line 25: I think I have the types wrong here, fix in progress

t-col (approved):
LGTM! I might recommend we split out PRs for each entity in the future

PR #2013, opened 2023-10-09
feat: add installation param schema to app definition INTEG-1385

rusticpenguin:
## Summary

We are adding installation parameter schemas to the app definition

## Description

We need to update this here so we can reference the installation property in User Interface

## Motivation and Context

App creators have to currently use the frontend to validate user's installation params. This will allow them to offload most of that logic to our backend when the user saves


matthew-gordon (approved):
LGTM, probably should get some more 👀 from @contentful/team-phoenix

jsdalton (approved):

PR #2027, opened 2023-10-30
feat: fix comment types & support parentEntityReference, replies & status

andipaetzold:
Follow-up of https://github.com/contentful/contentful-management.js/pull/2023

* You can create replies by passing `parentCommentId`
* You can add a parent entity reference by passing `parentEntityReference`
* Types for deletion and updating now support content types and workflows
* Add `status` type
* Allow filtering by `status`

Do we want to deprecate the old `{ entryId: string }` syntax for creating comments?

### DIFF
--- a/lib/adapters/REST/endpoints/comment.ts
+++ b/lib/adapters/REST/endpoints/comment.ts
@@ -23,7 +23,10 @@ import { RestEndpoint } from '../types'
 import * as raw from './raw'
 import { normalizeSelect } from './utils'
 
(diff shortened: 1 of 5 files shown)


massao (approved):
LGTM

bgutsol (approved):
The best! Thanks a lot for bringing `status` param as well 🏅

> Do we want to deprecate the old { entryId: string } syntax for creating comments?

I think it should be on @contentful/team-homer to decide

jfctfl (approved):
LGTM!

PR #2029, opened 2023-11-01
feat: add webhook signing secrets to legacy and plain client interfaces [EXT-4817]

t-col:
## Summary

Adds plain client methods for fetching, upserting, and deleting webhook signing secrets.

When adding a new method:


ghepting (approved):
nice!

mgoudy91 (approved):

jjolton-contentful (approved):
🧼

PR #2030, opened 2023-11-01
fix: update createUpload to accept environment id in its params [INTEG-1492]

ryunsong-contentful:
Somewhat recently there was a CEP for updating the upload-api that was approved and a subsequent PR that was merged. This effectively means the Upload API now takes in both the Environment Id and Space Id for a given request.

### DIFF
--- a/lib/adapters/REST/endpoints/upload.ts
+++ b/lib/adapters/REST/endpoints/upload.ts
@@ -1,13 +1,25 @@
 import type { AxiosInstance } from 'contentful-sdk-core'
 import { Stream } from 'stream'
-import { GetSpaceParams } from '../../../common-types'
+import { GetSpaceEnvironmentParams, GetSpaceEnvironmentUploadParams } from '../../../common-types'
 import { getUploadHttpClient } from '../../../upload-http-client'
 import { RestEndpoint } from '../types'
 import * as raw from './raw'
 
+const getBaseUploadUrl = (params: GetSpaceEnvironmentParams) => {
+  const spacePath = `/spaces/${params.spaceId}/uploads`
+  const environmentPath = `/spaces/${params.spaceId}/environments/${params.environmentId}/uploads`
+  const path = params.environmentId ? environmentPath : spacePath
+  return path
+}
+
+const getEntityUploadUrl = (params: GetSpaceEnvironmentUploadParams) => {
+  const path = getBaseUploadUrl(params)
+  return path + `/${params.uploadId}`
+}
+
 export const create: RestEndpoint<'Upload', 'create'> = (
   http: AxiosInstance,
-  params: GetSpaceParams,
+  params: GetSpaceEnvironmentParams,
   data: { file: string | ArrayBuffer | Stream }
 ) => {
   const httpUpload = getUploadHttpClient(http)
@@ -16,7 +28,8 @@ export const create: RestEndpoint<'Upload', 'create'> = (
   if (!file) {
     return Promise.reject(new Error('Unable to locate a file to upload.'))
   }
-  return raw.post(httpUpload, `/spaces/${params.spaceId}/uploads`, file, {
+  const path = getBaseUploadUrl(params)
+  return raw.post(httpUpload, path, file, {
     headers: {
       'Content-Type': 'application/octet-stream',
     },
@@ -25,18 +38,18 @@ export const create: RestEndpoint<'Upload', 'create'> = (
 
 export const del: RestEndpoint<'Upload', 'delete'> = (
   http: AxiosInstance,
-  params: GetSpaceParams & { uploadId: string }
+  params: GetSpaceEnvironmentUploadParams
 ) => {
   const httpUpload = getUploadHttpClient(http)
-
-  return raw.del(httpUpload, `/spaces/${params.spaceId}/uploads/${params.uploadId}`)
+  const path = getEntityUploadUrl(params)
+  return raw.del(httpUpload, path)
 }
 
 export const get: RestEndpoint<'Upload', 'get'> = (
   http: AxiosInstance,
-  params: GetSpaceParams & { uploadId: string }
+  params: GetSpaceEnvironmentUploadParams
 ) => {
   const httpUpload = getUploadHttpClient(http)
-
-  return raw.get(httpUpload, `/spaces/${params.spaceId}/uploads/${params.uploadId}`)
+  const path = getEntityUploadUrl(params)
+  return raw.get(httpUpload, path)
 }
--- a/lib/common-types.ts
+++ b/lib/common-types.ts
@@ -1668,13 +1668,13 @@ export type MRActions = {
     update: { params: GetUIConfigParams; payload: UIConfigProps; return: UIConfigProps }
   }
   Upload: {
-    get: { params: GetSpaceParams & { uploadId: string }; return: any }
+    get: { params: GetSpaceEnvironmentUploadParams; return: any }
     create: {
-      params: GetSpaceParams
+      params: GetSpaceEnvironmentParams
       payload: { file: string | ArrayBuffer | Stream }
       return: any
     }
-    delete: { params: GetSpaceParams & { uploadId: string }; return: any }
+    delete: { params: GetSpaceEnvironmentUploadParams; return: any }
   }
   Usage: {
     getManyForSpace: {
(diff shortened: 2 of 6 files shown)


lib/create-environment-api.ts
```diff
@@ -1152,11 +1152,13 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
      */
     createUpload: function createUpload(data: { file: string | ArrayBuffer | Stream }) {
       const raw = this.toPlainObject() as EnvironmentProps
+      const envId = raw.sys.aliasedEnvironment?.sys.id || ''
```
ruderngespra, line 1155: I think this is not targeting the public environment id. As far as I understand the code this either:

1. in the case that we are looking at an alias route, using the underlying environment id

2. in the case that no aliases are used, just providing an empty string.

Not sure about all the details of how the upload endpoint has been adjusted, but I am assuming it will always use either the alias-id or the environment-id, but never the environment-id that is being targeted by the alias, not empty string.

ruderngespra (changes requested):
No expert on the changes that have been introduced to the upload endpoint, so happy to discuss this further async as well :)

lib/adapters/REST/endpoints/upload.ts
```diff
@@ -25,18 +28,22 @@ export const create: RestEndpoint<'Upload', 'create'> = (
...
 
 export const get: RestEndpoint<'Upload', 'get'> = (
   http: AxiosInstance,
-  params: GetSpaceParams & { uploadId: string }
+  params: GetSpaceEnvironmentParams & { uploadId: string }
 ) => {
   const httpUpload = getUploadHttpClient(http)
-
-  return raw.get(httpUpload, `/spaces/${params.spaceId}/uploads/${params.uploadId}`)
+  const spacePath = `/spaces/${params.spaceId}/uploads/${params.uploadId}`
```
jsdalton, line 45: This is the third repetition of the same code and I think it would not be premature to abstract it into a helper function (eg in this file)

lib/common-types.ts
```diff
@@ -1668,13 +1668,13 @@ export type MRActions = {
     update: { params: GetUIConfigParams; payload: UIConfigProps; return: UIConfigProps }
   }
   Upload: {
-    get: { params: GetSpaceParams & { uploadId: string }; return: any }
+    get: { params: GetSpaceEnvironmentParams & { uploadId: string }; return: any }
```
jsdalton, line 1671: `GetSpaceEnvironmentParams & { uploadId: string }`

we repeat this string everywhere and it represents the fully upload params. why not give it a named type instead of repeating the union type everywhere?

eg `UploadParams`

ryunsong-contentful: Done -  created this instead 
```
export type GetSpaceEnvironmentUploadParams = {
  spaceId: string
  environmentId: string
  uploadId?: string
}
```

ruderngespra: We don't want the required ` & {uploadId: string}}` here anymore then, right?

lib/adapters/REST/endpoints/upload.ts
```diff
@@ -1,13 +1,23 @@
...
 import { RestEndpoint } from '../types'
 import * as raw from './raw'
 
+const getUploadUrlPath = (params: GetSpaceEnvironmentUploadParams) => {
+  const spacePath = `/spaces/${params.spaceId}/uploads`
+  const environmentPath = `/spaces/${params.spaceId}/environments/${params.environmentId}/uploads`
+  const path = params.environmentId ? environmentPath : spacePath
+
+  // Return path as is if uploadId is not provided, otherwise append it to the path
+  if (!params.uploadId) return path
```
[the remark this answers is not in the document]
ryunsong-contentful: Hmm... I don't think so. The path for the post route would look like `/spaces/${params.spaceId}/uploads` OR like `/spaces/${params.spaceId}/environments/${params.environmentId}/uploads`. So no asset ids, or upload ids in the route being added here. This is basically just a refactoring of the path logic since the post doesn't need an upload id but get and deletes do. 

The asset itself is passed as a function parameter like this
```
  const path = getUploadUrlPath(params)
  return raw.post(httpUpload, path, file, {
    headers: {
      'Content-Type': 'application/octet-stream',
    },
  })
}
```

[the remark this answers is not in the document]
ruderngespra: 👍🏻 But then we shouldn't add the `uploadId` to the path here, right?

lib/common-types.ts
ryunsong-contentful: I think we do because for gets and deletes, the uploadId is required. I only made this type for the generalized `getUploadUrlPath` because that creates the path for post/get/delete and the post doesn't need the uploadId but get/delete does so I made it optional to make the function work for all three http action types

lib/adapters/REST/endpoints/upload.ts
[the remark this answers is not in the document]
ryunsong-contentful: Correct, `const path = getUploadUrlPath(params)` will ensure that the path will contain the upload id for get/delete but not for the create operation.

test/integration/asset-integration.js
```diff
@@ -99,7 +99,8 @@ describe('Asset api', function () {
       await unarchivedAsset.delete()
     })
 
-    test('Create and process asset with multiple locales', async () => {
+    // Skip because this is a flakey test
```
ryunsong-contentful, line 102: These are something the DX team will fix once they take ownership of this repo

lib/adapters/REST/endpoints/upload.ts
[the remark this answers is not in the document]
ruderngespra: Got it. 👍🏻 A bit of a nit but I think I would prefer to have this reflected in the types of the `create` [method](https://github.com/contentful/contentful-management.js/blob/018840f2b6d07ad1f312964beb82319f889e0dc1/lib/adapters/REST/endpoints/upload.ts#L20C11-L20C42) then, basically the `create` method then should not have `GetSpaceEnvironmentUploadParams` as type, because it will never accept the optional `uploadId` (or if it did, it would result in a 400 api call), but only `GetSpaceEnvironmentParams` right?

lib/adapters/REST/endpoints/upload.ts
```diff
@@ -1,13 +1,23 @@
...
+  return path + `/${params.uploadId}`
+}
+
 export const create: RestEndpoint<'Upload', 'create'> = (
   http: AxiosInstance,
-  params: GetSpaceParams,
+  params: GetSpaceEnvironmentUploadParams,
```
ruderngespra, line 20: Basically this should then be `GetSpaceEnvironmentParams` right?

lib/adapters/REST/endpoints/upload.ts
```diff
@@ -1,13 +1,23 @@
...
-import { GetSpaceParams } from '../../../common-types'
+import { GetSpaceEnvironmentUploadParams } from '../../../common-types'
 import { getUploadHttpClient } from '../../../upload-http-client'
 import { RestEndpoint } from '../types'
 import * as raw from './raw'
 
+const getUploadUrlPath = (params: GetSpaceEnvironmentUploadParams) => {
+  const spacePath = `/spaces/${params.spaceId}/uploads`
+  const environmentPath = `/spaces/${params.spaceId}/environments/${params.environmentId}/uploads`
+  const path = params.environmentId ? environmentPath : spacePath
```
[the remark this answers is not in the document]
ruderngespra: Discusses this async and decided to go for a save route to have this as fallback even though we never expect it to happen.

ruderngespra (approved):
lgtm 🎉 
I would double check this with the owners of the upload api as well though!

lib/adapters/REST/endpoints/upload.ts
ryunsong-contentful: Yes you are right!

lib/adapters/REST/endpoints/upload.ts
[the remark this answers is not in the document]
ryunsong-contentful: Done, I agree, the create method should take in only the GetSpaceEnvironmentParams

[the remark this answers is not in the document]
jsdalton: Yeah I have a few cents to add here as well... I think the comments and the suggestions on the approach make a lot of sense.

My opinion is that we should not overload the method here to handle both uses cases but treat them as different things. This also makes the type definition itself a bit cleaner.

So to give you an example picked at random from the same folder we're in here (REST/endpoints): https://github.com/contentful/contentful-management.js/blob/cebe70ac8963204e8341506eb6c93e1e4d1874d3/lib/adapters/REST/endpoints/organization-membership.ts#L15-L19

In general this is true of every "restful" resource, there will always be two endpoints the "single" (for read/update/delete) and the "collection" (for create/query). So I suggest we embrace it, create two methods which have different input types, and then update `GetSpaceEnvironmentUploadParams` type to require the uploadid (so it can be reused throughout).

I think with this small change you might even end up with less lines of code, certainly one less conditional, and hopefully a cleaner solution.

Pretty small thing at the end of the day but i think worth getting to a nice clean state.

lib/common-types.ts
jsdalton: Per my comment we can remove the optional uploadId and bring this change back in (ie to use the main `GetSpaceEnvironmentUploadParams` type)

jsdalton (changes requested):
I had some more thoughts in this thread of discussion around the path method and types, LMK if makes sense!

jsdalton (approved):
Thanks for the changes, look great! 😎

PR #2035, opened 2023-11-09
fix: add test step for relase version [NONE]

marcolink:
## Summary
Since version `10.27.1` we are not baking in the right version anymore. The problem must have been introduced with [these](https://github.com/contentful/contentful-management.js/compare/v10.27.0...v10.27.1) changes.  

## Description

With the first commit, we added a new test script `npm run test:version` to check for the occurrence of `0.0.0-determined-by-semantic-release` in the build output ([example run](https://app.circleci.com/pipelines/github/contentful/contentful-management.js/8958/workflows/23ff7b85-1afd-4c01-a2e3-4468bbd4800a/jobs/27465)) 

In a second step, we execute the build step and the version test **AFTER** `npm` already set the right version in the package.json be using npm's the lifecycle hook `prepublishOnly`.  (similar to what we do for [contentful.js](https://github.com/contentful/contentful.js/blob/master/package.json#L67))

In the end - we will only know if this works AFTER this PR is merged :) 

## Motivation and Context
We found that out the the client doesn't know its version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,6 @@ jobs:
     steps:
       - with-workspace
       - run: npm run build
-      - run: npm run test:version
 
   lint:
     executor: docker-with-node
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,7 @@ jobs:
     steps:
       - with-workspace
       - run: npm run build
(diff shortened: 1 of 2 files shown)


ruderngespra (approved):
Hope this does the job 🤞🏻 :)

PR #2051, opened 2023-11-15
chore: add new param types to the plain client for upload [INTEG-1492]

ryunsong-contentful:
Here's the other PR that got merged in to update the upload create,get,delete operations to accept envIds. We forgot to do the plain client which is what this PR fixes https://github.com/contentful/contentful-management.js/pull/2030/files


jjolton-contentful (approved):
🟢

Issue #2054, opened 2023-11-20
Upgrade axios to remove all CVE vulnerabilities


kyrielia:
Axios has a number of CVEs associated with it:
https://security.snyk.io/package/npm/axios

Contentful-management is currently using v1.4.x:
https://github.com/contentful/contentful-management.js/blob/master/package.json#L59

Axios should be updated to the latest version as soon as possible to remove all vulnerability issues.  Ideally this should be released as soon as possible.

t-col:
fixed in `contentful-management.js` version 11.5.7 and `contentful-migration` 4.18.0, thank you for reporting @kyrielia

PR #2055, opened 2023-11-20
[fix] Update axios to 1.6.2 to remove vulnerabilities

kyrielia:
## Summary

See: https://github.com/contentful/contentful-management.js/issues/2054

## Description

Upgrade axios to latest available version

## Motivation and Context

Using contentful-management introduces a vulnerable transitive dependency into any user that uses this library. This fix will mitigate this.


t-col (approved):
thanks again for your contribution @kyrielia !

PR #2082, opened 2023-12-04
fix: release axios upgrade

t-col:
## Summary

release axios vulnerability fix: https://github.com/contentful/contentful-management.js/pull/2055


mgoudy91 (approved):
wow!

Issue #2075, opened 2023-12-01
v11.5.2 broke asset processing (AssetProcessingTimeout)


restfulhead:
We wanted to upgrade to the latest version, but noticed that processing assets after the upload consistently fails with the following error:  `AssetProcessingTimeout - "Asset is taking longer then expected to process."`

I went through the release notes and noticed that there was a change regarding uploads mentioned in `v11.5.2`. So I upgraded to the previous version `11.5.1` and everything works as usual. So something in `11.5.2` must be the cause of the AssetProcessingTimeout error.

Here's the code that I'm using to upload and then process assets:

```
const assetResult = await this.client.asset
        .createFromFiles(
          {
            spaceId: this.cmsSpace,
            environmentId: this.cmsEnv,
          },
          {
            fields: {
              title: {
                [language]: fileName,
              },
              description: {
                [language]: '',
              },
              file: {
                [language]: {
                  contentType: fileType,
                  fileName,
                  file: data,
                },
              },
            },
          }
        )

      const assetProcessed = await this.client.asset.processForAllLocales(
              {
                spaceId: this.cmsSpace,
                environmentId: this.cmsEnv,
              },
              asset
            )
```

theochp:
I have been stuck on this for a few days, thank you for creating the issue.

t-col:
hi @restfulhead thank you reporting. We reverted the problematic commit and released that fix in `v11.5.8`. We will reintroduce the asset processing changes from the original commit at a later time.

jsdalton:
@restfulhead @theochp just wanted to give you both a heads up that we have shipped the API change again. further investigation revealed the original issue you both experienced was a backend issue and not related to the the API change.

if you do notice this issue (or something similar) please do feel free to open a new issue or comment here.

thanks for the original report and sorry about the difficulties here.

PR #2088, opened 2023-12-08
fix: environment-scoped uploads

jsdalton:
## Summary

(Re)introduce environment scoped uploads.

## Description

This PR restores environment scope uploads that were introduced here https://github.com/contentful/contentful-management.js/pull/2030 and reverted here https://github.com/contentful/contentful-management.js/pull/2084 

The issue with the original PR was raised here https://github.com/contentful/contentful-management.js/issues/2075 . Further investigation revealed that the problem was in fact an issue with the backend asset processing service. This service was missing a key piece of logic to make it aware that the upload passed for processing could be in a specific environment instead of at the space level. The SDK change revealed that backend issue when it exposed environment scoping to uploads.

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/asset.ts
+++ b/lib/adapters/REST/endpoints/asset.ts
@@ -154,7 +154,7 @@ export const createWithId: RestEndpoint<'Asset', 'createWithId'> = (
   )
 }
 
-export const createFromFiles: RestEndpoint<'Asset', 'createFromFiles'> = (
+export const createFromFiles: RestEndpoint<'Asset', 'createFromFiles'> = async (
   http: AxiosInstance,
   params: GetSpaceEnvironmentParams & { uploadTimeout?: number },
   data: Omit<AssetFileProp, 'sys'>
@@ -163,7 +163,7 @@ export const createFromFiles: RestEndpoint<'Asset', 'createFromFiles'> = (
 
   const { file } = data.fields
   return Promise.all(
-    Object.keys(file).map((locale) => {
+    Object.keys(file).map(async (locale) => {
       const { contentType, fileName } = file[locale]
 
       return createUpload(httpUpload, params, file[locale]).then((upload) => {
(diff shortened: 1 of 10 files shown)


ryunsong-contentful (approved):
Is there anything with uploadFrom we need to also adjust before merging this? Or is that fixed now with Richard's PR?

mgoudy91 (approved):

lib/adapters/REST/endpoints/asset.ts
```diff
@@ -204,7 +204,7 @@ export const createFromFiles: RestEndpoint<'Asset', 'createFromFiles'> = (
 const ASSET_PROCESSING_CHECK_WAIT = 3000
 const ASSET_PROCESSING_CHECK_RETRIES = 10
 
-function checkIfAssetHasUrl(
+async function checkIfAssetHasUrl(
```
RichardM99, line 207: Why this change (and in some of the other functions) considering there is no `await`?

RichardM99 (approved):
🥳 

Small note - when I was looking at the customer ticket you mentioned in the [slack thread](https://contentful.slack.com/archives/C059CFR8ZLN/p1702051184713559) I was unable to use the `createFromFiles` method on the SDK without passing the env id, which I think is a bug since it should be possible to use it with space id only.

Not sure if that's due to the changes as part of this PR or not.

jsdalton:
> Small note - when I was looking at the customer ticket you mentioned in the [slack thread](https://contentful.slack.com/archives/C059CFR8ZLN/p1702051184713559) I was unable to use the `createFromFiles` method on the SDK without passing the env id, which I think is a bug since it should be possible to use it with space id only.

@RichardM99  gotcha, but just as a note the type definition requiring the environment id has long existed in this library: https://github.com/contentful/contentful-management.js/blob/master/lib/plain/common-types.ts#L504

so if we want/need to change this, my assumption is we should do it in a separate PR i.e. unrelated to this PR.

PR #2115, opened 2024-01-08
feat: add `getRoles` for organization [CFCX-1226]

andipaetzold:
## Summary

Add method to query all roles of an organization (`/organization/:orgId/roles`)

## Description

Plain: `client.roles.getManyForOrganization(...)`
Nested: `organization.getRoles(...)`

When adding a new method:


mgoudy91 (approved):
Looks good to me.

What's the use case for this out of curiosity?

andipaetzold:
> What's the use case for this out of curiosity?

@mgoudy91 It's used in the Web App. The endpoint already existed for a while, it was just not implemented in CMA.js.

PR #2130, opened 2024-01-17
feat: add `getSpaces` for organization

andipaetzold:
## Summary

Adds `client.spaces.getManyForOrganization` (plain) and `organization.getSpaces()` (nested)

Required for requests in the Web App

## Motivation and Context

Currently, only the `/spaces` endpoint is implemented. It's not possible to query the spaces of a specific org.

When adding a new method:


lib/create-organization-api.ts
```diff
@@ -79,7 +108,7 @@ export default function createOrganizationApi(makeRequest: MakeRequest) {
      *
      * client.getOrganization('<organization_id>')
      * .then((organization) => organization.getUsers())
-     * .then((user) => console.log(user))
```
stephanLeece, line 82: Little bonus fix? :)

stephanLeece (approved):
Noice

PR #2134, opened 2024-01-19
feat: add support for GET published asset endpoint []

chrishelgert:
## Summary

The GET endpoint for published assets was not yet added to the client.

## Motivation and Context

https://www.contentful.com/developers/docs/references/content-management-api/#/reference/assets/published-assets-collection (JavaScript)

When adding a new method:


colomolo (approved):
Looks great to me. Do we need it only for async locales at the moment?

PR #2138, opened 2024-01-24
fix: renames delivery function to function [MONET-1720]

RichardM99:
## Summary

- This change aligns the code base with the latest naming conventions. 


marcolink:
Is this a breaking change? If so, we could also deprecated the old one and flag it, and introduce the new name.

RichardM99:
@marcolink 

Delivery functions still in EAP so I think deprecating the existing entity is not necessary.

Vinz93 (approved):

cemreyuksel (approved):
We need to update `user-interface` as well.

PR #2139, opened 2024-01-24
fix: renames function to contentful function [MONET-1720]

RichardM99:
## Summary

- @marcolink commented that the current naming may cause some issues if one wants to deconstruct the prop due to `function` being a reserved keyword.
- As such, using the alternative naming of `ContentfulFunction` (the `sys.type` still remains as `Function`).


cemreyuksel (approved):
👍

marcolink (approved):

RichardM99:
In the end it was decided that the current naming is what we want to go with.

PR #2161, opened 2024-02-02
chore(axios): bump to latest

cf-remylenoir:
## Summary

Upgrades `axios` to the latest available version in the lock file to prevent sniffers from reporting a security vulnerability fixed in `1.6.4`.

## Motivation and Context

Axios `1.6.4` release: https://github.com/axios/axios/releases/tag/v1.6.4

Customer signal with sniffer report (Snyk): 
Screenshot 2024-02-01 at 9 09 57 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@contentful/rich-text-types": "^16.3.0",
         "@types/json-patch": "0.0.30",
-        "axios": "^1.6.2",
+        "axios": "^1.6.7",
         "contentful-sdk-core": "^8.1.0",
         "fast-copy": "^3.0.0",
         "lodash.isplainobject": "^4.0.6",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -11,7 +11,7 @@
       "dependencies": {
         "@contentful/rich-text-types": "^16.3.0",
         "@types/json-patch": "0.0.30",
-        "axios": "^1.6.2",
+        "axios": "^1.6.7",
```
andipaetzold, line 14: This line should be unchanged, right?

andipaetzold (changes requested):
We should not merge as `build:` as that would release a new version. Please use `chore:` instead

package-lock.json
cf-remylenoir, line 14:
```suggestion
        "axios": "^1.6.2",
```

cf-remylenoir: You are right, I forgot to revert this line.

package-lock.json
```diff
@@ -5087,9 +5087,9 @@
       }
     },
     "node_modules/axios": {
-      "version": "1.6.5",
```
denkristoffer, line 5090: I don't think I understand, the old version is already newer than the one with the fix. What am I missing? 😄

cf-remylenoir:
Pausing, for now. I asked the Solution Architect who shared this customer signal with us when the customer got the Snyk audit (since the package already resolves to `axios 1.6.5`).

cf-remylenoir:
Also, duplicate of #2143 but the integration test is failing.

andipaetzold:
> Also, duplicate of #2143 but the integration test is failing.

I think we can still merge as it's a flaky test

t-col:
looks like the linked dependabot PR got merged, so we can close this one

PR #2170, opened 2024-02-10
chore: add 'access' to Roles `ActionType` type

sunehinz:
## Summary

Add missing `access` attribute to Roles ActionType type.

## Description

The `access` attribute on Roles ActionType type is missing. The API support this, and returns policies with this value. The issue is solely a Typescript type issue.

## Motivation and Context

Removes Typescript errors and adds `access` to intellisense.

When adding a new method:


mgoudy91 (approved):
good catch, thanks for the PR!

PR #2171, opened 2024-02-12
feat: add accepts as an optional parameter [MONET-1825]

cemreyuksel:
[Ticket](https://contentful.atlassian.net/browse/MONET-1825)

## Summary

This PR adds `accepts` as an optional parameter to `appBundle` and `function` props.

### DIFF
--- a/lib/entities/function.ts
+++ b/lib/entities/function.ts
@@ -14,6 +14,6 @@ export type FunctionProps = {
   name: string
   description: string
   path: string
+  accepts: string[]
   allowNetworks?: string[]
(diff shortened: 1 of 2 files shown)


lib/entities/function.ts
```diff
@@ -14,5 +14,6 @@ export type FunctionProps = {
   name: string
   description: string
   path: string
+  accepts: string[]
```
mrlaessig, line 17: should this be optional as well?

```suggestion
  accepts?: string[]
```

mrlaessig (approved):

PR #2179, opened 2024-02-20
feat: PlainClient docs - user [EXT-4960]

david-shibley-contentful:

When adding a new method:

### DIFF
--- a/lib/plain/entities/user.ts
+++ b/lib/plain/entities/user.ts
@@ -1,0 +1,84 @@
+import {
+  CollectionProp,
+  GetOrganizationParams,
+  GetSpaceParams,
+  QueryParams,
+} from '../../common-types'
+import { UserProps } from '../../export-types'
+import { OptionalDefaults } from '../wrappers/wrap'
+
+export type UserPlainClientAPI = {
+  /** Fetches all users in a space
+   *
+   * @param params space ID and query parameters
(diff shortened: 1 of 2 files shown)


ghepting (approved):
nice

lib/plain/entities/user.ts
```diff
@@ -0,0 +1,84 @@
...
+  /** Fetches all users in a space
+   *
+   * @param params space ID and query parameters
+   * @returns a collection of users
+   * @throws if the request fails or the users are not found
+   * @example
+   * ```javascript
+   * const users = await client.user.getManyForSpace({
+   *  spaceId: '<space_id>',
+   *  limit: 100,
```
jjolton-contentful, line 20: this goes in a query object no?

lib/plain/entities/user.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Fetches all users in an organization
+   * @param params organization ID and query parameters
+   * @returns a collection of users
+   * @throws if the request fails or the users are not found
+   * @example
+   * ```javascript
+   * const users = await client.user.getManyForOrganization({
+   *   organizationId: '<organization_id>',
+   *   limit: 100,
```
jjolton-contentful, line 77: same with here

PR #2181, opened 2024-02-20
feat: PlainClient docs - task [EXT-4924]

david-shibley-contentful:

When adding a new method:

### DIFF
--- a/lib/plain/plain-client-test.ts
+++ b/lib/plain/plain-client-test.ts
@@ -162,13 +162,13 @@ describe('Plain Client', () => {
         ],
       }
 
-      const props: Props = {
+      const props = {
         ...defaultProps,
         bodyFormat: 'rich-text',
-      }
+      } as any
 
       it('should create a get object', async () => {
-        await plainClient.comment.get(props as any)
+        await plainClient.comment.get(props)
         expect(stub).to.have.been.calledWithMatch({
           entityType: 'Comment',
           action: 'get',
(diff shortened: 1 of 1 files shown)


lib/plain/entities/task.ts
```diff
@@ -0,0 +1,114 @@
...
+   * @param params Space ID, Entry ID, Environment ID, and query parameters
+   * @returns a collection of tasks
+   * @throws if the request fails or the tasks are not found
+   * @example
+   * ```javascript
+   * const tasks = await client.task.getMany({
+   *  spaceId: '<space_id>',
+   *  entryId: '<entry_id>',
+   *  environmentId: '<environment_id>',
+   *  limit: 100,
```
jjolton-contentful, line 45: I thought this gets wrapped in a "query" object

lib/plain/entities/task.ts
```diff
@@ -0,0 +1,114 @@
...
+   * ```javascript
+   * const task = await client.task.update(
+   *  {
+   *    spaceId: '<space_id>',
+   *    entryId: '<entry_id>',
+   *    environmentId: '<environment_id>',
+   *    taskId: '<task_id>',
+   *  },
+   *  {
+   *    title: 'My Task',
```
jjolton-contentful, line 91: where does title come from? might consider using the example body they have here: https://www.contentful.com/developers/docs/references/content-management-api/#/reference/entry-tasks/task/update-a-task/console/js

lib/plain/entities/task.ts
david-shibley-contentful: excellent point

jjolton-contentful (approved):

PR #2196, opened 2024-02-26
docs: add inline docs for the plain client Team interfaces [EXT-4961]

jjolton-contentful:
… 

## Summary
adds inline documentation for the Team, TeamMembership, and TeamSpaceMembership plain client interfaces

## Description

When adding a new method:


t-col (approved):
:shipit:

PR #2197, opened 2024-02-26
feat: Plain client docs - usage [EXT-4956]

david-shibley-contentful:

When adding a new method:

### DIFF
--- a/lib/plain/entities/usage.ts
+++ b/lib/plain/entities/usage.ts
@@ -1,0 +1,42 @@
+import { CollectionProp, QueryParams } from '../../common-types'
+import { UsageProps } from '../../export-types'
+import { OptionalDefaults } from '../wrappers/wrap'
+
+export type UsagePlainClientAPI = {
+  /** Fetches all usage by space
+   *
+   * @param params organization ID and query parameters
+   * @returns a collection of usage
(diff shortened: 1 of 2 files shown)


lib/plain/entities/usage.ts
```diff
@@ -0,0 +1,42 @@
...
+   * });
+   * ```
+   */
+  getManyForSpace(
+    params: OptionalDefaults<{ organizationId: string } & QueryParams>
+  ): Promise<CollectionProp<UsageProps>>
+  /** Fetches all usage for a given organization
+   *
+   * @param params organization ID and query parameters
+   * @returns a collection of usage
```
t-col, line 27: can we rephrase this pls

lib/plain/entities/usage.ts
```diff
@@ -0,0 +1,42 @@
...
+   * @throws if the request fails or the usage is not found
```
t-col, line 28: same, `the usage` doesn't really make sense

t-col (changes requested):
can we change the `the usage` verbiage here to make sense for what the API is for: https://www.contentful.com/developers/docs/references/content-management-api/#/reference/usage

t-col (approved):

PR #2202, opened 2024-02-28
feat: add client methods for AppAccessTokens [EXT-4754]

jjolton-contentful:
## Summary

Adds the `create` client method for AppAccessTokens

## Description

CMA JS client <-> API feature parity

When adding a new method:


mgoudy91 (approved):
solid to me

PR #2279, opened 2024-05-16
fix(entities/field-type): correct casing of resourcelink array item types

cysp:
## Summary

Correct typing of arrays containing resource link items

## Description

The typings exposed for arrays containing resource link items appear to be incorrect, using the casing `Resourcelink` where all other documentation and examples capitalise the `l` like `ResourceLink`

When adding a new method:


mgoudy91 (approved):
Great catch, thanks for the contribution!

PR #2315, opened 2024-06-12
feat: taxonomy endpoints [GROOT-1273]

momenzadehvida:
Adds all endpoints for the new `Concept` and `ConceptScheme` entity

## Documentation
- [Overview](https://www.contentful.com/help/taxonomy-manager/#concepts)
- [Taxonomy Concept](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/taxonomy/concept)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,5 +134,4 @@ workflows:
             branches:
               only:
                 - master
-#                - next
-#                - next-major
+                - beta
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,8 +11,8 @@
       "dependencies": {
         "@contentful/rich-text-types": "^16.3.0",
         "@types/json-patch": "0.0.30",
-        "axios": "~1.6.8",
-        "contentful-sdk-core": "^8.1.0",
+        "axios": "^1.7.2",
+        "contentful-sdk-core": "^8.3.1",
         "fast-copy": "^3.0.0",
         "lodash.isplainobject": "^4.0.6",
         "type-fest": "^4.0.0"
@@ -4805,9 +4805,9 @@
       }
     },
     "node_modules/axios": {
-      "version": "1.6.8",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.6.8.tgz",
-      "integrity": "sha512-v/ZHtJDU39mDpyBoFVkETcd/uNdxrWRrg3bKpOKzXFA6Bvqopts6ALSMU3y6ijYxbw2B+wPrIv46egTzJXCLGQ==",
+      "version": "1.7.2",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-1.7.2.tgz",
+      "integrity": "sha512-2A8QhOMrbomlDuiLeK9XibIBzuHeRcqqNOHp0Cyp5EoJ1IFDh+XZH3A6BkXtv0K4gFGCI0Y4BM7B1wOEi0Rmgw==",
       "dependencies": {
         "follow-redirects": "^1.15.6",
         "form-data": "^4.0.0",
@@ -5730,12 +5730,18 @@
       }
     },
     "node_modules/call-bind": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/call-bind/-/call-bind-1.0.2.tgz",
-      "integrity": "sha512-7O+FbCihrB5WGbFYesctwmTKae6rOiIzmz1icreWJ+0aA7LJfuqhEso2T9ncpcFtzMQtzXf2QGGueWJGTYsqrA==",
+      "version": "1.0.7",
+      "resolved": "https://registry.npmjs.org/call-bind/-/call-bind-1.0.7.tgz",
+      "integrity": "sha512-GHTSNSYICQ7scH7sZ+M2rFopRoLh8t2bLSW6BbgrtLsahOIB5iyAVJf9GjWK3cYTDaMj4XdBpM1cA6pIS0Kv2w==",
       "dependencies": {
-        "function-bind": "^1.1.1",
-        "get-intrinsic": "^1.0.2"
+        "es-define-property": "^1.0.0",
+        "es-errors": "^1.3.0",
+        "function-bind": "^1.1.2",
+        "get-intrinsic": "^1.2.4",
+        "set-function-length": "^1.2.1"
+      },
+      "engines": {
+        "node": ">= 0.4"
       },
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
@@ -6934,11 +6940,11 @@
       "integrity": "sha512-ozrGwyuCTAy7YgFCua8rmqmytECYk/JYAMXcswOcm0qvGoE3tPb7ivBeIHTOK2DiapBhDZgacIhzhQIKU5TCfA=="
     },
     "node_modules/contentful-sdk-core/node_modules/qs": {
-      "version": "6.11.2",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.11.2.tgz",
-      "integrity": "sha512-tDNIz22aBzCDxLtVH++VnTfzxlfeK5CbqohpSqpJgj1Wg/cQbStNAz3NuqCs5vV+pjBsK4x4pN9HlVh7rcYRiA==",
+      "version": "6.12.2",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.12.2.tgz",
+      "integrity": "sha512-x+NLUpx9SYrcwXtX7ob1gnkSems4i/mGZX5SlYxwIau6RrUSODO89TR/XDGGpn5RPWSYIB+aSfuSlV5+CmbTBg==",
       "dependencies": {
-        "side-channel": "^1.0.4"
+        "side-channel": "^1.0.6"
       },
       "engines": {
         "node": ">=0.6"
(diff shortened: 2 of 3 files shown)


jjolton-contentful (dismissed):
awesome 👏 I know lots of people will be excited to see these go out!

lib/entities/concept-scheme.ts
```diff
@@ -0,0 +1,29 @@
+import { LocalizedEntity } from './utils'
+import { TaxonomyConceptLink } from './concept'
+
+export type ConceptScheme = {
+  uri: string | null
+  prefLabel: string
+  altLabels: string[]
+  hiddenLabels: string[]
+  definition: string | null
+  notations: string[]
```
maxcheremisin, line 10:
```suggestion
  definition: string | null
```

lib/entities/concept-scheme.ts
```diff
@@ -0,0 +1,29 @@
...
+    createdAt: string
+    updatedAt: string
+    id: string
+    version: number
+  }
+}
+
+export type ConceptSchemeProps<Locales extends keyof any = 'en-US'> = LocalizedEntity<
+  ConceptScheme,
+  'prefLabel' | 'altLabels' | 'hiddenLabels' | 'definition',
```
maxcheremisin, line 25:
```suggestion
  'prefLabel' | 'definition',
```

lib/entities/concept-scheme.ts
```diff
@@ -0,0 +1,29 @@
...
+  sys: {
+    type: 'TaxonomyConceptScheme'
+    createdAt: string
+    updatedAt: string
+    id: string
+    version: number
+  }
+}
+
+export type ConceptSchemeProps<Locales extends keyof any = 'en-US'> = LocalizedEntity<
```
maxcheremisin, line 23: with localization in mind, does it make sense to specify `'en-US'` locale? we don't know how this api could look like in the future, maybe customers will create their own default locales. maybe fallback it to just `string`?

lib/entities/utils.ts
```diff
@@ -0,0 +1,33 @@
+import { isNil, isObject, omitBy } from 'lodash'
+
+export type LocalizedEntity<
+  Entity,
+  LocalizedFields extends keyof Entity,
+  Locales extends keyof any
```
maxcheremisin, line 6:
```suggestion
  Locales extends string
```

lib/entities/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+export function toUrlParams(params: Record<string, string | number> | undefined = {}) {
+  const urlQuery = new URLSearchParams()
+
+  Object.entries(sanitizeParams(params)).forEach(([key, value]) => {
+    urlQuery.set(key, `${value}`)
+  })
+
+  const result = urlQuery.toString()
+  return result ? `?${result}` : ''
+}
```
maxcheremisin, line 20: curious shouldn't we already have examples for query params in this repo?

lib/plain/entities/concept-scheme.ts
```diff
@@ -0,0 +1,114 @@
...
+   * @see {@link https://www.contentful.com/developers/docs/references/content-management-api/#/reference/taxonomy/concept-scheme}
+   * @example
+   * ```javascript
+   * const concept = await client.conceptScheme.get({
+   *   organizationId: '<organization_id>',
+   *   conceptSchemeId: '<concept_scheme_id>',
+   * });
+   * ```
+   */
+  get(params: OptionalDefaults<GetConceptSchemeParams>): Promise<ConceptSchemeProps>
```
maxcheremisin, line 65: `OptionalDefaults` resolves to `Record<string, unknown>`, which allows to pass any object

test/unit/adapters/REST/endpoints/concept-scheme-test.ts
```diff
@@ -0,0 +1,153 @@
...
+    return adapterMock
+      .makeRequest({
+        entityType: 'ConceptScheme',
+        action: 'get',
+        params: {
+          organizationId: 'organization-id',
+          conceptSchemeId: 'concept-scheme-id',
+        },
+      })
+      .then((r) => {
```
maxcheremisin, line 27: curious why not `await`

test/unit/adapters/REST/endpoints/concept-scheme-test.ts
```diff
@@ -0,0 +1,153 @@
...
+        expect(r).to.eql(entityMock)
```
maxcheremisin, line 28: not sure what actually this tests

maxcheremisin: i would maybe prefer to see that url and query params are resolved correctly. 

also, since it's a public lib in TS, maybe it's worth testing the TS interface

test/unit/mocks/entities.js
```diff
@@ -771,6 +771,104 @@ const commentMock = {
...
+  },
+  hiddenLabels: {
+    'en-US': [],
+  },
+  note: null,
+  changeNote: null,
+  editorialNote: null,
+  historyNote: null,
+  scopeNote: null,
+  notations: [],
```
maxcheremisin, line 868:
```suggestion
  definition: null,
  concepts: [],
```

lib/entities/utils.ts
marcolink: c&p tricked me, we can handle this more elegant.

lib/entities/concept-scheme.ts
marcolink: it does, will change it!

lib/entities/concept-scheme.ts
marcolink: fixed

lib/entities/concept-scheme.ts
marcolink: fixed

lib/entities/concept-scheme.ts
marcolink: fixed

lib/entities/utils.ts
marcolink: fixed

test/unit/adapters/REST/endpoints/concept-scheme-test.ts
marcolink: c&p :)

lib/plain/entities/concept-scheme.ts
marcolink: @maxcheremisin good catch - we've solved it for the new endpoints introduced in this PR, but something we should look into in general (like fixing the actual util type `OptionalDefaults`

lib/entities/concept-scheme.ts
```diff
@@ -0,0 +1,27 @@
...
+  topConcepts: TaxonomyConceptLink[]
+  concepts: TaxonomyConceptLink[]
+  totalConcepts: number
+  sys: {
+    type: 'TaxonomyConceptScheme'
+    createdAt: string
+    updatedAt: string
+    id: string
+    version: number
+  }
```
maxcheremisin, line 17: same here

test/integration/taxonomy-integration.ts
```diff
@@ -0,0 +1,404 @@
...
+    const result = await client.concept.create({}, concept)
+
+    isConceptProps(result)
+    expect(result.prefLabel['en-US']).to.equal('Test Concept')
+    expect(result.uri).to.null
+
+    await client.concept.delete({
+      conceptId: result.sys.id,
+      version: result.sys.version,
+    })
```
AdrianLThomas, line 57: Is there nothing we want to assert after performing the deletion? probably at minimum a successful status code?

test/integration/taxonomy-integration.ts
```diff
@@ -0,0 +1,404 @@
...
+    organizationId: getTestOrganizationId(),
+  })
+
+  beforeEach(() => {
+    conceptsToDelete = []
+    conceptSchemesToDelete = []
+  })
+
+  afterEach(async () => {
+    for (const concept of conceptsToDelete) {
```
AdrianLThomas, line 21: More of a general question: if/when this fails to trigger (e.g. someone kills the build half way through), what is the impact? is it just debris left in the space or could it cause flakey tests in subsequent test runs?

AdrianLThomas: Also I'm not quite clued up on the test setup: but would it be easier and cleaner to wipe the whole space before/after a test run?

lib/adapters/REST/endpoints/concept.ts
```diff
@@ -0,0 +1,111 @@
...
+      ...headers,
+    },
+  })
+
+export const getMany: RestEndpoint<'Concept', 'getMany'> = (
+  http: AxiosInstance,
+  params: GetManyConceptParams
+) => {
+  const { url, queryParams } = cursorBasedCollection('', params)
+  // console.log({url, queryParams})
```
AdrianLThomas, line 71: ?

AdrianLThomas (commented):
Left some comments. Looking good on the surface though. Would love someone who is more familiar with the repo to take a closer look.

test/integration/taxonomy-integration.ts
momenzadehvida: Maybe you're right, I'm not sure about the thought behind it either.

test/integration/taxonomy-integration.ts
momenzadehvida: I added a test to be sure the item is deleted.

AdrianLThomas: Nice. Do you think we should check the status code for the DELETE too?

lib/common-types.ts
```diff
@@ -1947,6 +2031,33 @@ export type GetWebhookParams = GetSpaceParams & { webhookDefinitionId: string }
...
+export type UpdateConceptParams = GetOrganizationParams & { conceptId: string; version: number }
+export type DeleteConceptParams = GetOrganizationParams & { conceptId: string; version: number }
+export type GetConceptDescendantsParams = GetOrganizationParams & { conceptId: string } & {
+  query?: { depth?: number; pageUrl?: string }
+}
+export type GetManyConceptParams = GetOrganizationParams & {
+  query?:
+    | { pageUrl?: string }
+    | ({ conceptScheme?: string; query?: string } & BasicCursorPaginationOptions &
+        Omit<PaginationQueryOptions, 'skip'>)
```
maxcheremisin, line 2044: doesn't `BasicCursorPaginationOptions` already extend `Omit<PaginationQueryOptions, 'skip'>`?

lib/entities/utils.ts
maxcheremisin: @momenzadehvida what do you think about this?

maxcheremisin (approved):
very nice 🚀

lib/entities/utils.ts
t-col: couple questions:
- can you point me to where this util is being used?
- can we use an off the self solutions with the `qs` lib like this instead? https://github.com/contentful/contentful-sdk-core/blob/0afab0fb02a05eaddd136d0679cdcfaaf448c6f5/src/create-http-client.ts#L100

momenzadehvida: This function is not used, I removed. it looks axios is supporting this. 
https://github.com/axios/axios?tab=readme-ov-file#request-config

jjolton-contentful (approved):

Issue #2334, opened 2024-06-26
☂️ Umbrella: full and modern ESM compatability


axe312ger:
# 🛣️ The road to modern ESM
Working with ESM only dependencies is still hard in 2024. And we want to do our part and make all people of all code styles and environments happy!

# 👟 Core and CDA are ready(ish)
To make this happen, we have to do kind of a leap forward. We already have the [SDK-CORE prepared](https://github.com/contentful/contentful-sdk-core/pull/438) and the [CDA.js is also ready for review](https://github.com/contentful/contentful.js/pull/2169)

# 🔨 List of todos

These are based on  the learnings from our other repos:

* [ ] migrate to latest typescript (handles ESM better - plus a major that has to come anyways)
* [ ] switch to rollup (better ESM handling, faster, all packages use same bundler now)
* [ ] remove node (minified) bundle (no need!)
* [ ] make available in cjs (as a bundle for node without ESM support)
* [ ] make available in pure esm (through a simple tsc compile)
* [ ] support ESM only dependencies (it is time, we are starting to lack behind and create dependency issues with modern setups)
* [ ] compatibility for browser modules and still provide a minified bundle for browser (hopefully used for demos, and not production)
* [ ] switch to vitest - #2328 (less trouble with ESM as with jest. Here we don't even have jest, this still runs on chai+mocha+chai which have (big?) issues with ESM as well)

PR #2335, opened 2024-06-26
build: dependency update that passes all tests

axe312ger:
This aims to update multiple dependencies while passing integration tests, basically covering many dependabot issues at once that currently fail integration tests

As part of https://github.com/contentful/contentful-management.js/issues/2334


marcolink (approved):
Out if curiosity, do we know what doubles the bundle sizes?

axe312ger:
> Out if curiosity, do we know what doubles the bundle sizes?

I think its a combo:

- https://bundlephobia.com/package/axios@1.7.2
- babel core updates / polyfills
- changes to webpack

But its crazy that it doubles the size!

axe312ger:
Or, just a guess, but could also be because of dependency duplication and the current bundle in a bundle situation

axe312ger:
@mgoudy91 I guess ready 4 merge & release :)

mgoudy91 (approved):
👍

PR #2336, opened 2024-06-26
Build: cleanup dependencies

axe312ger:
Some dependencies are obsolete, so lets get rid of them.

Part of #2334


package.json
```diff
@@ -62,11 +62,10 @@
   ],
   "dependencies": {
     "@contentful/rich-text-types": "^16.6.1",
-    "@types/json-patch": "0.0.30",
     "axios": "^1.7.2",
     "contentful-sdk-core": "^8.1.0",
     "fast-copy": "^3.0.0",
-    "lodash.isplainobject": "^4.0.6",
+    "json-patch": "^0.7.0",
```
marcolink, line 68: Are we using any implementation from it, or just types?

marcolink: What about not using it at all and maybe changing the imports🤔

axe312ger: well we support json-patch requests in the API, so we should also provide the correct types. We can fix the imports to use `import type` to make sure we never include these imports outside of TS

marcolink: But all current references are only pointing to the types package. Is the plan to already pull it in because we will use the actual implementation in the future?

axe312ger: I added it because it was a missing dependency which got installed through sth else.

I guess the original author added it so we don't need to maintain our own type for this action. As we need https://json.schemastore.org/json-patch as a TS type :)


I see two options:
* use `import type` and move it to dev dependencies
* add our own TS type for the json-patch request. Maybe the same as the CTF backend uses? 🤔

marcolink: what ever it takes to not maintain our own implementation :)

axe312ger: @marcolink https://github.com/contentful/contentful-management.js/pull/2339

is this worth the changes? 🤔

marcolink: I'd say yes :)  - and since you already did it ...

mgoudy91 (approved):

PR #2328, opened 2024-06-20
Refactor: replace test suite with vitest

axe312ger:
As this is the only public repo with mocha/chai/sinon left, this moves it to the modern jest/vitest syntax.

We did the same for the sdk-core and the contentful.js --> https://github.com/contentful/contentful-sdk-core/pull/438 + https://github.com/contentful/contentful.js/pull/2169

* [x] unit tests with vitest
  * [x] no more memory overflows / timeouts
  * [x] test time down from ~2m 36s to ~56s
* [x] integration tests with vitest
  * [x] no more false negatives / integration tests now pass every time (at lest as long noone else is running them at the same time)
  * [x] test time a little up from ~7m 41s to ~11m (but as they pass all the time, its still a win!)
* [x] run unit and integration tests in browser with vitest (only for manual execution)
* [x] turn all tests into TypeScript
* [x] adjusted mocked data to match TS interfaces
* [x] small cleanups and removed outdated and unused e2e tests

PR #2374, opened 2024-07-24
fix: entity mocks now reflect real API structure

axe312ger:
While rewriting and typing the tests #2328, I found out that a good chunk of our unit test mocks are in wrong shape.

This PR fixes the mocks and is based on my [mock typing commit](https://github.com/contentful/contentful-management.js/pull/2328/commits/f786052a844cffe3aa40a0e67b6477555ae6f280)


axe312ger:
Closing this as its part of #2328

PR #2408, opened 2024-08-16
fix: patch axios to include fetch adapter fix [EXT-5371]

jjolton-contentful:
## Summary

The axios fetch adapter needs this patch to work on Cloudflare workers

## Motivation and Context

see https://github.com/axios/axios/pull/6547


mgoudy91 (approved):
🩹

PR #2415, opened 2024-08-21
feat: add resourceProvider methods [DANTE-1405]

mohamedturki:
## Summary

To give our users and ourselves more convenient access to the new entities: `resourceProviders`, `resourceTypes` and `resources`, we should add them to the contentful-management.js JavaScript library. The library currently contains two separate client implementations that need to be updated: the legacy client and the plain client. This PR adds the first entity, `resourceProvider`, to both clients.

## Description

This PR adds new methods to CRUD resourceProviders:

- Legacy client: 

1. AppDefinition methods to upsert and get resourceProviders
2. ResourceProvider methods to upsert and delete a given resource provider

- Plain client:

1. ResourceProvider methods to get, upsert and delete a given resource provider

When adding a new method:


lib/create-app-definition-api.ts
```diff
@@ -191,5 +193,74 @@ export default function createAppDefinitionApi(makeRequest: MakeRequest) {
...
+     * })
+     *
+     * client.getOrganization('<org_id>')
+     * .then((org) => org.getAppDefinition('<app_def_id>'))
+     * .then((appDefinition) => appDefinition.getResourceProvider())
+     * .then((resourceProvider) => console.log(resourceProvider))
+     * .catch(console.error)
+     * ```
+     */
+    getResourceProvider() {
```
mayakarabula, line 254: wouldn't we want here to have also `deleteResourceProvider`?

mayakarabula (approved):
lgtm

mgoudy91 (approved):

PR #2429, opened 2024-09-02
feat: add Resource Type to contentful-management [DANTE-1832]

mayakarabula:
## Summary

Add Resource Type for Native External References

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/resource-type.ts
+++ b/lib/adapters/REST/endpoints/resource-type.ts
@@ -1,0 +1,35 @@
+import type { RawAxiosRequestHeaders } from 'axios'
+import type { AxiosInstance } from 'contentful-sdk-core'
+import * as raw from './raw'
+import copy from 'fast-copy'
+import type { GetResourceTypeParams } from '../../../common-types'
+import type { RestEndpoint } from '../types'
+import type { ResourceTypeProps, UpsertResourceTypeProps } from '../../../entities/resource-type'
+
+const getBaseUrl = (params: GetResourceTypeParams) =>
+  `/organizations/${params.organizationId}/app_definitions/${params.appDefinitionId}/resource_provider/resource_types/${params.resourceTypeId}`
+
+export const get: RestEndpoint<'ResourceType', 'get'> = (
+  http: AxiosInstance,
+  params: GetResourceTypeParams
+) => {
+  return raw.get<ResourceTypeProps>(http, getBaseUrl(params))
+}
+
+export const upsert: RestEndpoint<'ResourceType', 'upsert'> = (
+  http: AxiosInstance,
+  params: GetResourceTypeParams,
+  rawData: UpsertResourceTypeProps,
+  headers?: RawAxiosRequestHeaders
+) => {
+  const data = copy(rawData)
+
+  return raw.put<ResourceTypeProps>(http, getBaseUrl(params), data, { headers })
+}
+
+export const del: RestEndpoint<'ResourceType', 'delete'> = (
+  http: AxiosInstance,
+  params: GetResourceTypeParams
+) => {
+  return raw.del(http, getBaseUrl(params))
+}
--- a/lib/adapters/REST/endpoints/resource-type.ts
+++ b/lib/adapters/REST/endpoints/resource-type.ts
@@ -7,7 +7,9 @@ import { type GetResourceTypeParams } from '../../../common-types'
 import type { RestEndpoint } from '../types'
 import type { ResourceTypeProps, UpsertResourceTypeProps } from '../../../entities/resource-type'
 
-const getBaseUrl = (params: GetResourceTypeParams) =>
+const getBaseUrl = (
+  params: GetResourceTypeParams | Omit<GetResourceTypeParams, 'resourceTypeId'>
+) =>
   `/organizations/${params.organizationId}/app_definitions/${params.appDefinitionId}/resource_provider/resource_types`
 
 const getEntityUrl = (params: GetResourceTypeParams) =>
@@ -40,7 +42,7 @@ export const del: RestEndpoint<'ResourceType', 'delete'> = (
 
 export const getMany: RestEndpoint<'ResourceType', 'getMany'> = (
   http: AxiosInstance,
-  params: GetResourceTypeParams
+  params: Omit<GetResourceTypeParams, 'resourceTypeId'>
 ) => {
   return raw.get<CollectionProp<ResourceTypeProps>>(http, getBaseUrl(params))
 }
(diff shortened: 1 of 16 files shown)


lib/entities/resource-type.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  sys: Omit<BasicMetaSysProps, 'version'> & {
+    appDefinition: SysLink
+    organization: SysLink
+    resourceProvider: SysLink
+  }
+  /**
+   * Resource Provider name, TODO
+   */
+  name: string
```
veu, line 24: The resource provider currently doesn’t have a `name` property. At the very least it should be optional but I’d prefer to leave it out completely.

lib/plain/entities/resource-type.ts
```diff
@@ -0,0 +1,83 @@
...
+   * });
+   * ```
+   */
+  delete(params: OptionalDefaults<GetResourceTypeParams>): Promise<any>
+
+  /*
+   * Fetch all Resource Types
+   * @param params entity IDs to identify the Resource Type
+   * @returns all Resource Types
+   * @throws if the request fails, or the Resource Type is not found
```
msieroslawska, line 71: Does it really throw if there are no resource types? I'd expect an empty array.

lib/plain/entities/resource-type.ts
```diff
@@ -0,0 +1,83 @@
...
+   * ```
+   */
+  get(params: OptionalDefaults<GetResourceTypeParams>): Promise<ResourceTypeProps>
+
+  /*
+   * Creates or updates a Resource Type
+   * @param params entity IDs to identify the Resource Type
+   * @param rawData the ResourceType
+   * @returns the created or updated Resource Type
+   * @throws if the request fails, the App Definition or Resource Type is not found, or the payload is malformed
```
msieroslawska, line 33: Why does it throw if Resource Type is not found? This also covers the use case when I want to create a new Resource Type.

lib/plain/entities/resource-type.ts
```diff
@@ -0,0 +1,83 @@
...
+   * @example
+   * ```javascript
+   * // You need a valid AppDefinition with an activated AppBundle that has a contentful function configured
+   * const resourceType = await client.resourceType.upsert(
```
msieroslawska, line 37: I think the example is missing `rawData`

lib/entities/resource-type.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const getUpsertParams = (data: ResourceTypeProps): UpsertResourceTypeProps => ({
+  name: data.name,
+  defaultFieldMapping: data.defaultFieldMapping,
+})
+
+/**
+ * @private
+ * @param makeRequest - function to make requests via an adapter
+ * @param data - Raw Resource Provider data
```
msieroslawska, line 130: I think these params should mention Resource Type, not Provider.

lib/entities/resource-type.ts
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * System metadata
+   */
+  sys: Omit<BasicMetaSysProps, 'version'> & {
+    appDefinition: SysLink
+    organization: SysLink
+    resourceProvider: SysLink
+  }
+  /**
+   * Resource Provider name, TODO
```
msieroslawska, line 22: Are the two `TODO` still to be done? What is missing?

lib/entities/resource-type.ts
veu: You’re right, I got confused by it saying "provider name".

lib/plain/entities/resource-type.ts
mayakarabula: true, yeah, it does not throw, just copied it too fast

lib/plain/entities/resource-type.ts
mayakarabula: yeah, it should be Resource Provider

lib/entities/resource-type.ts
mayakarabula: it's done yes, forgot to remove it

lib/adapters/REST/endpoints/resource-type.ts
```diff
@@ -0,0 +1,46 @@
+import type { RawAxiosRequestHeaders } from 'axios'
+import type { AxiosInstance } from 'contentful-sdk-core'
+import * as raw from './raw'
+import copy from 'fast-copy'
+import type { CollectionProp } from '../../../common-types'
+import { type GetResourceProviderParams, type GetResourceTypeParams } from '../../../common-types'
+import type { RestEndpoint } from '../types'
+import type { ResourceTypeProps, UpsertResourceTypeProps } from '../../../entities/resource-type'
+
+const getBaseUrl = (params: GetResourceTypeParams) =>
```
mohamedturki, line 10: shouldn't the base URL not include the resource ID?

lib/entities/resource-provider.ts
```diff
@@ -92,6 +105,44 @@ function createResourceProviderApi(makeRequest: MakeRequest) {
         params: getParams(data),
       })
     },
+
+    getResourceType: function getResourceType(id: string) {
```
mohamedturki, line 109: Is it a common pattern in this SDK to expose sub resources both in parent resource methods and standalone? Also, do we need docs for these methods?

Cyberxon: there is a separate ticket for updating the docs after we finish this one

Cyberxon: for the other question about the standalone vs parent resource in the legacy client, usually in normal cases the parent resource has create and get methods and the standalone resource (child resource) has the update and delete. Let me give you an example to make it clearer: if you want to create a new `appDefinition`, you do it through the `organization` resource, so you init the client, `getOrganisation`, and then through the organisation entity you do the `createAppDefition` so you end up with the AppDefintion entity that you can update or delete. You cannot add a get on that standalone entity because you can't call that method on the entity itself. however, in the plain client you can do it because basically you call everything related to the standalone entity on the standalone entity.

lib/adapters/REST/endpoints/resource-type.ts
Cyberxon: looking at the other examples in the same directory, you are right, the base url should be the one without the resourceId and then you have a separate method for the other url.

lib/entities/resource-provider.ts
```diff
@@ -92,6 +105,44 @@ function createResourceProviderApi(makeRequest: MakeRequest) {
...
+    upsertResourceType: function upsertResourceType(id: string, data: UpsertResourceTypeProps) {
+      return makeRequest({
+        entityType: 'ResourceType',
+        action: 'upsert',
+        params: {
+          organizationId: this.sys.organization.sys.id,
+          appDefinitionId: this.sys.appDefinition.sys.id,
+          resourceTypeId: id,
+        },
+        headers: {},
```
jjolton-contentful, line 129: Was surprised to see headers as empty object here, since headers is optional. Just entirely out of curiosity, do you know if this was required for some reason?

jjolton-contentful (approved):

lib/entities/resource-provider.ts
mayakarabula: you are right, yeah, we can remove it 👍

jjolton-contentful (approved):

PR #2430, opened 2024-09-04
fix: README.md - ES6 import example

jamespearson:
Fixed the ES6 import, as the example code did not work (contentful was undefined)

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,8 +156,8 @@ You can use the es6 import with the library as follows
 
 ```js
 // import createClient directly
-import contentful from 'contentful-management'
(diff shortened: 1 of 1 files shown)


axe312ger:
@mgoudy91 should we merge this to ensure readme is correct?

mgoudy91:
Yep this looks great, we'll get this in next week after the actual code changes

axe312ger:
Actually, if we do this, we should also update the readme to say that the browser bundles are available with `createClient` and not `contentfulManagement` (Bug itself is fixed in next major via #2472)

PR #2436, opened 2024-09-10
feat: add resource methods [DANTE-1405]

Cyberxon:
## Summary

As a last step of introducing the Native external resources' entities to CMA.js, we are adding the resource entity now.

## Description

Added Methods:

- GET all resources for a resourceType in a space/environment
- Get all resourceTypes for an environment

When adding a new method:


screamcha (approved):

lib/adapters/REST/endpoints/resource-type.ts
```diff
@@ -44,3 +58,13 @@ export const getMany: RestEndpoint<'ResourceType', 'getMany'> = (
     `/organizations/${params.organizationId}/app_definitions/${params.appDefinitionId}/resource_provider/resource_types`
   )
 }
+
+export const getForEnvironment: RestEndpoint<'ResourceType', 'getForEnvironment'> = (
```
jjolton-contentful, line 62: would this be better as `getManyForEnvironment`? definitely not mandating that, just wanted to ask in case that naming helps better convey what this is used for

lib/create-environment-api.ts
```diff
@@ -2280,5 +2283,80 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * })
+     *
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment_id>'))
+     * .then((environment) => environment.getResourceTypes({limit: 10}))
+     * .then((installations) => console.log(installations.items))
+     * .catch(console.error)
+     * ```
+     */
+    async getResourceTypes(query: BasicCursorPaginationOptions) {
```
jjolton-contentful, line 2305: is the query truly required for getResourceTypes and getResourcesForResourceType or is it optional?

lib/create-environment-api.ts
```diff
@@ -2280,5 +2283,80 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * const contentful = require('contentful-management')
+     *
+     * const client = contentful.createClient({
+     *   accessToken: '<content_management_api_key>'
+     * })
+     *
+     * // Search Query
+     * client.getSpace('<space_id>')
+     * .then((space) => space.getEnvironment('<environment_id>'))
+     * .then((environment) => environment.getResourcesForResourceType('<resource_type_id>', {query: '<search_query>', limit: 10}))
```
jjolton-contentful, line 2334: the search query can sometimes be confusing for newer users not familiar with the syntax, you might find value in providing a more explicit search query example when it comes time to toss these inline snippets into the documentation

lib/plain/entities/resource-type.ts
```diff
@@ -70,6 +74,24 @@ export type ResourceTypePlainClientAPI = {
...
+   * @example
+   * ```javascript
+   * const resourceTypes = await client.resourceType.getForEnvironment({
+   *   spaceId: '<space_id>',
+   *   environmentId: '<environment_id>',
+   * });
+   * ```
+   */
+  getForEnvironment(
+    params: OptionalDefaults<GetSpaceEnvironmentParams> & { query?: BasicCursorPaginationOptions }
```
jjolton-contentful, line 92: query is optional here but required in `lib/adapters/REST/endpoints/resource-type.ts`

jjolton-contentful (dismissed):
This implementation reads to me as comprehensive and consistent with the existing codebase, very nice work. Left a few questions and optional suggestions that could potentially improve clarity/usability.

lib/adapters/REST/endpoints/resource-type.ts
Cyberxon: I used the same naming for EnvironmentTemplateInstallation which is also a getMany.

lib/create-environment-api.ts
Cyberxon: I just rechecked, it is optional for both, thank you for noticing that

lib/create-environment-api.ts
Cyberxon: We have a follow up ticket for the docs snippet, I will make sure to incorporate that

jjolton-contentful (approved):

PR #2459, opened 2024-11-05
fix: soc2 vulns [EXT-5919]

BobHemphill76:
## Summary

Fixing security vulns found here: https://github.com/contentful/contentful-management.js/security

## Description

- BodyParser vuln via express + karma bump
- Switch to size-limit from bundlesize
- remove better-docs
- bump semantic-release

When adding a new method:


.circleci/config.yml
```diff
@@ -62,6 +62,7 @@ jobs:
     steps:
       - with-workspace
       - run: npm run build
+      - persist-workspace
```
BobHemphill76, line 65: so size can use the build artifacts

package.json
```diff
@@ -94,8 +95,6 @@
     "babel-plugin-inline-replace-variables": "^1.3.1",
     "babel-plugin-lodash": "^3.3.4",
     "babel-plugin-rewire-ts": "1.4.0",
-    "better-docs": "2.7.3",
```
BobHemphill76, line 97: doc generation wasn't built into our CI process and from what I can tell, this was unused with typedoc

package.json
```diff
@@ -127,9 +126,10 @@
     "nyc": "^15.1.0",
     "prettier": "^2.6.0",
     "rimraf": "^5.0.0",
-    "semantic-release": "^19.0.5",
+    "semantic-release": "^22.0.12",
```
BobHemphill76, line 129: tested with beta release: https://app.circleci.com/pipelines/github/contentful/contentful-management.js/10804/workflows/7d5aca0e-4cd4-4bca-ab50-40810dd12771/jobs/35194

t-col (approved):

PR #2469, opened 2024-11-18
feat: [GROOT-1550] introduce proposal for fixing update behavior in taxonomy

LiamStokingerContentful:
## Summary

This proposal looks to handle a mismatch in how taxonomy update methods are used when compared to other entities.

Long term plan in next major version is to:
- introduce new `patch` method
- change behavior of `update` method to use `PUT`

Short term plan is to:
- introduce new `patch` method
- mark `update` as deprecated, and explain change to `PUT`
- introduce temporary `updatePut` method and mark as deprecated, so that we can use `PUT` requests for the time being
- create change log post explaining upcoming changes

When adding a new method:


AdrianLThomas (commented):
Nice one Liam, this makes sense to me: but would love Tundra's eyes on it too.

LiamStokingerContentful:
> Nice one Liam, this makes sense to me: but would love Tundra's eyes on it too.

They seem to agree with the general idea - I would recommend we approve this and merge it into my other PR, and they can then review that all together 👍

lib/adapters/REST/endpoints/concept.ts
```diff
@@ -34,6 +34,26 @@ export const createWithId: RestEndpoint<'Concept', 'createWithId'> = (
...
+  data: OpPatch[],
+  headers?: RawAxiosRequestHeaders
+) => {
+  return raw.patch<ConceptProps>(
+    http,
+    `${basePath(params.organizationId)}/${params.conceptId}`,
+    data,
+    {
+      headers: {
+        'X-Contentful-Version': params.version ?? 0,
```
maxcheremisin, line 49: why do we want to provide a default version? i think the error will be confusing in case it doesn't match

lib/plain/entities/concept-scheme.ts
```diff
@@ -51,10 +51,13 @@ export type ConceptSchemePlainClientAPI = {
    * @returns the updated Concept Scheme
    * @throws if the request fails
    * @see {@link https://www.contentful.com/developers/docs/references/content-management-api/#/reference/taxonomy/concept-scheme}
+   * @deprecated The behavior of this method as a PATCH is being deprecated, and will be replaced with a PUT in the next major version. Use the `patch` method instead.
```
maxcheremisin, line 54: > in the next major version

how can we make sure we don't forget to update it?

maxcheremisin: btw the last time we updated the major version was more than a year ago. long process

LiamStokingerContentful: Fair point, but the impression I get from Tundra from our slack thread is that they'll work with us on this as planning for the next version.

maxcheremisin: what just if...

```ts
import * as package from 'package.json'

it('should not pass if major version is updated', () => {
  if (package.version.split('.')[0] === 12) {
    expect('updatePut' in sdk).to.be(false)
  }
})
```

maxcheremisin (approved):

PR #2491, opened 2024-12-02
fix: don't allow additional parameters

andipaetzold:
## Summary

Have stricter types and forbid passing additional params

## Description

Currently, the following code is valid:

```ts
import { createClient } from "contentful-management";

const client = createClient({ accessToken: "..." }, { type: "plain" });
await client.entry.getMany({
  foo: "bar",
});

```

We allow passing additional params and loose type-safety that way. The goal with this PR is to improve type-safety and DX by having stricter types and forbidding passing additional params.

With this change, the developer will receive an error:

> Object literal may only specify known properties, and 'foo' does not exist in type 'OptionalDefaults<GetSpaceEnvironmentParams & QueryParams>'

As we are touching types, you can consider this a breaking change. However, argued that way, every change is breaking.

To test this change, I introduced vitest typechecking tests.

To make typescript happy, `skipLibCheck` had to be enabled. But, IMO, that's anyway recommended to be enabled. This should speed-up general type checking as well.


denkristoffer (approved):
I absolutely think type changes should be handled like code changes but in this case I think it's fine and can be considered a fix 🙂

PR #2505, opened 2025-01-07
fix: app installation parameter types [NONE]

marcolink:
## Summary

This changes adds partially missing types to align with the types defined [here](https://github.com/contentful/contentful-management.js/blob/master/lib/adapters/REST/endpoints/app-installation.ts). 

This allows for querying app installations by `spaceId`.

When adding a new method:


BasKiers (approved):

vikaskumr (approved):

marcolink:
This is the change that broke it in our consumer: 
https://github.com/contentful/contentful-management.js/pull/2491

t-col (approved):

PR #2513, opened 2025-01-16
feat: add function endpoints [EXT-5995]

michaelpineirocontentful:
## Summary
Add function endpoints to be used through plain client

- Updated plain-client.ts to include new function operations: get, getMany, getManyForEnvironment
- Created function.ts
- Enhanced common-types to support new function operations
- Updated REST/endpoints/function.ts to handle new function endpoints

## Motivation and Context
We want to have access to basic Functions API's through CMA.

When adding a new method:

- [X] Added a unit test for the new method
- [X] Added an integration test for the new method **note: integration/e2e lives in [app-hosting-monitoring](https://github.com/contentful/app-hosting-monitoring/blob/bcea68a708dc7aa78b6245e5857a6e45a7406426/index.js#L188)**


lib/common-types.ts
```diff
@@ -2140,6 +2149,11 @@ export type GetEditorInterfaceParams = GetSpaceEnvironmentParams & { contentType
 export type GetEntryParams = GetSpaceEnvironmentParams & { entryId: string }
 export type GetExtensionParams = GetSpaceEnvironmentParams & { extensionId: string }
 export type GetEnvironmentTemplateParams = GetOrganizationParams & { environmentTemplateId: string }
+export type GetFunctionParams = GetAppDefinitionParams & { functionId: string }
+export type GetManyFunctionParams = GetAppDefinitionParams
+export type GetFunctionForEnvParams = GetSpaceParams & { environmentId: string } & {
```
t-col, line 2154:
```suggestion
export type GetFunctionForEnvParams = GetSpaceEnvironmentParams & {
```

PR #2520, opened 2025-01-22
feat: add function logs API [EXT-6001]

michaelpineirocontentful:
## Summary
Add function log endpoints to be used through plain client

- Added function log entity and plain client
- Added unit tests for function log
- Added js docstrings

## Motivation and Context
We want to have access to basic FunctionLog API's through CMA.

When adding a new method:

- [X] Added a unit test for the new method
- [X] Added an integration test for the new method **note: integration/e2e lives in [app-hosting-monitoring](https://github.com/contentful/app-hosting-monitoring/pull/293)**
- [Documentation will be added **after** function logs are added ](https://contentful.atlassian.net/browse/EXT-6133)


lib/adapters/REST/endpoints/function-log.ts
```diff
@@ -0,0 +1,41 @@
...
+} from '../../../common-types'
+import type { RestEndpoint } from '../types'
+import type { FunctionLogProps } from '../../../entities/function-log'
+
+const FunctionLogAlphaHeaders = {
+  'x-contentful-enable-alpha-feature': 'function-logs',
+}
+
+const baseURL = (params: GetAllFunctionLogParams) =>
+  `/spaces/${params.spaceId}/environments/${params.environmentId}/app_installations/${params.appInstallationId}/functions/${params.functionId}/logs`
```
jjolton-contentful, line 16: I think needs a tweak to support the skip/limit pagination query params if provided

test/unit/create-function-log-api.ts
```diff
@@ -0,0 +1,38 @@
...
+      methodToTest: 'get',
+    })
+  })
+
+  test('API call getAll', async () => {
+    return makeGetEntityTest(setup, {
+      entityType: 'FunctionLog',
+      mockToReturn: functionLogCollectionMock,
+      methodToTest: 'getMany',
+    })
```
t-col, line 36: `getMany` or `getAll`?

lib/adapters/REST/endpoints/function-log.ts
jjolton-contentful: %s/skip\/limit/cursor/ 🤦 my bad

test/unit/create-function-log-api.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+    makeRequest,
+    entitiesMock,
+  }
+}
+
+describe('A createFunctionLogApi', () => {
+  test('API call getFunctionLog', async () => {
+    return makeGetEntityTest(setup, {
+      entityType: 'Function',
```
t-col, line 25:
```suggestion
      entityType: 'FunctionLog',
```
?

lib/plain/entities/function-log.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { OptionalDefaults } from '../wrappers/wrap'
+
+export type FunctionLogPlainClientAPI = {
+  /**
+   * Fetches the specified FunctionLog
+   * @param spaceId
+   * @param environmentId
+   * @param appInstallationId
+   * @param functionId
+   * @param logId
```
t-col, line 16: I forgot that the proper usage of the `@param` directive is actually for each argument to a function, so all of these functions actually only have a single param, i.e. ...
```suggestion
   * @param {OptionalDefaults<GetSpaceEnvironmentParams>} params
```

PR #2528, opened 2025-01-27
feat: [] Bulk actions v2 support

colomolo:
## Summary
Adding types for Bulk Actions v2.

### DIFF
--- a/lib/adapters/REST/endpoints/bulk-action.ts
+++ b/lib/adapters/REST/endpoints/bulk-action.ts
@@ -5,6 +5,9 @@ import type {
   BulkActionPublishPayload,
   BulkActionUnpublishPayload,
   BulkActionValidatePayload,
+  PublishBulkActionV2Payload,
+  UnpublishBulkActionV2Payload,
+  ValidateBulkActionV2Payload,
 } from '../../../entities/bulk-action'
 import type { RestEndpoint } from '../types'
 import * as raw from './raw'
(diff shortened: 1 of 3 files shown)


jjolton-contentful (dismissed):
💪 this iteration is looking good, just some linting to resolve

colomolo:
> 💪 this iteration is looking good, just some linting to resolve

Okay, I'll add tests as well.

lib/methods/bulk-action.ts
```diff
@@ -27,3 +31,20 @@ export async function waitForBulkActionProcessing<T extends BulkActionPayload =
     options,
   )
 }
+
+/** Waits for a BulkAction V2 status to be either succeeded or failed.
+ * Used by the Plain client */
+export async function waitForBulkActionV2Processing<T extends BulkActionV2Payload = any>(
```
ethan-ozelius-contentful: Is this actually used by the plain client?  Or is it just a test helper?  If it's just a test helper, I think this belongs in the [test/helpers.ts file](https://github.com/contentful/contentful-management.js/blob/master/test/helpers.ts).

test/integration/bulk-action-integration.test.ts
```diff
@@ -257,3 +265,83 @@ describe('BulkActions Api', () => {
     })
   })
 })
+
+describe('BulkActions Api v2 (Plain Client only)', () => {
```
ethan-ozelius-contentful, line 272: Is there a reason to only support bulk actions V2 in the plain client and not the legacy?  I can see that V1 is supported in both.

lib/methods/bulk-action.ts
colomolo: I will take a look on the next week. I need to recall this stuff 🙂

colomolo: Yes, seems that it's used in tests only

ethan-ozelius-contentful (dismissed):

PR #2547, opened 2025-02-10
fix: add `version` param to patch method of plain client

andipaetzold:
## Summary

Adding missing `version` param to the plain client entry patch params

## Description

`client.entry.patch` expects `version` to be passed in, however `version` is currently not included in the TypeScript types.


BobHemphill76 (approved):
version

PR #2588, opened 2025-03-25
Refactor: clean up what many devs left untouched (improves DX)

axe312ger:
While working on the docs, I ran into a bunch of structural annoyances I couldn’t ignore anymore. Quick list:

- Got rid of unnecessary `index.ts` files, they made imports messier than they needed to be. We’ve got modern IDEs, direct imports are fine.
- Renamed a few misleadingly named files.
- Cleaned up types: some were meant for export-only but were being reused internally.
- Exported finally our actual plugin config options type (`ClientOptions`), the old deprected one was confusing and outdated, so I removed `ConfigParams`.
- Main file imports and exports were a mess, I separated type imports from regular ones and sorted everything alphabetically.
- Might do more cleanup later... but for now, lunch 😄

## Great side-effect, the bundles shrink with these changes:

original size in current master:

    dist/contentful-management.browser.js
    Size limit: 161 kB
    Size:       114.03 kB brotlied

    dist/contentful-management.browser.min.js
    Size limit: 70 kB
    Size:       55.48 kB brotlied

    dist/contentful-management.node.js
    Size limit: 177 kB
    Size:       148.42 kB brotlied

    dist/contentful-management.node.min.js
    Size limit: 88 kB
    Size:       79.58 kB brotlied

I removed the index file in `lib/entities` and forced the files to use "one layer less" to import types. Also helps with DX.

    dist/contentful-management.browser.js
    Size limit: 161 kB
    Size:       113.06 kB brotlied

    dist/contentful-management.browser.min.js
    Size limit: 70 kB
    Size:       53.24 kB brotlied

    dist/contentful-management.node.js
    Size limit: 177 kB
    Size:       148.09 kB brotlied

    dist/contentful-management.node.min.js
    Size limit: 88 kB
    Size:       77.59 kB brotlied

Then I made sure our `exported-types` are actually only used to export types

    dist/contentful-management.browser.js
    Size limit: 161 kB
    Size:       112.55 kB brotlied

    dist/contentful-management.browser.min.js
    Size limit: 70 kB
    Size:       53.01 kB brotlied

    dist/contentful-management.node.js
    Size limit: 177 kB
    Size:       147.65 kB brotlied

    dist/contentful-management.node.min.js
    Size limit: 88 kB
    Size:       77.28 kB brotlied

We still have plenty of circular dependencies, lets see if more optimization reduces bundle-size further :)


README.md
whitelisab, line 1: The changes in the `README` look great, thanks for addressing this!

lib/contentful-management.ts
```diff
@@ -42,70 +42,93 @@ interface UserAgentParams {
   feature?: string
 }
 
-/**
- * @deprecated
- */
-export type ClientParams = RestAdapterParams & UserAgentParams
```
whitelisab, line 48: I know this is already labeled as deprecated, do you think this is safe to remove?

whitelisab (commented):
Just a few small questions. Love the decreased bundle sizes!

README.md
axe312ger: Cheers :)

lib/contentful-management.ts
axe312ger: No, as these are used by the legacy client.

The whole legacy vs plain client situation should be adressed by us soon IMHO

axe312ger: @whitelisab so i checked, this is a left over from ages ago and just confuses people. so lets get rid of it! The real config types for plain and legacy client are available for our users

PR #2589, opened 2025-03-25
fix: update app event subscription to support functions [EXT-6289]

whitelisab:
## Summary

App Event Subscription types did not include the `functions` property that is needed to add a filter, transformation, or handler function to a subscription.

## Description

This PR adds a `functions` property to the `AppEventSubscriptionProps` type. This property accepts 3 different types of function links: filter, transformation, and handler. I also made the `targetUrl` property optional, because you can have an app event subscription that targets a handler function instead of a url. Finally, I updated the comments for the upsert function to show examples of targeting a url vs. targeting a function. I tested these changes in a script where I am upserting app event subscriptions that contain functions.

## Motivation and Context

This change will make it easier for developers looking to use App Event Functions.

When adding a new method:

### DIFF
--- a/lib/entities/app-event-subscription.ts
+++ b/lib/entities/app-event-subscription.ts
@@ -1,7 +1,13 @@
 import copy from 'fast-copy'
 import { toPlainObject } from 'contentful-sdk-core'
 import type { Except } from 'type-fest'
-import type { BasicMetaSysProps, DefaultElements, MakeRequest, SysLink } from '../common-types'
+import type {
+  BasicMetaSysProps,
+  DefaultElements,
+  Link,
+  MakeRequest,
+  SysLink,
+} from '../common-types'
 import enhanceWithMethods from '../enhance-with-methods'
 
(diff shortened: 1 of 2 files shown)


lewisjcs (approved):

lib/entities/app-event-subscription.ts
```diff
@@ -15,9 +21,15 @@ export type AppEventSubscriptionProps = {
    */
   sys: AppEventSubscriptionSys
   /** Subscription url that will receive events */
-  targetUrl: string
+  targetUrl?: string
```
michaelpineirocontentful, line 24: Why optional?

michaelpineirocontentful (approved):

lib/entities/app-event-subscription.ts
whitelisab: I made the `targetUrl` property optional because you can have an app event subscription that targets a handler function instead of a url. In this case, you wouldn't have anything for the `targetUrl` property and instead would have a function link for a `handler`.

PR #2652, opened 2025-06-03
feat: [GROOT-1871] add concept scheme on concepts

momenzadehvida:
This PR introduces the conceptSchemes property to taxonomy concepts.

Ticket: https://contentful.atlassian.net/browse/GROOT-1871

### DIFF
--- a/lib/entities/concept.ts
+++ b/lib/entities/concept.ts
@@ -17,6 +17,7 @@ type Concept = {
   notations: string[]
   broader: TaxonomyConceptLink[]
   related: TaxonomyConceptLink[]
+  conceptSchemes?: Link<'TaxonomyConceptScheme'>[]
   sys: {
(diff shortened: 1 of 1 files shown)


lib/entities/concept.ts
```diff
@@ -17,6 +17,7 @@ type Concept = {
   notations: string[]
   broader: TaxonomyConceptLink[]
   related: TaxonomyConceptLink[]
+  conceptSchemes?: Link<'TaxonomyConceptScheme'>[]
```
LiamStokingerContentful: How soon are we anticipating adding this as an option on writes?

I'm not sure how I feel about making this typing available for the entity in its entirety if we don't allow it on writes just yet.

LiamStokingerContentful (approved):

PR #2664, opened 2025-06-23
feat: add releaseSchema defaults to plain client creation [DX-184]

elylucas:


lib/plain/wrappers/wrap.ts
```diff
@@ -4,6 +4,7 @@ export type DefaultParams = {
   spaceId?: string
   environmentId?: string
   organizationId?: string
+  releaseSchema?: 'Release.v1' | 'Release.v2'
```
michaelphamcf: I can't remember if we talked about this... was there a reason for using `releaseSchema` instead of `releaseSchemaVersion` as described in the AC?

elylucas: good catch, it is suppose to be releaseSchemaVersion

PR #2665, opened 2025-06-24
feat: add support for releaseSchemaVersion in create/query methods [DX-187]

elylucas:


lib/entities/release.ts
```diff
@@ -81,6 +84,7 @@ export type ReleaseSysProps = {
   createdAt: ISO8601Timestamp
   updatedAt: ISO8601Timestamp
   lastAction?: Link<'ReleaseAction'>
+  schemaVersion?: 'Release.v2'
```
ethan-ozelius-contentful, line 87: Can this only be V2?  Why would this exist already if it was only for v2?

ethan-ozelius-contentful: Oh, I might understand.  V1 responses don't include this currently, is that the case?

test/integration/releasev2-integration.test.ts
```diff
@@ -0,0 +1,123 @@
+import { describe, it, beforeAll, afterAll, expect, beforeEach } from 'vitest'
```
ethan-ozelius-contentful, line 1: Nice!

test/integration/releasev2-integration.test.ts
```diff
@@ -0,0 +1,154 @@
...
+        entryId: entry.sys.id, // Using the same ID for simplicity
+        environmentId: TestDefaults.environmentId,
+        spaceId: TestDefaults.spaceId,
+      })
+      expect(entryInSpace.sys.id).toEqual(entry.sys.id)
+      expect(entryInSpace.sys.release.sys.id).toEqual(release.sys.id)
+    })
+  })
+
+  describe('and releaseSchemaVersion is not provided as a default', () => {
```
ethan-ozelius-contentful: is this description missing a word at the beginning?

PR #2667, opened 2025-06-25
feat: add support for release.entry.* pattern [DX-192]

elylucas:
Adds support for defining APIs on the release method like client.release.entry.get/getMany, etc...

This doesn't yet fully implement the get method thats in there, it was just there to implement the feature.

### DIFF
--- a/lib/entities/environment-template.ts
+++ b/lib/entities/environment-template.ts
@@ -38,6 +38,7 @@ export type EnvironmentTemplateProps = {
     contentTypeTemplates: Array<ContentTypeTemplateProps>
     editorInterfaceTemplates: Array<EditorInterfaceTemplateProps>
   }
+  forTemplatedSpaces?: boolean
 }
 
 export type CreateEnvironmentTemplateProps = Omit<EnvironmentTemplateProps, 'sys'>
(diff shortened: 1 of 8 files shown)


lib/entities/release.ts
```diff
@@ -81,6 +84,7 @@ export type ReleaseSysProps = {
   createdAt: ISO8601Timestamp
   updatedAt: ISO8601Timestamp
   lastAction?: Link<'ReleaseAction'>
+  schemaVersion?: 'Release.v2'
```
michaelphamcf: Should this be `'Release.v1' | 'Release.v2'`?

lib/common-types.ts
```diff
@@ -2404,3 +2418,7 @@ export type GetUserParams = { userId: string }
 export enum ScheduledActionReferenceFilters {
   contentTypeAnnotationNotIn = 'sys.contentType.metadata.annotations.ContentType[nin]',
 }
+
```
ethan-ozelius-contentful: This type is a repeat of line 2338

lib/entities/release.ts
elylucas: The only time this is populated when the response comes back from the API is when the version of release is v2. So it shouldn't ever be 'Release.v1'.

lib/common-types.ts
elylucas: good catch

PR #2670, opened 2025-06-26
chore: update test entry IDs in TestDefaults for failing integ tests [DX-198]

elylucas:
Created new Entries that seem to have been inadvertently deleted and updated the const file with the new ids


michaelphamcf (approved):
Yay for fixing tests! Curious how/why those entry IDs got deleted.

PR #2679, opened 2025-07-02
DX-202 Update entry.get to pass in releaseId

michaelphamcf:
## Summary

JIRA: [DX-202](https://contentful.atlassian.net/browse/DX-202)

Updates existing `entry.get` method to handle `releaseId` that is passed in. This is for plain client.

## Description

If `releaseId` is passed in to `entry.get`, a query param is added to the entry get endpoint to fetch a release entry.

When adding a new method:

[DX-202]: https://contentful.atlassian.net/browse/DX-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/adapters/REST/endpoints/entry.ts
```diff
@@ -16,10 +17,16 @@ import { normalizeSelect } from './utils'
 
 export const get: RestEndpoint<'Entry', 'get'> = <T extends KeyValueMap = KeyValueMap>(
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { entryId: string } & QueryParams,
+  params: GetReleaseEntryParams & QueryParams,
   rawData?: unknown,
   headers?: RawAxiosRequestHeaders
 ) => {
+  const releaseId = params.releaseId ?? undefined
```
elylucas: I think you could do `const releaseId = params.releaseId;` here instead of using the null coalescing operator

elylucas (approved):
Let one small nitpick, feel free to do with as you wish, besides that 🚀

lib/adapters/REST/endpoints/entry.ts
michaelphamcf: Thanks! Saw that `releaseId` wasn't referenced much, so cleaned it up a bit more, removing that variable assignment.

PR #2685, opened 2025-07-03
feat: update entry.getMany to include releasev2 params [DX-203]

tylerpina:
## Summary
https://contentful.atlassian.net/browse/DX-203
This work updates the entry.getMany endpoint to handle new releasev2 params


lib/common-types.ts
```diff
@@ -1597,8 +1597,8 @@ export type MRActions = {
       return: CollectionProp<EntryProps<any>>
     }
     getMany: {
-      params: GetSpaceEnvironmentParams & QueryParams
-      return: CollectionProp<EntryProps<any>>
+      params: GetSpaceEnvironmentParams & QueryParams & { releaseId?: string }
+      return: CollectionProp<EntryProps<any, any>>
```
ethan-ozelius-contentful, line 1601: These `<any, any>` have always confused me a bit, is the return type changing?

ethan-ozelius-contentful: Never mind, I see in `common-types.ts` that we're extending the return type to inclue the release.sys...

test/integration/entry-integration.test.ts
```diff
@@ -768,6 +786,19 @@ describe('Entry Api', () => {
...
 
     describe('releaseId is provided in params, but not in default params', () => {
+      //the entry.getMany API endpoint currently does not return release metadata
+      test('entry.getMany works', async () => {
+        const response = await createEntryClient.entry.getMany({
+          query: {
+            'release[lte]': release.sys.id,
+          },
+        })
+        expect(response.items[0].sys.id).toEqual(entry2.sys.id)
```
ethan-ozelius-contentful, line 797: do we have a guarantee that the items will always be in the correct order?  Seems like that would make sense for releases to order by publish date or release date...  But typically I don't think order is guaranteed unless you provide an order prop to the query object.  `order: 'sys.createdAt'`

https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/order/query-entries/console/js

package.json
```diff
@@ -45,7 +45,7 @@
     "test:types": "npx vitest --project types --run",
     "test:unit-watch": "npx vitest --project unit",
     "test:integration": "npx vitest --project integration --run --no-file-parallelism",
-    "test:integration-watch": "npx vitest --project integration --no-file-parallelism",
+    "test:integration-watch": "npx vitest --project integration --no-file-parallelism entry-integration",
```
ethan-ozelius-contentful, line 48: probably wanna remove this change.

PR #2691, opened 2025-07-14
fix(roles): update types

leeroyrose:
## Summary

Update role types to reflect documentationa and better reflect the api.

## Description

Adds string to the ContentModel type so it can accept the value "all". When retrieving roles using the apis, the ContentModel is set to "all" but when creating them, you are forced to use an array - ["all"]. 

Adds an action type to allow better typing and support for the "access" action.

Also, there is some confusion about the not operator. In the documentation [here](https://www.contentful.com/developers/docs/tutorials/general/roles-via-api/), it says "You can specify nested constraints for and, or, not". When referencing the table of examples, it states that not can only accept one constraint as an argument. I have personally concluded that not can be used for both, would it be possible to confirm this and possibly update the documentation? 

## Motivation and Context

This change fixes a bug wth the ContentModel permission not accepting a string and adds access as an action type for writting environment based policies.

When adding a new method:


michaelphamcf:
Hey @leeroyrose — sorry for the delay, and thanks for opening this PR!

I’ve flagged a few folks who might be impacted to take a look. In the meantime, I’d recommend introducing a new `RoleActionType` instead of replacing `ActionType`. While I agree that `ActionType` could eventually be replaced, it should be mark it as `@deprecated` first to avoid breaking changes for anyone importing it directly.

leeroyrose:
Hey, thank you for the approval.

Before I merge, would it be possible to get some clarification on my question regarding the `not` operator?

> Also, there is some confusion about the not operator...

PR #2694, opened 2025-07-15
fix: removed inapplicable query params and improved AiAction types

jfctfl:
- Use `AiActionQueryOptions` for nested client implementation of `AiAction.getMany`
- Convert `InvocationStatus`, `InvocationResultType`, `VariableType` and `StatusFilter` from enums to string types to improve compatibility (no requirement to import them)
- Use `Link<'User'> | Link<'AppDefinition'>` for subject links in `AiActionProps`

### DIFF
--- a/lib/common-types.ts
+++ b/lib/common-types.ts
@@ -180,7 +180,7 @@ import type {
   UpdateOAuthApplicationProps,
 } from './entities/oauth-application'
 import type { FunctionLogProps } from './entities/function-log'
-import type { AiActionProps, CreateAiActionProps } from './entities/ai-action'
+import type { AiActionProps, AiActionQueryOptions, CreateAiActionProps } from './entities/ai-action'
 import type {
   AiActionInvocationProps,
   AiActionInvocationType,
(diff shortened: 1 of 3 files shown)


andreascful (approved):
Good changes. Thank you!

lib/common-types.ts
```diff
@@ -948,7 +948,7 @@ export type MRActions = {
   AiAction: {
     get: { params: GetSpaceParams & { aiActionId: string }; return: AiActionProps }
     getMany: {
-      params: GetSpaceParams & QueryParams
```
elylucas, line 951: Was it intentional to remove extending QueryParams type here? This has a lot of the basic query params (like select, include, paging, etc...).

If the answer is yes and the params on QueryParams aren't applicable to th AiAction endpoints, then I would update the PR title to be a bug fix with something like:

`fix: removed inapplicable query params and improved AiAction types`

else this would technically be a breaking change

elylucas (approved):
left a comment around removing the QueryParams type and updating the semantic release comment, but besides that looks good

lib/common-types.ts
jfctfl: That's a good point, updated the PR title!

PR #2717, opened 2025-08-20
Beta: migrate to new Rollup setup

axe312ger:
* node v20
* using the `react-native` field for exports
* having the whole build setup (including ts) within rollup

based on our findings in https://github.com/contentful/contentful-batch-libs/pull/913


package.json
```diff
@@ -17,7 +20,7 @@
   "main": "./dist/cjs/index.cjs",
   "types": "dist/types/index.d.ts",
   "engines": {
-    "node": ">=18"
+    "node": ">=20"
```
elylucasctfl, line 23: Lets update the node used in CI here too https://github.com/contentful/contentful-management.js/blob/master/.circleci/config.yml#L10

elylucasctfl (approved):
Left a note about updating the version of node ran in circle. If that change requires a reapproval let me know.

package.json
axe312ger: Good point!!

PR #2724, opened 2025-08-27
feat: add release.entry.createWithId method [DX-61]

whitelisab:
## Summary

Add the `release.entry.createWithId` method.

## Description

Follow the pattern and format for other release.entry methods and other createWithId methods. 

Also a few other changes in this PR:
- Add `ReleaseEntry` methods to the `MRInternal` type in `common-types.ts` 
- Adjust the types for release entry parameters and adjust types where this is used
- Refactor the unit tests to align with how we are approaching other REST endpoint tests

I was having some trouble implementing integration tests for this method as I kept getting an error when trying to clean up the created entry. I'll address this in a future PR.

When adding a new method:


michaelphamcf (approved):
lgtm

PR #2725, opened 2025-08-28
feat: entry.patch handle releaseId [DX-209]

michaelphamcf:
## Summary

JIRA: [DX-209](https://contentful.atlassian.net/browse/DX-209)

- entry.patch takes `releaseId`
- if `releaseId` provided, use the `release.entry.patch` method instead
- light refactor of existing entry method params that use `releaseId`

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/entry.ts
+++ b/lib/adapters/REST/endpoints/entry.ts
@@ -5,19 +5,22 @@ import type { OpPatch } from 'json-patch'
 import type { SetOptional } from 'type-fest'
 import type {
   CollectionProp,
-  GetReleaseEntryParams,
+  GetEntryParams,
+  GetManyEntryParams,
   GetSpaceEnvironmentParams,
   KeyValueMap,
+  PatchEntryParams,
   QueryParams,
 } from '../../../common-types'
 import type { CreateEntryProps, EntryProps, EntryReferenceProps } from '../../../entities/entry'
 import type { RestEndpoint } from '../types'
 import * as raw from './raw'
 import { normalizeSelect } from './utils'
+import * as releaseEntry from './release-entry'
 
 export const get: RestEndpoint<'Entry', 'get'> = <T extends KeyValueMap = KeyValueMap>(
   http: AxiosInstance,
-  params: GetReleaseEntryParams & QueryParams,
+  params: GetEntryParams & QueryParams,
   rawData?: unknown,
   headers?: RawAxiosRequestHeaders
 ) => {
@@ -55,7 +58,7 @@ export const getPublished: RestEndpoint<'Entry', 'getPublished'> = <
 
 export const getMany: RestEndpoint<'Entry', 'getMany'> = <T extends KeyValueMap = KeyValueMap>(
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & QueryParams & { releaseId?: string },
+  params: GetManyEntryParams & QueryParams,
   rawData?: unknown,
   headers?: RawAxiosRequestHeaders
 ) => {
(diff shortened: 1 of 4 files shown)


lib/adapters/REST/endpoints/entry.ts
```diff
@@ -75,10 +78,16 @@ export const getMany: RestEndpoint<'Entry', 'getMany'> = <T extends KeyValueMap
 
 export const patch: RestEndpoint<'Entry', 'patch'> = <T extends KeyValueMap = KeyValueMap>(
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { entryId: string; version: number },
+  params: PatchEntryParams & { version: number } & QueryParams,
   data: OpPatch[],
   headers?: RawAxiosRequestHeaders
 ) => {
+  if (params.releaseId) {
+    return releaseEntry.patch(http, { ...params, releaseId: params.releaseId }, data, {
```
whitelisab: Just wondering about the spreading here, could it instead be written as:

```
if (params.releaseId) {
    return releaseEntry.patch(http, params, data, headers)
}
```

lib/common-types.ts
```diff
@@ -1604,15 +1604,15 @@ export type MRActions = {
...
       return: CollectionProp<EntryProps<any, any>>
     }
     get: {
-      params: GetReleaseEntryParams & QueryParams
+      params: GetEntryParams & QueryParams
       return: EntryProps<any, any>
     }
     patch: {
-      params: GetSpaceEnvironmentParams & { entryId: string; version: number }
+      params: PatchEntryParams & { version: number }
```
whitelisab: Do you also need to add `QueryParams` here? Also, is there a reason why `{ version: number}` is not part of the `PatchEntryParams` type?

lib/plain/common-types.ts
```diff
@@ -322,7 +325,7 @@ export type PlainClientAPI = {
       headers?: RawAxiosRequestHeaders
     ): Promise<EntryProps<T>>
     patch<T extends KeyValueMap = KeyValueMap>(
-      params: OptionalDefaults<GetSpaceEnvironmentParams & { entryId: string; version?: number }>,
+      params: OptionalDefaults<PatchEntryParams & { version?: number }>,
```
whitelisab: Same question about `QueryParams` and `version` here. Also line 305, should the get method have `QueryParams` as part of the type?

test/integration/entry-integration.test.ts
```diff
@@ -366,6 +371,7 @@ describe('Entry Api', () => {
       return environment
         .createEntry(contentType.sys.id, { fields: { title: { 'en-US': 'this is the title' } } })
         .then((entry) => {
+          console.log('entry', entry)
```
whitelisab: Remove this?

test/integration/utils/release-entry.utils.ts
whitelisab, line 1: Nice, I like these separated into their own utils file

test/integration/entry-integration.test.ts
michaelphamcf: Yes!

lib/common-types.ts
michaelphamcf: Hmm, yea `QueryParams` looks like it should be here, but was missing originally -- I'll add it.

For the `version`, I agree it should be part of PatchEntryParams. I think I was deciding whether or not to have `PatchEntryParams` or to just tack on releaseId to the existing params list and this slip through somehow.

lib/plain/common-types.ts
michaelphamcf: Yes i think so. Wondering why QueryParams was missing in 305 also. Updated

lib/adapters/REST/endpoints/entry.ts
michaelphamcf: yea,  this one was a little weird. I had it this way because `releaseEntry.patch`'s releaseId is required. For the headers spread, typescript didn't like it if I passed directly since it could be undefined and I think it wants those cases to be explicitly handled. Here's the TS error that shows:

```
Type 'undefined' is not assignable to type 'Partial<RawAxiosHeaders & { Accept: AxiosHeaderValue; "Content-Length": AxiosHeaderValue; "User-Agent": AxiosHeaderValue; "Content-Encoding": AxiosHeaderValue; Authorization: AxiosHeaderValue; } & { ...; }>'.ts(2345)
```

Updated to use typecasting for params.
Updated to be `headers ?? {}`

PR #2727, opened 2025-08-29
feat: update entry.createWithId to accept releaseId [DX-208]

whitelisab:
## Summary

Update the `entry.createWithId` method to support releases.

## Description

When `releaseId` is passed in as a parameter to `entry.createWithId`, the `release.entry.createWithId` method is called. Added a unit test to verify this behavior.

When adding a new method:

### DIFF
--- a/lib/adapters/REST/endpoints/entry.ts
+++ b/lib/adapters/REST/endpoints/entry.ts
@@ -7,6 +7,7 @@ import type {
   CollectionProp,
   GetEntryParams,
   GetManyEntryParams,
+  CreateWithIdReleaseEntryParams,
   GetSpaceEnvironmentParams,
   KeyValueMap,
   PatchEntryParams,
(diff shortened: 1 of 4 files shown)


lib/adapters/REST/endpoints/entry.ts
```diff
@@ -211,21 +213,29 @@ export const createWithId: RestEndpoint<'Entry', 'createWithId'> = <
   T extends KeyValueMap = KeyValueMap
 >(
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { entryId: string; contentTypeId: string },
+  params: GetSpaceEnvironmentParams & {
+    entryId: string
+    contentTypeId: string
+    releaseId?: string
```
michaelphamcf, line 227: What are your thoughts on tacking on releaseId here vs. creating a dedicated Param object for each endpoint? I saw both happening, so I went with the dedicated param for entry.patch. Looking at the other  endpoints now, I see more are structured this way. Now I'm starting to wonder if I should structure it the same way.

whitelisab: Yeah I was also unsure of which pattern we wanted to do here. I don't have a strong opinion about it. Whatever you end up doing in #2725 I can follow that pattern here.

michaelphamcf: I'll keep 2725 as is. I'm fine with this PR also. I'm probably overthinking it.

PR #2728, opened 2025-09-02
feat: entry.update supports releaseId [DX-210]

michaelphamcf:
## Summary

JIRA: [DX-210](https://contentful.atlassian.net/browse/DX-210)

`entry.update` takes `releaseId` and delegates to release entry update endpoint.

## Description

- Add `releaseId` to entry.update params
- If request has `releaseId`, call `release.entry.update` endpoint
- Unit tests for entry.update
- Integration tests for entry.update

When adding a new method:

[DX-210]: https://contentful.atlassian.net/browse/DX-210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/adapters/REST/endpoints/entry.ts
+++ b/lib/adapters/REST/endpoints/entry.ts
@@ -13,6 +13,7 @@ import type {
   PatchReleaseEntryParams,
   QueryParams,
   UpdateEntryParams,
+  UpdateReleaseEntryParams,
 } from '../../../common-types'
 import type { CreateEntryProps, EntryProps, EntryReferenceProps } from '../../../entities/entry'
 import type { RestEndpoint } from '../types'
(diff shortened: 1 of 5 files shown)


lib/adapters/REST/endpoints/entry.ts
```diff
@@ -105,10 +107,14 @@ export const patch: RestEndpoint<'Entry', 'patch'> = <T extends KeyValueMap = Ke
 
 export const update: RestEndpoint<'Entry', 'update'> = <T extends KeyValueMap = KeyValueMap>(
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { entryId: string },
+  params: UpdateEntryParams,
```
whitelisab: Do we need to support `QueryParams` here if `releaseEntry.update` has that as part of the Params type?

michaelphamcf: Oh right, I had that mentally noted but forgot! Thanks for the reminder.

PR #2732, opened 2025-09-05
feat: entry.create supports releaseId

michaelphamcf:

When adding a new method:


lib/adapters/REST/endpoints/entry.ts
```diff
@@ -206,9 +207,13 @@ export const unarchive: RestEndpoint<'Entry', 'unarchive'> = <T extends KeyValue
 
 export const create: RestEndpoint<'Entry', 'create'> = <T extends KeyValueMap = KeyValueMap>(
   http: AxiosInstance,
-  params: GetSpaceEnvironmentParams & { contentTypeId: string },
+  params: GetSpaceEnvironmentParams & { contentTypeId: string; releaseId?: string } & QueryParams,
   rawData: CreateEntryProps<T>
 ) => {
+  if (params.releaseId) {
+    return releaseEntry.create(http, params as CreateReleaseEntryParams, rawData, {})
```
michaelphamcf, line 214: The `{}` is for headers param. Since entry.create doesn't take headers, should it be the same with release.entry.create?

whitelisab: Good question. We're pretty inconsistent around when we allow headers to be passed in. I'm good with your implementation here -- allow all release.entry.* methods to have headers, but then when that method is invoked from the entry.create method, pass in an empty object for headers

PR #2737, opened 2025-09-09
feat: [GROOT-1944] Add conceptSchemes property to Concept and ConceptProps types.

gbasilio:
## Summary

Add the property `conceptSchemes` to  the `Concept` and `ConceptProps` types. The API already sends this data.

## Description

To be able to list all the concept schemes per concept (see https://contentful.atlassian.net/browse/GROOT-1944) I need to read the conceptSchemes property from the concept object. To do that, I added a conceptSchemes property to the concept type on `user_interface`. 

It works like a charm because the API sends this data. However, it causes a lot of typing issues because the client doesn't have this property on the concept type. This PR adds the property `conceptSchemes` to  the `Concept` and `ConceptProps` types on the contentful-management.js package

## Motivation and Context

Need to allow listing all the concept schemes per concept on the UI.

- [X] Changes are reflected in the documentation (type documentation automatically generated).

### DIFF
--- a/lib/entities/concept.ts
+++ b/lib/entities/concept.ts
@@ -2,6 +2,7 @@ import type { Link } from '../common-types'
 import type { LocalizedEntity } from './utils'
 
 export type TaxonomyConceptLink = Link<'TaxonomyConcept'>
+export type TaxonomyConceptSchemeLink = Link<'TaxonomyConceptScheme'>
 
 type Concept = {
   uri: string | null
@@ -15,6 +16,7 @@ type Concept = {
   note: string | null
   scopeNote: string | null
   notations: string[]
+  conceptSchemes: TaxonomyConceptSchemeLink[]
   broader: TaxonomyConceptLink[]
   related: TaxonomyConceptLink[]
   sys: {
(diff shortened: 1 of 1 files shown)


lib/entities/concept.ts
```diff
@@ -29,7 +31,7 @@ type Concept = {
 }
 
 export type ConceptProps<Locales extends string = string> = LocalizedEntity<
-  Omit<Concept, 'conceptSchemes'>,
+  Concept,
```
LiamStokingerContentful, line 34: So I think `ConceptProps` is used in context with creating and updating concepts.

From what I see in the taxonomy service, I'm not sure we actually accept any updates on this property, rather only doing these updates on the concept scheme endpoints - ie it might just be read only.

So what I'm getting at is that we might not want to change this particular line. Though @luizfonseca might wanna correct me since I think he worked on this feature.

luizfonseca: Correct ^ -- you cannot send this prop (as of today) to create/update/patch endpoints -- so it's only for query/reads.

gbasilio: Thanks @LiamStokingerContentful and @luizfonseca for your comments. Great catch during PR review!

The problem seems to be that `ConceptProps` is used for both requests and responses so adding the `conceptSchemes` property only to the responses will require a larger change than I expected. I don't think it's worth the effort and risk. 

I'll change the consumer (`user_interface`) to handle this instead and drop this PR. I'm happy to revisit this if we decide that's the right long term approach.

PR #2758, opened 2025-09-22
fix: AppActionCallProps result xor error [EXT-6783]

michaelpineirocontentful:
## Summary

AppActionCallProps should allow one of (but not both) of `result` and `error`.  This is an intermediate change. We will have a quick follow to remove the duplicate `result`  and `error`  fields from the root level, in favor of having them contained at `sys.result`  and `sys.error `. This will align with consistent org-wide usage of the `sys` field: as fields that cannot be altered by the client.

## Description

Discriminating union of `result` and `error` field:
```
export type AppActionCallProps = {
  /**
   * System metadata
   */
  sys: AppActionCallSys
  /** The execution status of the app action call, if available */
  status?: AppActionCallStatus
  & (
    | { /** Structured result when execution succeeded */ result: unknown; error?: never }
    | { /** Structured error when execution failed */ error: AppActionCallErrorProps; result?: never }
  )
}
```

## Motivation and Context
This prevents unsuitable states from being represented via the content-management SDK.

- [ ] Changes are reflected in the documentation (this will be done as a fast follow in the next PR)

### DIFF
--- a/lib/entities/app-action-call.ts
+++ b/lib/entities/app-action-call.ts
@@ -1,6 +1,6 @@
 import copy from 'fast-copy'
 import { toPlainObject } from 'contentful-sdk-core'
-import type { Except } from 'type-fest'
+import type { Except, JsonValue } from 'type-fest'
 import type {
   BasicMetaSysProps,
   AppActionCallRetryOptions,
@@ -21,7 +21,7 @@ type AppActionCallSys = Except<BasicMetaSysProps, 'version'> & {
   environment: SysLink
   action: SysLink
   appActionCallResponse?: SysLink
-}
+} & (AppActionCallSucceeded | AppActionCallProcessing | AppActionCallFailed)
 
 type RetryOptions = AppActionCallRetryOptions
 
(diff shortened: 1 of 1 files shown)


lib/entities/app-action-call.ts
```diff
@@ -34,17 +34,32 @@ export interface AppActionCallErrorProps {
   statusCode?: number
 }
 
+export type AppActionCallSucceeded = {
+  status: 'succeeded'
+  result: unknown
+  error?: never
```
jsdalton: I wonder if we should just leave these out?

Like -- we own this API, and if i'm not mistaken we _omit_ `error` from the result. So should we leave it out? This would only be needed if the API might return `error: undefined` somehow (and even so it wouldn't really matter because error is itself undefined.

This would lead to better error messaging where someSuccessfulAppActionCall.error would be a type error vs. someSuccessfulAppActionCall.error being allowed by simply being undefined (and therefore forcing the developer to handle the undefined case).

lib/entities/app-action-call.ts
```diff
@@ -34,17 +34,32 @@ export interface AppActionCallErrorProps {
   statusCode?: number
 }
 
+export type AppActionCallSucceeded = {
+  status: 'succeeded'
+  result: unknown
```
jsdalton: I noticed this: [someSuccessfulAppActionCall](https://github.com/contentful/contentful-management.js/blob/c3819daa90801e96403a68f4d46e85c453ca2a5d/lib/entities/webhook.ts#L73)

Isn't that a more accurate type? (Since know for sure it will be some valid JSON value.) Makes it a bit easier to work with vs. unknwon

PR #2767, opened 2025-09-29
feat: [CFISO-2747] add types for variables in workflow defintion and workflow

loweisz:

When adding a new method:


mdmjg (approved):
LGTM

PR #2775, opened 2025-10-01
fix: update params type for release asset process methods [DX-443]

whitelisab:
## Summary

Update `processForLocale` and `processForAllLocales` param types for release assets. 

## Description

Previously, `asset` and `locale` properties were passed in both the `params` property and as their own parameters. This removes the `asset` and `locale` properties from the `params` parameter for these methods. This aligns with these methods for asset processing ([see here](https://github.com/contentful/contentful-management.js/blob/master/lib/plain/common-types.ts#L399-L409)).

When adding a new method:

### DIFF
--- a/lib/plain/common-types.ts
+++ b/lib/plain/common-types.ts
@@ -27,8 +27,6 @@ import type {
   KeyValueMap,
   Link,
   PatchReleaseEntryParams,
-  ProcessForAllLocalesReleaseAssetParams,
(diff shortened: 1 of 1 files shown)


elylucasctfl (approved):

2wce (approved):
LGTM 🚀thank you

chrishelgert (approved):

PR #2780, opened 2025-10-15
Upcoming v12 major

axe312ger:
* migrate to new hybrid module structure that should work in all common JS environments
  * Node.js: CJS, ESM, TypeScript
  * Browsers & Mobile: Tested on desktop, iOS, and Android
  * Frameworks: Next.js, Gatsby, Nuxt, Svelte, Angular, React Native
  * Bundlers: Webpack, Rollup, Vite, Parcel
  * Tested via https://github.com/axe312ger/contentful-compatibility-cma
* build process via rollup
* dependency cleanup
* These have been added but got potentially reverted in merges. We have to double check to keep them in the release:
  * https://github.com/contentful/contentful-management.js/pull/2478
  * https://github.com/contentful/contentful-management.js/pull/2610
* These can be removed from the release if it makes sense / causes trouble:
  * https://github.com/contentful/contentful-management.js/pull/2611
  * https://github.com/contentful/contentful-management.js/pull/2590

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -41,3 +41,6 @@ jobs:
       
       - name: Test package size
         run: npm run test:size
+
+      - name: Run ES version checks
+        run: npm run check
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "build": "npm run clean && npm run build:rollup",
     "build:rollup": "NODE_ENV=production rollup -c",
     "build:docs": "typedoc",
-    "postbuild": "npm run check && npm run test:demo-projects",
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -115,23 +140,53 @@ Using [yarn](https://yarnpkg.com/lang/en/):
...
+const contentful = require('contentful-management/dist/cjs/index.cjs')
+
+// node bundle
+const contentful = require('contentful-management/dist/contentful-management.node.cjs')
+
+// react-native bundle
+const contentful = require('contentful-management/dist/contentful-management.react-native.js')
+
+// browser bundle
+const contentful = require('contentful-management/dist/contentful-management.browser.js')
```
axe312ger, line 170: This list is potentially outdated, I think we have less bundles now!

PR #2792, opened 2025-10-29
feat: add types for cursor base pagination [CAPI-2270]

mrlaessig:
Add ts support for cursor based pagination. To return the correct TypeScript types when `cursor: true` is passed in into collection enpoints for assets entries and content types.

### DIFF
--- a/lib/export-types.ts
+++ b/lib/export-types.ts
@@ -198,6 +198,7 @@ export type {
   ScheduledAction,
   ScheduledActionProps,
   ScheduledActionSysProps,
+  ScheduledActionStatus,
 } from './entities/scheduled-action'
 export type { Snapshot, SnapshotProps } from './entities/snapshot'
 export type { Space, SpaceProps } from './entities/space'
--- a/test/integration/taxonomy-integration.test.ts
+++ b/test/integration/taxonomy-integration.test.ts
@@ -231,7 +231,7 @@ describe('Taxonomy Integration', () => {
 
     const { total } = await client.concept.getTotal({})
 
-    expect(total).toBe(3)
+    expect(total).toBeGreaterThan(1)
   })
 
   it('gets a list of all concepts', async () => {
@@ -251,7 +251,7 @@ describe('Taxonomy Integration', () => {
     )
 
     const { items } = await client.concept.getMany({})
-    expect(items.length).toBe(3)
+    expect(items.length).toBeGreaterThan(1)
   })
 
   it('gets a list of all paginated concepts', async () => {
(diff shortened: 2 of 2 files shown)


lib/plain/wrappers/wrap.ts
```diff
@@ -53,11 +58,20 @@ export type WrapFn<
 /**
  * @private
  */
-export const wrap = <ET extends keyof MRActions, Action extends keyof MRActions[ET]>(
+export function wrap<ET extends keyof MRActions, Action extends keyof MRActions[ET], T>(
```
marcolink: Can we name the the generic type `T` more specific?

lib/create-environment-api.ts
```diff
@@ -727,7 +728,7 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
      * .catch(console.error)
      * ```
      */
-    getEntries(query: QueryOptions = {}) {
+    getEntries: withOptionalCursorApi(function (query = {}) {
```
marcolink:
```suggestion
    getEntries: withOptionalCursorApi(async function (query = {}) {
```

lib/common-utils.ts
```diff
@@ -3,23 +3,35 @@
...
-  CollectionProp,
   CursorPaginatedCollection,
   CursorPaginatedCollectionProp,
   MakeRequest,
+  OptionalCursorApi,
+  WrappedCollection,
 } from './common-types'
 
+/**
+ * @private
```
marcolink: Can we add a description what this is doing? I assume it mainly provides type safety?

lib/create-environment-api.ts
mrlaessig: As we there is no `await` don't think we need `async`. `withOptionalCursorApi` returns with `Promise`.

lib/common-utils.ts
mrlaessig: Done ✅

lib/plain/wrappers/wrap.ts
mrlaessig: Reverted this change because overloading the function was breaking other places where we use `wrap` to assign requests to the plain client. I ended up casting the assignments to match the plain client definitions.

test/integration/entry-integration.test.ts
```diff
@@ -26,6 +26,7 @@ import {
   updateReleaseWithEntries,
   updateReleaseEntryTitle,
 } from './utils/release-entry.utils'
+import { get } from 'lodash'
```
pavel-mareev: Is it needed?

mrlaessig: I think I already removed it :)

chasepoirier (approved):

PR #2799, opened 2025-11-10
feat: add 'agent' app location type (internal) [EXT-6969]

jjolton-contentful:
## Summary

Internal Contentful apps will soon be able to leverage the “agent” app location. While this location is currently limited to internal use, we’re exploring opportunities to extend this capability to customers building AI-powered apps. If you’re interested in creating your own AI Agent app, please contact Contentful Support to discuss your use case and upcoming possibilities.

## Description

Adds `agent` as a valid app location for app definitions, denoting via comment and type name that this is only available for internal Contentful apps at this time. Users are prevented at runtime by the CMA from setting this location for their apps, and this location will explicitly not be added to the [create-contentful-app CLI prompt menu](https://github.com/contentful/create-contentful-app/blob/main/packages/contentful--app-scripts/src/location-prompts.ts#L3-L15). 

## Motivation and Context

[create-contentful-app](https://github.com/contentful/create-contentful-app/tree/main/packages/contentful--app-scripts) leverages `AppLocation['location'][]` from the contentful-management package.

### DIFF
--- a/lib/entities/app-definition.ts
+++ b/lib/entities/app-definition.ts
@@ -24,7 +24,8 @@ export interface NavigationItem {
 
 /**
  * These locations are currently restricted to internal Contentful apps only.
- * If you are interested in using these locations for your app, please reach out to Contentful Support.
+ * If you are interested in using these locations for your app,
+ * please reach out to Contentful Support (https://www.contentful.com/support/).
  */
 type InternalLocationType = 'agent'
 


lib/entities/app-definition.ts
```diff
@@ -22,7 +22,20 @@ export interface NavigationItem {
   path: string
 }
 
-type LocationType = 'app-config' | 'entry-sidebar' | 'entry-editor' | 'dialog' | 'page' | 'home'
+/**
+ * These locations are currently restricted to internal Contentful apps only.
+ * If you are interested in using these locations for your app, please reach out to Contentful Support.
```
chasepoirier: Small nit: maybe we can add a link to an email or support url right from here for people to easily contact us if needed.

chasepoirier (dismissed):
Thanks for adding in the clarity about this being internal only, left a small comment but approved otherwise!

lib/entities/app-definition.ts
jjolton-contentful: good call, added support url in 564c4337bf36c0712de9774e656c4b14e303ec5b

PR #2800, opened 2025-11-11
chore: update to github actions and npm trusted publishing

elylucasctfl:

When adding a new method:

### DIFF
--- a/test/integration/scheduled-action-integration.test.ts
+++ b/test/integration/scheduled-action-integration.test.ts
@@ -223,7 +223,6 @@ describe('Scheduled Actions API', () => {
       } catch (error: any) {
         const parsed = JSON.parse(error.message)
         expect(parsed.status).toBe(400)
-        expect(parsed.statusText).toBe('Bad Request')
         expect(parsed.message).toBe('The resource could not be found.')
       }
     })


.github/workflows/release.yaml
```diff
@@ -0,0 +1,88 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get latest release tag
+        id: get-tag
+        run: |
+          TAG=$(gh api repos/${{ github.repository }}/releases/latest --jq .tag_name)
+          echo "tag=$TAG" >> $GITHUB_OUTPUT
+        env:
+          GITHUB_TOKEN: ${{ steps.vault.outputs.GITHUB_TOKEN }}
```
michaelphamcf, line 80: I've just started noticing that this is not getting the correct tag for the summary in the next step.
https://github.com/contentful/contentful-management.js/actions/runs/19354829208#summary-55378237318 

I'm putting together a PR in the OSS template that uses `git describe` instead to get the correct tag. https://github.com/contentful/open-source-project-template/pull/33/files

Can update now or wait til later since our other repos will need to be updated also.

michaelphamcf (approved):

PR #2801, opened 2025-11-11
fix(types): for cursor paginated calls [NONE]

marcolink:
## Summary
Fixes an issue where the inferred type from a client instance method didn't match the previous shape, and therefore introduced a breaking change on the type level. 

cc @mrlaessig what do you think?

### DIFF
--- a/lib/plain/entities/editor-interface.ts
+++ b/lib/plain/entities/editor-interface.ts
@@ -1,6 +1,8 @@
 import type { RawAxiosRequestHeaders } from 'axios'
 import type {
   CollectionProp,
+  CursorPaginatedCollectionProp,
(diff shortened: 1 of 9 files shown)


lib/plain/entities/editor-interface.ts
```diff
@@ -37,9 +39,17 @@ export type EditorInterfacePlainClientAPI = {
    * });
    * ```
    */
-  getMany(
-    params: OptionalDefaults<GetSpaceEnvironmentParams & QueryParams>,
-  ): Promise<CollectionProp<EditorInterfaceProps>>
+  getMany<
```
mrlaessig, line 42: We only support assets, entries and content types at the moment.

PR #2823, opened 2025-11-20
refactor: include op-patch from json-patch

axe312ger:
In the current state of the new-beta branch, typescript projects using the new beta would complain about the missing dependency, as it won't be installed there as a devDependency.
 
I chose in my variant to include it simply as a regular dependency and continue, but I felt now it might be better to include that one type we use everywhere in our own project, avoiding having this dependency + types completely.

See: https://github.com/contentful/contentful-management.js/blob/beta/package.json#L81

Just a suggestion, but we need that type available for projects similar to https://github.com/axe312ger/contentful-compatibility-cda/tree/main/environments/node-ts-mjs

### DIFF
--- a/lib/common-types.ts
+++ b/lib/common-types.ts
@@ -60,6 +60,7 @@ import type {
   UpdateCommentParams,
   UpdateCommentProps,
 } from './entities/comment'
+import type { ComponentTypeProps, ComponentTypeQueryOptions } from './entities/component-type'
 import type { ContentTypeProps, CreateContentTypeProps } from './entities/content-type'
 import type { EditorInterfaceProps } from './entities/editor-interface'
 import type { CreateEntryProps, EntryProps, EntryReferenceProps } from './entities/entry'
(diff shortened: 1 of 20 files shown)


lib/common-types.ts
```diff
@@ -909,6 +909,34 @@ export interface Adapter {
   makeRequest: MakeRequestWithUserAgent
 }
 
+export type OpPatch = AddPatch | RemovePatch | ReplacePatch | MovePatch | CopyPatch | TestPatch
```
marcolink, line 956: Just a nit: 
By [definition](https://datatracker.ietf.org/doc/html/rfc6902/), a **Patch Document** is a list of **Operations**. Operations can have different Operation types. Operation can be mixed within a Patch.

axe312ger: It is a 1to1 copy of the type in the depdencency, I would not change it now. It worked fine for many years.

We can still adjust then if we feel like its needed in a later version!

whitelisab (approved):

PR #2824, opened 2025-11-21
feat: add dedicated methods for cursor based pagination [CAPI-2357]

ebefarooqui:
## Summary

Introduces three dedicated cursor pagination methods, one per each relevant entity:

`getAssetsWithCursor`
`getContentTypesWithCursor`
`getEntriesWithCursor`

## Description

These changes expose three new dedicated methods in the plain client for cursor based pagination. These methods map to the `getMany` endpoint per each entity and are wrapped accordingly with the appropriate cursor pagination collection. Relevant unit and integration tests are also included.

## Motivation and Context

Previously we tried to provide correct typing for `cma.js` by checking the cursor parameter to be true. As the type system of `cma.js` is already quite complex, this introduced a bug we were not able to foresee. As a result, this was rolled back and the decision was taken to introduce dedicated methods for cursor-based pagination.

## PR Checklist

### DIFF
--- a/test/integration/comment-integration.test.ts
+++ b/test/integration/comment-integration.test.ts
@@ -5,6 +5,7 @@ import {
   createTestSpace,
   initPlainClient,
   timeoutToCalmRateLimiting,
+  waitForEnvironmentToBeReady,
 } from '../helpers'
 import type { ContentType, Entry, Environment, PlainClientAPI, Space } from '../../lib/export-types'
 
@@ -19,11 +20,9 @@ describe('Comment Api', () => {
 
   beforeAll(async () => {
     plainClient = initPlainClient()
-    space = (await createTestSpace(defaultClient, 'Comment')) as Space
-    environment = (await createTestEnvironment(
-      space,
-      'Comment Testing Environment',
-    )) as unknown as Environment
+    space = await createTestSpace(defaultClient, 'Comment')
+    environment = await createTestEnvironment(space, 'Comment Testing Environment')
+    await waitForEnvironmentToBeReady(space, environment)
     contentType = await environment.createContentType({
       name: 'Content Type',
       fields: [
--- a/test/integration/content-type-integration.test.ts
+++ b/test/integration/content-type-integration.test.ts
@@ -6,6 +6,7 @@ import {
   generateRandomId,
   getDefaultSpace,
   timeoutToCalmRateLimiting,
+  waitForEnvironmentToBeReady,
 } from '../helpers'
 import type { Environment, ContentType, Space } from '../../lib/export-types'
 
@@ -23,6 +24,7 @@ describe('ContentType Api', () => {
 
     writeSpace = await createTestSpace(defaultClient, 'ContentType')
     writeEnvironment = await createTestEnvironment(writeSpace, 'Testing Environment')
+    await waitForEnvironmentToBeReady(writeSpace, writeEnvironment)
   })
 
   afterAll(async () => {
--- a/test/integration/locale-integration.test.ts
+++ b/test/integration/locale-integration.test.ts
@@ -4,6 +4,7 @@ import {
   createTestEnvironment,
   createTestSpace,
   timeoutToCalmRateLimiting,
+  waitForEnvironmentToBeReady,
 } from '../helpers'
 import type { Space, Environment } from '../../lib/export-types'
 
@@ -14,6 +15,7 @@ describe('Locale API', () => {
   beforeAll(async () => {
     space = await createTestSpace(defaultClient, 'Locale')
     environment = await createTestEnvironment(space, 'Test')
+    await waitForEnvironmentToBeReady(space, environment)
   })
 
   afterAll(async () => {
--- a/test/integration/plain-client.test.ts
+++ b/test/integration/plain-client.test.ts
@@ -6,6 +6,7 @@ import {
   generateRandomId,
   getDefaultSpace,
   timeoutToCalmRateLimiting,
+  waitForEnvironmentToBeReady,
 } from '../helpers'
 import type { Environment, ContentType, Space } from '../../lib/export-types'
 
@@ -26,6 +27,7 @@ describe('ContentType Api', () => {
       writeSpace,
       'Testing Environment',
     )) as unknown as Environment
+    await waitForEnvironmentToBeReady(writeSpace, writeEnvironment)
   })
 
   afterAll(async () => {
--- a/test/integration/release-action-integration.test.ts
+++ b/test/integration/release-action-integration.test.ts
@@ -56,10 +56,8 @@ describe('ReleaseAction Api', () => {
       },
     }
 
-    ;[testRelease, testRelease2] = await Promise.all([
-      testEnvironment.createRelease(releasePayload),
-      testEnvironment.createRelease(releasePayload),
-    ])
+    testRelease = await testEnvironment.createRelease(releasePayload)
+    testRelease2 = await testEnvironment.createRelease(releasePayload)
 
     testReleaseAction = await testRelease.validate()
     testReleaseAction2 = await testRelease2.validate()
--- a/test/integration/release-integration.test.ts
+++ b/test/integration/release-integration.test.ts
@@ -87,22 +87,20 @@ describe('Release Api', () => {
     })
 
     it('Get Releases', async () => {
-      const [release1, release2] = await Promise.all([
-        testEnvironment.createRelease({
-          title: 'First release',
-          entities: {
-            sys: { type: 'Array' },
-            items: [makeLink('Entry', TestDefaults.entry.testEntryReleasesId)],
-          },
-        }),
-        testEnvironment.createRelease({
-          title: 'Second release',
-          entities: {
-            sys: { type: 'Array' },
-            items: [makeLink('Entry', TestDefaults.entry.testEntryReleasesId)],
-          },
-        }),
-      ])
+      const release1 = await testEnvironment.createRelease({
+        title: 'First release',
+        entities: {
+          sys: { type: 'Array' },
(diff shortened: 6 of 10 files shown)


lib/adapters/REST/make-request.ts
```diff
@@ -29,6 +29,8 @@ export const makeRequest = async <R>({
     // @ts-ignore
     endpoints[entityType]?.[action]
 
+  console.debug(endpoint)
```
asleepysamurai: Any specific reason for keeping this debug log?

lib/common-utils.ts
```diff
@@ -47,3 +49,50 @@ export function shouldRePoll(statusCode: number) {
...
+}
+
+function extractQueryParam(key: string, url?: string): string | undefined {
+  if (!url) return
+
+  const queryIndex = url.indexOf('?')
+  if (queryIndex === -1) return
+
+  const queryString = url.slice(queryIndex + 1)
+  return new URLSearchParams(queryString).get(key) ?? undefined
```
asleepysamurai, line 74:
```suggestion
  const parsedURL = new URL(url)
  return parsedURL.searchParams.get(key) ?? undefined
```

lib/adapters/REST/make-request.ts
ebefarooqui: Good catch, not at all

lib/common-utils.ts
ebefarooqui: The url we are parsing from the response is a relative url

`pagePrev: "/spaces/segpl12szpe6/environments/master/assets?pageNext=SMK3D8KEZk..."`

so we couldn't use URL here unless we mocked an appropriate base domain. Thoughts?

test/unit/test-creators/static-entity-methods.ts
```diff
@@ -7,6 +7,7 @@ export async function makeGetEntityTest(
   { entityType, mockToReturn, methodToTest, wrapperSuffix = '' },
 ) {
   const { api, entitiesMock } = setup(Promise.resolve(mockToReturn))
+  console.debug('HERE', entitiesMock[entityType])
```
pavel-mareev: Extra?

lib/common-utils.ts
```diff
@@ -47,3 +49,50 @@ export function shouldRePoll(statusCode: number) {
...
+  const pages: { prev?: string; next?: string } = {}
+
+  for (const key of PAGE_KEYS) {
+    const token = extractQueryParam(Pages[key], data.pages?.[key])
+    if (token) pages[key] = token
+  }
+
+  return {
+    ...data,
+    ...(Object.keys(pages).length && { pages }),
```
pavel-mareev: I believe API always returns `pages` object, even if it's empty. Maybe we can do the same here, what do you think?

README.md
```diff
@@ -226,6 +226,17 @@ The benefits of using the "plain" version of the client, over the legacy version
 - The ability to scope CMA client instance to a specific `spaceId`, `environmentId`, and `organizationId` when initializing the client.
   - You can pass a concrete values to `defaults` and omit specifying these params in actual CMA methods calls.
 
+## Cursor Based Pagination
```
pavel-mareev, line 230: Can we also add the title in TOC?

test/integration/asset-integration.test.ts
```diff
@@ -39,6 +39,73 @@ describe('Asset API - Read', () => {
...
+      expect((response as { total?: number }).total).toBeUndefined()
+
+      response.items.forEach((item) => {
+        expect(item.sys.type).toEqual('Asset')
+        expect(item.fields).toBeDefined()
+      })
+    })
+
+    test('returns [limit] number of items', async () => {
+      const response = await environment.getAssetsWithCursor({ limit: 3 })
```
pavel-mareev, line 62: nit: atm we can provide `skip` param here. Maybe we can disallow this param with TS to help the users avoid validation errors?

Image

test/unit/test-creators/static-entity-methods.ts
ebefarooqui: Sorry! Yes, definitely need to remove

lib/entities/entry.ts
```diff
@@ -71,3 +72,11 @@ export function wrapEntry(makeRequest: MakeRequest, data: EntryProps): Entry {
...
  */
 export const wrapEntryCollection = wrapCollection(wrapEntry)
+
+/**
+ * @private
+ */
+export const wrapEntryTypeCursorPaginatedCollection: (
+  makeRequest: MakeRequest,
+  data: CursorPaginatedCollectionProp<EntryProps>,
+) => CursorPaginatedCollectionProp<EntryProps> = wrapCursorPaginatedCollection(wrapEntry)
```
mrlaessig: Type should be infered.

```suggestion
export const wrapEntryTypeCursorPaginatedCollection  = wrapCursorPaginatedCollection(wrapEntry)
```

lib/common-utils.ts
mrlaessig: I'm fine with both

lib/common-utils.ts
```diff
@@ -47,3 +49,50 @@ export function shouldRePoll(statusCode: number) {
...
+}
+
+const Pages = {
+  prev: 'pagePrev',
+  next: 'pageNext',
+} as const
+
+const PAGE_KEYS = ['prev', 'next'] as const
+
+export function normalizeCursorPaginationResponse<T>(
```
mrlaessig, line 84: Not quite sure if this will break already existing code expecting the full URL to be present.

https://github.com/contentful/experience-packages/blob/67e3f11abb5de00f5e7984f357ad4fc020693726/packages/experience-cma-utils/src/batch/utils.ts#L81

ebefarooqui: Right now, if we assume the user doesn't parse the token themselves and send it in their next request, they will run into a validation error, which is why we're parsing here. Is this logic downstream?

mrlaessig: After some thought, we’re introducing new methods, so this should be fine. As we are already at least three people expecting to be able to just pass `page.next` from the response to query the next page (I ran into this validation error myself when trying it out).

I was just a little worried that it might break some existing behavior.

lib/common-types.ts
```diff
@@ -359,6 +359,7 @@ export interface BasicQueryOptions {
 }
 
 export interface BasicCursorPaginationOptions extends Omit<BasicQueryOptions, 'skip'> {
+  skip?: never
```
ethan-ozelius-contentful, line 362: Isn't this skip unnecessary given the Omit above?

lib/create-environment-api.ts
```diff
@@ -492,6 +497,44 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
         },
       }).then((data) => wrapContentTypeCollection(makeRequest, data))
     },
+
+    /**
+     * Gets a collection of Content Types with cursor based pagination
```
ethan-ozelius-contentful, line 502: These two URLs listed in this @param should probably point to the follow URL right?

https://www.contentful.com/developers/docs/references/content-management-api/#/introduction/cursor-pagination?

ethan-ozelius-contentful (approved):
Left 1-2 non-blocking questions, but otherwise looks good.

README.md
```diff
@@ -226,6 +227,17 @@ The benefits of using the "plain" version of the client, over the legacy version
...
+
+Cursor-based pagination is supported on collection endpoints for content types, entries, and assets. To use cursor-based pagination, use the related entity methods `getAssetsWithCursor`, `getContentTypesWithCursor`, and `getEntriesWithCursor`
+
+```js
+const response = await environment.getEntriesWithCursor({ limit: 10 });
+console.log(response.items); // Array of items
+console.log(response.pages?.next); // Cursor for next page
+```
+Use the value from `response.pages.next` to fetch the next page.
```
marcolink, line 239: Should we also add an example on how to use the `pages.next` token?

lib/common-types.ts
marcolink: Why is this type used in all the non cursor based functions as signature? 🤔

README.md
ebefarooqui: Can do

lib/common-types.ts
ebefarooqui: @ethan-ozelius-contentful It's there because unless strictly forbidden, the value can still be passed in at runtime. I wanted to make sure to have a type error there just in case.

@marcolink Could you point to where that is the case? I'm not sure I follow right now, apologies

marcolink: Here for example: 

https://github.com/contentful/contentful-management.js/pull/2824/files#diff-5d65b07a85ecfbaae4af51feabd68aefb18ecbf40d80f8076768882c66d01b73L947

(not changed in this PR)

ebefarooqui: I see. I'm not sure why, but can investigate. I was using the same type to idiomatically match how we were using it elsewhere.

marcolink: Unfortunately, changing that now would likely create a breaking change ...

ebefarooqui: Agreed, we can make a point to investigate what's happening here and if need be spin up another ticket?

README.md
```diff
@@ -226,6 +227,25 @@ The benefits of using the "plain" version of the client, over the legacy version
...
+
+```js
+const secondPage = await environment.getEntriesWithCursor({
+  limit: 2,
+  pageNext: response.pages?.next,
```
marcolink, line 244: is this a naming pattern we've already established? COuldn't we just name it `cursor`?

ebefarooqui: Yes, this is keeping in step with CMA patterns

lib/common-types.ts
marcolink: Not blocking, but we have a new major version coming up which would allow to "fix" it.

PR #2848, opened 2025-12-05
feat: expose referencingEntryId on resources [ARC-727] 

jberry92:
## Summary
Exposing referencingEntryId for resources on contentful management client

## Description

The Extensibility API exposes referencingEntryId to enable passing to invoked functions for customers to do additional computation, however it is not exposed on this client which is used in other places to invoke the call to the management API. This PR exposes this new optional field on the management client.

## Motivation and Context

Customers sometimes want to pass on the referencingEntryId for Native External References to do additional lookups or computation.

## PR Checklist

- [X] I have read the `CONTRIBUTING.md` file
- [X] All commits follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] Documentation is updated (if necessary)
- [X] PR doesn't contain any sensitive information

### DIFF
--- a/test/unit/plain/resource.test.ts
+++ b/test/unit/plain/resource.test.ts
@@ -30,4 +30,83 @@ describe('Resource', () => {
       `/spaces/spaceId/environments/envId/resource_types/${resourceTypeId}/resources`,
     )
   })
+
+  test('getMany with referencingEntryId parameter', async () => {
+    const { httpMock, adapterMock } = setupRestAdapter(
+      Promise.resolve({ data: { items: [resourceMock] } }),
+    )
+    const plainClient = createClient({ apiAdapter: adapterMock }, { type: 'plain' })
+    const response = await plainClient.resource.getMany({
+      spaceId,
+      environmentId,
+      resourceTypeId,
+      query: {
+        query: 'test',
+        referencingEntryId: 'entry-123',
+      },
+    })
+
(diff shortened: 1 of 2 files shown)


test/unit/plain/resource.test.ts
```diff
@@ -30,4 +30,83 @@ describe('Resource', () => {
...
+    const { httpMock, adapterMock } = setupRestAdapter(
+      Promise.resolve({ data: { items: [resourceMock] } }),
+    )
+    const plainClient = createClient({ apiAdapter: adapterMock }, { type: 'plain' })
+    const response = await plainClient.resource.getMany({
+      spaceId,
+      environmentId,
+      resourceTypeId,
+      query: {
+        'sys.urn[in]': '123,456',
```
elylucasctfl, line 71: curious if this test was suppose to include the referencingEntryId?

elylucasctfl (approved):
left a comment about a test but besides that 👍

test/unit/plain/resource.test.ts
jberry92: I think this is a test that makes some sense in that locale can be an optional parameter as well without the referencingEntryId - probably not one I would have added, I think copilot got a bit over excited and just tested all possibilities.

elylucasctfl: sounds good, feel free to merge when you are ready

PR #2852, opened 2025-12-16
feat: add agents and agent runs [PIC-828]

elliotmassen:
## Summary

This PR adds support for the new agent and agent run endpoints. Agents provide generative capabilities with optional access to tools and workflows, and agent runs store conversation history between a user and an agent.

## Description

The agent endpoints provide the following functionalities:

- `get`: Get a single agent's details via it's `id`
- `getMany`: Get a collection of agent details.
- `generate`: Use a single agent to generate an output via it's `id`.

The agent run endpoints provide the following functionalities:

- `get`: Get a single agent run's details via it's `id`
- `getMany`: Get a collection of agent run details.

## Motivation and Context

We want to provide access to agents and agent runs to Contentful apps, and therefore we are adding support here as well as in the Apps SDK.

## PR Checklist

### DIFF
--- a/lib/adapters/REST/endpoints/agent.ts
+++ b/lib/adapters/REST/endpoints/agent.ts
@@ -7,7 +7,7 @@ import type { RestEndpoint } from '../types'
 import * as raw from './raw'
 
 const AgentAlphaHeaders = {
-  'x-contentful-enable-alpha-feature': 'agents',
+  'x-contentful-enable-alpha-feature': 'agents-api',
 }
 
 export const get: RestEndpoint<'Agent', 'get'> = (
--- a/lib/adapters/REST/endpoints/agent-run.ts
+++ b/lib/adapters/REST/endpoints/agent-run.ts
@@ -6,7 +6,7 @@ import type { RestEndpoint } from '../types'
 import * as raw from './raw'
 
 const AgentRunAlphaHeaders = {
-  'x-contentful-enable-alpha-feature': 'agents',
+  'x-contentful-enable-alpha-feature': 'agents-api',
 }
 
 export const get: RestEndpoint<'AgentRun', 'get'> = (
--- a/test/unit/plain/agent-run.test.ts
+++ b/test/unit/plain/agent-run.test.ts
@@ -63,7 +63,7 @@ describe('AgentRun', () => {
       expect.objectContaining({
         baseURL: 'https://api.contentful.com',
         headers: expect.objectContaining({
-          'x-contentful-enable-alpha-feature': 'agents',
+          'x-contentful-enable-alpha-feature': 'agents-api',
         }),
       }),
     )
@@ -91,7 +91,7 @@ describe('AgentRun', () => {
         baseURL: 'https://api.contentful.com',
         params: { statusIn: ['COMPLETED'] },
         headers: expect.objectContaining({
-          'x-contentful-enable-alpha-feature': 'agents',
+          'x-contentful-enable-alpha-feature': 'agents-api',
         }),
       }),
     )
@@ -117,7 +117,7 @@ describe('AgentRun', () => {
         baseURL: 'https://api.contentful.com',
         params: { agentIn: ['agent-1', 'agent-2'], statusIn: ['COMPLETED', 'IN_PROGRESS'] },
         headers: expect.objectContaining({
-          'x-contentful-enable-alpha-feature': 'agents',
+          'x-contentful-enable-alpha-feature': 'agents-api',
         }),
       }),
     )
--- a/test/unit/plain/agent.test.ts
+++ b/test/unit/plain/agent.test.ts
@@ -37,7 +37,7 @@ describe('Agent', () => {
       expect.objectContaining({
         baseURL: 'https://api.contentful.com',
         headers: expect.objectContaining({
-          'x-contentful-enable-alpha-feature': 'agents',
+          'x-contentful-enable-alpha-feature': 'agents-api',
         }),
       }),
     )
@@ -62,7 +62,7 @@ describe('Agent', () => {
       expect.objectContaining({
         baseURL: 'https://api.contentful.com',
         headers: expect.objectContaining({
-          'x-contentful-enable-alpha-feature': 'agents',
+          'x-contentful-enable-alpha-feature': 'agents-api',
         }),
       }),
     )
@@ -90,7 +90,7 @@ describe('Agent', () => {
       expect.objectContaining({
         baseURL: 'https://api.contentful.com',
         headers: expect.objectContaining({
-          'x-contentful-enable-alpha-feature': 'agents',
+          'x-contentful-enable-alpha-feature': 'agents-api',
         }),
       }),
     )


lib/entities/agent-run.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export type AgentRunQueryOptions = {
+  agentIn?: string[]
+  statusIn?: AgentRunStatus[]
+}
+
+export interface AgentRun extends AgentRunProps, DefaultElements<AgentRunProps> {}
+
+export function wrapAgentRun(makeRequest: MakeRequest, data: AgentRunProps): AgentRun {
```
whitelisab: Maybe prefix `makeRequest` with an underscore (`_makeRequest`) since it's unused in this function?

lib/create-environment-api.ts
```diff
@@ -2918,5 +2922,177 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     *       }
+     *     ]
+     *   }))
+     *   .then((result) => console.log(result))
+     *   .catch(console.error)
+     * ```
+     */
+    generateWithAgent(agentId: string, payload: AgentGeneratePayload) {
+      const raw = this.toPlainObject() as EnvironmentProps
+      return makeRequest({
```
whitelisab, line 3019: For this one, it looks like you are returning the raw response; is this intentional or did you want to wrap the response like the others?

lib/entities/agent-run.ts
```diff
@@ -0,0 +1,66 @@
+import { freezeSys, toPlainObject } from 'contentful-sdk-core'
+import copy from 'fast-copy'
+import type { DefaultElements, Link, MakeRequest, MetaSysProps } from '../common-types'
+import { wrapCollection } from '../common-utils'
+
+export type AgentRunStatus = 'IN_PROGRESS' | 'FAILED' | 'COMPLETED' | 'PENDING_REVIEW' | 'DRAFT'
```
whitelisab, line 6: Maybe consider making this an enum?

lib/adapters/REST/endpoints/agent.ts
```diff
@@ -0,0 +1,68 @@
...
+export const getMany: RestEndpoint<'Agent', 'getMany'> = (
+  http: AxiosInstance,
+  params: GetSpaceEnvironmentParams & QueryParams,
+  headers?: RawAxiosRequestHeaders,
+) => {
+  return raw.get<CollectionProp<AgentProps>>(
+    http,
+    `/spaces/${params.spaceId}/environments/${params.environmentId}/ai_agents/agents`,
+    {
+      params: params.query,
```
whitelisab: Since this `getMany` method accepts `QueryParams` (which includes an optional `select` property), it should use `normalizeSelect` to ensure sys properties are always included. You can import this helper method `import { normalizeSelect } from './utils'` and then:

```suggestion
      params: normalizeSelect(params.query),
```

whitelisab (commented):
Overall looks great, just a few small comments for you to review

lib/entities/agent-run.ts
ivo-toby: In general, it's best to avoid using enums in TypeScript,  other options tend to be safer and easier to maintain.

lib/adapters/REST/endpoints/agent.ts
```diff
@@ -0,0 +1,68 @@
...
+        ...AgentAlphaHeaders,
+        ...headers,
+      },
+    },
+  )
+}
+
+export const getMany: RestEndpoint<'Agent', 'getMany'> = (
+  http: AxiosInstance,
+  params: GetSpaceEnvironmentParams & QueryParams,
```
ivo-toby: We actually don't need QueryParams here as we don't support those operations on the API (yet)

lib/entities/agent.ts
```diff
@@ -0,0 +1,79 @@
...
+export type AgentGeneratePayload = {
+  messages: Array<{
+    parts: Array<{
+      type: 'text'
+      text: string
+    }>
+    id?: string
+    role: AgentMessageRole
+  }>
+  [key: string]: unknown
```
ivo-toby: [suggestion] To tighten this type a bit

```suggestion
  threadId?: string
```

we don't allow anything else

lib/entities/agent-run.ts
```diff
@@ -0,0 +1,66 @@
...
+    id: string
+  }
+  agent: {
+    sys: Link<'Agent'>
+  }
+  space: {
+    sys: Link<'Space'>
+  }
+  title: string
+  messages?: Array<AgentRunMessage>
```
ivo-toby: Messages is not optional

```suggestion
  messages: Array<AgentRunMessage>
```

lib/entities/agent-run.ts
```diff
@@ -0,0 +1,66 @@
...
+  id: string
+  createdAt: string
+  role: AgentRunMessageRole
+  type?: AgentRunMessageType
+  content: {
+    parts: Array<AgentRunMessagePart>
+  }
+}
+
+export type AgentRunProps = {
```
ivo-toby, line 33: [nit] I'm bit confused about the naming, this is a response-type, maybe call it AgentRunResponse instead of a Prop, as that is commonly associated with input types

lib/create-environment-api.ts
```diff
@@ -2918,5 +2922,177 @@ export default function createEnvironmentApi(makeRequest: MakeRequest) {
...
+     * })
+     *
+     * client.getSpace('<space_id>')
+     *   .then((space) => space.getEnvironment('<environment_id>'))
+     *   .then((environment) => environment.getAgents())
+     *   .then((response) => console.log(response.items))
+     *   .catch(console.error)
+     * ```
+     */
+    getAgents(query: QueryOptions = {}) {
```
ivo-toby: [suggestion]
We don't have queryOptions for list-agents endpoint yet, so for now we should remove the query param

lib/entities/agent-run.ts
elliotmassen: This is a fair point, I had a question mark on this too - but as far as I can see this is the convention used elsewhere (e.g. AI Actions, Vectorization Status, Semantic Duplicates, Scheduled Actions etc). @whitelisab would you be able to let us know if I've understood that correctly - is this the right convention for return types?

lib/adapters/REST/endpoints/agent.ts
elliotmassen: This is now removed following Ivo's feedback 👍

lib/create-environment-api.ts
elliotmassen: Good call, thanks for the nudge here 👍 I've now clarified the return type here and updated it to be a wrapped agent run.

lib/entities/agent-run.ts
whitelisab: Good question! It is the convention that this should be called `AgentRunProps` as these types define the data structure of an entity, regardless of whether it's used for input or output. This pattern is used consistently across entity types in the SDK, so let's stick with the established convention on this one

whitelisab (approved):
Thanks for making the updates!

PR #2976, opened 2026-04-10
fix(ci): upstream npm self-upgrade regression [EXT-00]

jjolton-contentful:
## Summary

There is an upstream npm self-upgrade regression on Node 22.22.2: direct `npm install -g npm@latest` fails with the missing promise-retry module. That’s tracked in [npm/cli#9151](https://github.com/npm/cli/issues/9151), with the underlying fix described in [npm/cli#9152](https://github.com/npm/cli/pull/9152). 

## Description

I reproduced the failure in an isolated temp install of Node v22.22.2: direct `npm install -g npm@latest` failed with the same promise-retry stack trace.

I then verified the workaround in that same isolated environment: npm@11.11.1 installed successfully, then npm@latest upgraded successfully to 11.12.1.

## Motivation and Context

This is preventing releases from going out. We need to keep the npm upgrade because the release path is using trusted publishing guidance from semantic-release’s [GitHub Actions recipe](https://semantic-release.gitbook.io/semantic-release/recipes/ci-configurations/github-actions), dropping the upgrade entirely would be a riskier change. 

## PR Checklist

 <div id='description'>
<h3>Summary by Bito</h3>

<p>There is an upstream npm self-upgrade regression on Node 22.22.2: direct `npm install -g npm@latest` fails with the missing promise-retry module. That’s tracked in [npm/cli#9151](https://github.com/npm/cli/issues/9151), with the underlying fix described in [npm/cli#9152](https://github.com/npm/cli/pull/9152). This pull request fixes the issue by modifying the release workflow to use a workaround that first installs npm@11.11.1 before upgrading to the latest version.</p>

<details>
<ul>

<li>Introduces a multi-step npm installation process in the release workflow to bypass the promise-retry module issue on Node 22.22.2, with an explanatory comment added for clarity.</li>

</ul>
</details>

</div>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -51,7 +51,10 @@ jobs:
           cache: 'npm'
 
       - name: Install latest npm
-        run: npm install -g npm@latest
+        run: |
+          # Work around the npm self-upgrade regression on Node 22.22.2.
+          npm install -g npm@11.11.1
+          npm install -g npm@latest
 
       - name: Install dependencies
         run: npm ci


chasepoirier (approved):
Fix looks correct — bootstrapping to 11.11.1 first is the right workaround for the Node 22.22.2 self-upgrade regression. Confirmed the two-step approach resolves it.

PR #2987, opened 2026-04-20
fix(component-type): sync tree node types with upstream schema [DX-905]

tylerpina:
[DX-905](https://contentful.atlassian.net/browse/DX-905)

## Summary
- Add `FragmentNode` type (`nodeType: 'Fragment'`, `fragmentId`) and include it in the `TreeNode` union
- Add `name?: string` to all four tree node types: `ComponentNode`, `ViewNode`, `FragmentNode`, `SlotNode`
- Widen `ComponentNode.contentProperties` to accept `| string` for whole-object binding shorthand

## Test plan
- [ ] `npx vitest run test/unit` shows no regression vs. baseline (31 failed / 189 passed — pre-existing browser-unit infra failures)
- [ ] `FragmentNode` exported and present in `TreeNode` union
- [ ] `name?: string` on all four node types
- [ ] `ComponentNode.contentProperties` accepts both record and string

[DX-905]: https://contentful.atlassian.net/browse/DX-905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/entities/component-type.ts
+++ b/lib/entities/component-type.ts
@@ -108,7 +108,7 @@ export type SlotNode = {
   slotId: string
 }
 
-export type TreeNode = ComponentNode | ViewNode | FragmentNode | SlotNode
+export type TreeNode = ComponentNode | FragmentNode | SlotNode
 
 // Data type field for content bindings
 export type ComponentTypeDataTypeField = {
--- a/lib/entities/component-type.ts
+++ b/lib/entities/component-type.ts
@@ -78,9 +78,10 @@ export type DesignPropertyValue =
 // Tree node types for component tree
 export type ComponentNode = {
   id: string
+  name?: string
   nodeType: 'Component'
   componentTypeId: string
-  contentProperties: Record<string, ContentPropertyPointerValue | unknown>
+  contentProperties: Record<string, ContentPropertyPointerValue | unknown> | string
   designProperties: Record<string, DesignPropertyValue>
   slots: Record<string, TreeNode[]>
   contentBindings?: string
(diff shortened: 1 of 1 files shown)


lib/entities/component-type.ts
```diff
@@ -78,27 +78,37 @@ export type DesignPropertyValue =
...
 export type ViewNode = {
   id: string
+  name?: string
   nodeType: 'View'
   viewId: string
 }
 
+export type FragmentNode = {
+  id: string
+  name?: string
```
whitelisab, line 99: This optional name property wasn't in #2986, just checking that you are intentionally adding it?

lib/entities/component-type.ts
```diff
@@ -78,27 +78,37 @@ export type DesignPropertyValue =
...
   nodeType: 'Component'
   componentTypeId: string
-  contentProperties: Record<string, ContentPropertyPointerValue | unknown>
+  contentProperties: Record<string, ContentPropertyPointerValue | unknown> | string
   designProperties: Record<string, DesignPropertyValue>
   slots: Record<string, TreeNode[]>
   contentBindings?: string
 }
 
 export type ViewNode = {
```
whitelisab, line 90: Do `ViewNode` and `viewId` need to renamed to `experience`? Or is this a separate entity?

lib/entities/component-type.ts
tylerpina: Yes, the name property should be here according to the ticket. I shouldn't have introduced FragmentNode at all in 2986, but I did the tickets out of order, so messed up there. I'll go remove it from 2986 and merge this in first. Thanks!

lib/entities/component-type.ts
tylerpina: They are not called out in the renaming PR so I don't think so. The contract schema states that ViewNode is a deprecated reference:
```
/**
 * @deprecated Legacy nested view reference kept for management-api backward compatibility.
 * Prefer `FragmentNode`.
 */
```
So maybe it's better to remove it all together?

whitelisab (approved):

PR #3013, opened 2026-04-29
feat: type ExO query filter parameters for Template, Experience, and Fragment [DX-935]

tylerpina:
[DX-935](https://contentful.atlassian.net/browse/DX-935)

## Summary
- Introduced shared `ExoQueryFilters` interface in `lib/common-types.ts` mirroring the four upstream filter schemas (`SysFiltersSchema`, `TagFiltersSchema`, `TaxonomyConceptsFiltersSchema`, `TextFiltersSchema`) from `experiences-management-api-contract`
- Replaced `[key: string]: unknown` escape hatches on `ComponentTypeQueryOptions`, `ExperienceQueryOptions`, and `TemplateQueryOptions` with typed intersections of `CursorPaginationParams & ExoQueryFilters & { order? }`
- Fragment will use `ExoQueryFilters` the same way when its entity file is created

## Test plan
- [ ] Unit tests pass: `npm run test:unit`
- [ ] Confirm passing an unknown filter key to any of the three query option types produces a compile error
- [ ] Confirm all 40 upstream filter fields (`sys.*`, `metadata.tags.*`, `metadata.concepts.*`, `name`, `name[match]`) are accepted without error

[DX-935]: https://contentful.atlassian.net/browse/DX-935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/entities/experience.ts
+++ b/lib/entities/experience.ts
@@ -2,6 +2,7 @@ import type {
   CursorPaginatedCollectionProp,
   CursorPaginationParams,
   ExoMetadataProps,
+  ExoQueryFilters,
   Link,
 } from '../common-types'
 import type {
@@ -62,10 +63,9 @@ export type ExperienceProps = ExperienceCommonProps & {
   sys: ExperienceSys
 }
 
-// Query options for getMany - cursor-based pagination with mutual exclusivity
-export type ExperienceQueryOptions = CursorPaginationParams & {
+// Query options for getMany - cursor-based pagination with typed filter fields
+export type ExperienceQueryOptions = CursorPaginationParams & ExoQueryFilters & {
   order?: string
-  [key: string]: unknown
 }
 
 // Locale-based publish payload — add or remove specific locales.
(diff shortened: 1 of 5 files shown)


lib/entities/experience.ts
```diff
@@ -57,31 +54,73 @@ type ExperienceCommonProps = {
...
 
-// Query options for getMany - cursor-based pagination with mutual exclusivity
+// Query options for getMany - cursor-based pagination with typed filter fields
+// Fields mirror ExperienceListFilterSchema from experiences-management-api-contract:
+//   SysFiltersSchema (31) + TagFiltersSchema (3) + TaxonomyConceptsFiltersSchema (4) + TextFiltersSchema (2)
 export type ExperienceQueryOptions = CursorPaginationParams & {
   order?: string
-  [key: string]: unknown
+  // SysFiltersSchema
+  'sys.id'?: string
```
michaelphamcf: I think we could make a shared filters type since these are identical, and use that for experiences and templates. 

My bad on the PR description -- it missing information. Basically, we can also use the shared filters for component types and fragments in addition to experience and templates.

michaelphamcf (approved):

PR #3014, opened 2026-04-29
fix: align DesignPropertyValue with upstream schema refactor [DX-990]

tylerpina:
[DX-990](https://contentful.atlassian.net/browse/DX-990)

## Summary
- `DesignPropertyValue` is now the canonical base type (`ManualDesignValue | DesignTokenValue` only), matching upstream `DesignPropertyValueSchema` post INTEG-3511 (2026-03-26)
- New `ComponentTreeDesignPropertyValue` union introduced as the tree-node-level type — `DesignPropertyPointerValue` and `DimensionedDesignPropertyValue` are siblings of the canonical value, not nested inside it
- `ComponentNode.designProperties` updated to `Record<string, ComponentTreeDesignPropertyValue>`

## Test plan
- [ ] Unit tests pass: `npm run test:unit`
- [ ] Confirm assigning `{ breakpoint: pointer }` to `DesignPropertyValue` produces a type error
- [ ] Confirm `ComponentTreeDesignPropertyValue` accepts pointer, dimensioned record, and canonical value as siblings

[DX-990]: https://contentful.atlassian.net/browse/DX-990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/entities/component-type.ts
+++ b/lib/entities/component-type.ts
@@ -80,18 +80,23 @@ export type DesignTokenValue = {
 }
 
 // For designProperties[].defaultValue (definition level)
-// Upstream: DesignPropertyValueSchema (schema line 86-91)
+// Upstream: DesignPropertyValueSchema — canonical value only (no pointer, no dimensioned)
 export type DesignPropertyDefinitionValue = ManualDesignValue | DesignTokenValue
 
-// For componentTree node designProperties (tree-node level)
-// Upstream: ComponentTreeDesignPropertyValueSchema (schema line 101-104)
-export type DesignPropertyValue =
-  | ManualDesignValue
-  | DesignTokenValue
-  | DesignPropertyPointerValue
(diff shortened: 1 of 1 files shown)


lib/entities/component-type.ts
```diff
@@ -79,27 +80,32 @@ export type DesignTokenValue = {
 }
 
 // For designProperties[].defaultValue (definition level)
-// Upstream: DesignPropertyValueSchema (schema line 86-91)
+// Upstream: DesignPropertyValueSchema — canonical value only (no pointer, no dimensioned)
```
michaelphamcf: I think it's time to have these comments removed. If you think it would help for a future debugging session, maybe list the upstream mapping in the PR description? I'm okay without that too -- I think the naming is fairly close to the upstream schema that we won't need it.

michaelphamcf (approved):

PR #3033, opened 2026-05-13
fix: implement getManyWithCursor REST adapter for Entry, Asset, ContentType [DX-1078]

tylerpina:
[DX-1078](https://contentful.atlassian.net/browse/DX-1078)

## Summary
- `client.entry.getManyWithCursor()`, `client.asset.getManyWithCursor()`, and `client.contentType.getManyWithCursor()` were throwing `Unknown endpoint` because the corresponding REST adapter exports were missing
- Added `getManyWithCursor` to `lib/adapters/REST/endpoints/entry.ts`, `asset.ts`, and `content-type.ts` — each calls the same HTTP endpoint as `getMany` but injects `cursor: true` into query params to activate cursor pagination mode and returns `CursorPaginatedCollectionProp`
- Added explicit `releaseId` guard in Entry and Asset (throws instead of silently ignoring); added unit tests for all three adapters including a new `content-type.test.ts`

## Test plan
- [ ] `npm run test:unit` passes (791/791)
- [ ] `npx tsc --noEmit` clean
- [ ] `client.entry.getManyWithCursor({ spaceId, environmentId })` makes a real HTTP request and returns cursor-paginated response
- [ ] `client.asset.getManyWithCursor(...)` and `client.contentType.getManyWithCursor(...)` behave the same
- [ ] Passing `releaseId` to Entry or Asset `getManyWithCursor` throws a clear error

[DX-1078]: https://contentful.atlassian.net/browse/DX-1078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ 
 <div id='description'>
<h3>Summary by Bito</h3>

<p>This PR fixes 'Unknown endpoint' errors by implementing the missing getManyWithCursor REST adapter methods for Entry, Asset, and ContentType entities in the Contentful Management.js SDK. Each adapter injects cursor: true into API query parameters and normalizes responses using the existing normalizeCursorPaginationResponse utility, enabling cursor-based pagination as an alternative to offset-based pagination.</p>

<details>
<ul>

<li>Added getManyWithCursor method to entry.ts, asset.ts, and content-type.ts REST adapters that injects cursor: true into query params and returns CursorPaginatedCollectionProp via normalizeCursorPaginationResponse, fixing previously broken endpoint calls</li>

<li>Added explicit releaseId parameter guards in entry.ts and asset.ts getManyWithCursor methods that throw clear errors ('getManyWithCursor is not supported for release-scoped entries/assets') instead of silently proceeding</li>

<li>Added 3 unit tests each for Entry and Asset adapters covering cursor pagination, pageNext token handling, and releaseId guard enforcement; created new content-type.test.ts with 2 tests covering the same</li>

<li>Centralized integration test timeouts (testTimeout: 120000ms, hookTimeout: 180000ms) in vitest.workspace.js and removed redundant explicit beforeAll timeouts from locale-integration.test.ts</li>

<li>Enabled previously skipped integration tests for AiAction, BulkAction, ReleaseAction, and Release APIs by removing .skip from describe and it blocks</li>

</ul>
</details>

</div>


test/unit/adapters/REST/endpoints/entry.test.ts
```diff
@@ -153,6 +153,60 @@ describe('Rest Entry', () => {
...
+      })
+      .then((r: any) => {
+        expect(httpMock.get.mock.calls[0][1].params).to.eql({
+          cursor: true,
+          pageNext: 'next-cursor-token',
+        })
+        expect(r.pages).to.eql({})
+      })
+  })
+
```
michaelphamcf, line 209: maybe add a test case for the releaseId guard:

```
  test('getManyWithCursor throws when releaseId is provided', async () => {
    const { adapterMock } = setup(Promise.resolve({}))
    await expect(
      adapterMock.makeRequest({
        entityType: 'Entry',
        action: 'getManyWithCursor',
        userAgent: 'mocked',
        params: { spaceId: 'space123', environmentId: 'master', releaseId: 'release123' },
      })
    ).rejects.toThrow('getManyWithCursor is not supported for release-scoped entries')
  })
```

test/unit/adapters/REST/endpoints/asset.test.ts
```diff
@@ -301,4 +301,58 @@ describe('Rest Asset', { concurrent: true }, async () => {
...
+      })
+      .then((r: any) => {
+        expect(httpMock.get.mock.calls[0][1].params).to.eql({
+          cursor: true,
+          pageNext: 'next-cursor-token',
+        })
+        expect(r.pages).to.eql({})
+      })
+  })
 })
```
michaelphamcf, line 370: same as entry.test.ts -- can add a test case for the releaseId guard

michaelphamcf (dismissed):

lib/adapters/REST/endpoints/asset.ts
```diff
@@ -83,6 +86,28 @@ export const getMany: RestEndpoint<'Asset', 'getMany'> = (
...
+  if (params.releaseId) {
+    throw new Error('getManyWithCursor is not supported for release-scoped assets')
+  }
+
+  return raw
+    .get<CursorPaginatedCollectionProp<AssetProps>>(
+      http,
+      `/spaces/${params.spaceId}/environments/${params.environmentId}/assets`,
+      {
+        params: { cursor: true, ...(params.query ?? {}) },
```
ethan-ozelius-contentful, line 104: Do we need to `normalizeSelect()` the rest of the params?  https://github.com/contentful/contentful-management.js/blob/master/lib/adapters/REST/endpoints/asset.ts#L80

lib/adapters/REST/endpoints/asset.ts
```diff
@@ -83,6 +86,28 @@ export const getMany: RestEndpoint<'Asset', 'getMany'> = (
...
+        headers: headers ? { ...headers } : undefined,
+      },
+    )
+    .then(normalizeCursorPaginationResponse)
```
ethan-ozelius-contentful, line 108: 1. Isn't the `.then()` already handled in [create-environment-api](https://github.com/contentful/contentful-management.js/blob/master/lib/create-environment-api.ts#L1137)?  I think this `.then()` is redundant.

ethan-ozelius-contentful: Same with content-types & entries.

lib/adapters/REST/endpoints/asset.ts
ethan-ozelius-contentful: Same with entries.  I'm not sure if this is necessary for entries, as it is not currently implemented for the non-cursor-pagination getMany(), at first glance it seem like it _should_ be, but worth an investigation.

lib/adapters/REST/endpoints/asset.ts
tylerpina: I'm pretty sure the two paths are independent, as the create-environment-api calls `action: getMany`, not the new getManyWithCursor. It might even be worth changing the create-env-api to call this new endpoint instead?

lib/adapters/REST/endpoints/asset.ts
tylerpina: Since `select` isn't passed as a param for cursor paginated endpoints, I didn't think that we needed to normalize, since normalization just makes sure that select queries use 'sys'. Correct me if I'm wrong here.

ethan-ozelius-contentful: That is 100% correct, today I learned.  This feels like a gap that we don't provide a select param for cursor based pagination queries.  But not related to this PR.

ethan-ozelius-contentful (approved):

